PR #1032, opened 2026-07-07
fix(permissions): match star patterns for custom tools

fengjikui:
## Summary
- make `ToolName(*)` permission patterns match custom tools that do not have a built-in primary argument mapping
- cover both explicit deny and allow behavior for custom tools

## Root cause
`allowedTools`, `deniedTools`, and project/session approval persistence use `ToolName(*)` as the broad tool pattern. The permission matcher only tested argument patterns against a known primary argument, so custom tools without a built-in primary argument fell through to the mode fallback instead of matching `(*)`.

## Validation
- `corepack pnpm@8.15.4 --filter @robota-sdk/agent-core test -- src/permissions/__tests__/permission-gate.test.ts`
- `corepack pnpm@8.15.4 --filter @robota-sdk/agent-core test`
- `corepack pnpm@8.15.4 --filter @robota-sdk/agent-core typecheck`
- `corepack pnpm@8.15.4 --filter @robota-sdk/agent-core lint` (passes with existing warnings)
- `git diff --check`

### DIFF
--- a/packages/agent-core/src/permissions/__tests__/permission-gate.test.ts
+++ b/packages/agent-core/src/permissions/__tests__/permission-gate.test.ts
@@ -30,4 +30,20 @@ describe('evaluatePermission deny precedence', () => {
     });
     expect(decision).toBe('deny');
   });
+
+  it('ToolName(*) deny pattern matches custom tools without a primary argument', () => {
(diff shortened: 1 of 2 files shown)


woojubb:

PR #1033, opened 2026-07-07
fix(tools): bind search tools to session cwd

fengjikui:
## Summary
- create cwd-aware Glob/Grep factories and use them when assembling default session tools
- resolve default and relative Glob/Grep search roots against the session cwd
- reject explicit Glob/Grep search paths outside the session cwd

## Root cause
Read, Write, and Edit are created with the session cwd and use the path guard, but Glob and Grep were shared singleton tools. Their default searches used process.cwd(), and explicit path values were resolved without checking the session workspace boundary. Because these are read/search tools, that could make a session search the host process cwd or inspect a caller-provided path outside the configured workspace.

## Validation
- `corepack pnpm@8.15.4 --filter @robota-sdk/agent-tools test`
- `corepack pnpm@8.15.4 --filter @robota-sdk/agent-framework test -- src/assembly/__tests__/create-tools.test.ts`
- `corepack pnpm@8.15.4 --filter @robota-sdk/agent-tools typecheck`
- `corepack pnpm@8.15.4 --filter @robota-sdk/agent-framework typecheck`
- `corepack pnpm@8.15.4 --filter @robota-sdk/agent-tools lint` (passes with existing warnings)
- `corepack pnpm@8.15.4 --filter @robota-sdk/agent-framework lint` (passes with existing warnings)
- `git diff --check`


woojubb:

PR #1048, opened 2026-07-07
fix(tools): block symlink path escapes

fengjikui:
## Summary
- Harden `checkPathWithinCwd` with realpath validation for existing targets and the nearest existing ancestor.
- Block cwd-contained symlinked files and symlinked directories when they resolve outside the working directory.
- Preserve the existing lexical fallback for non-existent fake cwd paths used by unit tests.

## Why
The previous guard only compared `path.resolve()` strings. A path under `cwd` could still point outside the working directory through a symlink, which lets Read/Edit follow symlinked files or Write target files under symlinked directories outside the intended workspace boundary.

## Validation
- `corepack pnpm --filter @robota-sdk/agent-tools exec vitest run src/__tests__/path-guard.test.ts` — 10 passed
- `corepack pnpm --filter @robota-sdk/agent-framework exec vitest run src/__tests__/read-tool.test.ts src/__tests__/write-tool.test.ts src/__tests__/edit-tool.test.ts` — 39 passed
- `corepack pnpm --filter @robota-sdk/agent-tools exec eslint src/builtins/path-guard.ts src/__tests__/path-guard.test.ts` — passed
- `git diff --check` — passed

Note: `corepack pnpm --filter @robota-sdk/agent-tools test` is currently blocked locally by existing `FunctionTool` / `ToolRegistry` runtime export failures outside this diff; the targeted `path-guard` test in that same package passes.

### DIFF
--- a/packages/agent-tools/src/__tests__/path-guard.test.ts
+++ b/packages/agent-tools/src/__tests__/path-guard.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect } from 'vitest';
+import { mkdtempSync, mkdirSync, rmSync, symlinkSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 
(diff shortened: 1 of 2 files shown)


woojubb:

PR #1232, opened 2026-07-18
feat(evals): neutral eval-definition API + runner (SELFHOST-011 P1)

woojubb:
## SELFHOST-011 P1 — evals-as-code (neutral SDK library seam)

Promotes the design-gated SELFHOST-011 epic (GATE-APPROVAL ENDORSE iteration 2) toward VISION.md: an **SDK user building their own agent** gets a supported way to **define evals (metrics over their agent's runs) and run them** — the neutral library seam here; the `robota eval` CI gate + example land in P2.

### What ships (P1)
New `agent-framework/src/evals/` subsystem, mirroring the pure `src/self-hosting/`/`src/goal/` template and the `agent-session-analytics` metrics-over-a-run sibling:

| File | Contract |
|------|----------|
| `eval-types.ts` | `IMetric` — a **pure fn over the SSOT `IExecutionResult`** (response + tool trajectory + usage + history, **not** the string alone); `IEvalCase`, `IEvalDefinition` (cases × metrics × threshold), `IEvalReport`/`IEvalCaseResult`/`IEvalMetricScore`, `TEvalRunFn` |
| `runner.ts` | `defineEval` (validate + default threshold→1) + `runEval(def, runFn)` — runs each case through the **injected** `runFn`, scores over `IExecutionResult`, boolean→1/0, `overallScore` = mean of case×metric scores, `passed = overallScore >= threshold`. **IO-free / provider-free.** |
| `session-run-fn.ts` | `createSessionRunFn(runtime)` — the default `runFn` from `createAgentRuntime().createSession()`, capturing the terminal **`complete`-event full `IExecutionResult`** (fresh session per case). **NOT `createQuery`** (which yields only `result.response`, dropping `toolSummaries`/`usage`/`history` — the string-only shape Alternative 4 rejected). |

Barrel + root `// ── Evals ──` re-export. **Neutral:** no concrete metric/dataset **content** in `packages/` (consumer-supplied; example in P2).

### Verification
- **TC-01/TC-02/TC-06** green (`src/evals/__tests__/runner.test.ts`, 8 tests; injected fake `runFn` + synthetic `IExecutionResult`, no live provider).
- Regression: agent-framework **143 files / 1207 tests**, typecheck clean, evals lint **0 warnings**, **57/57 harness scans**.
- SPEC.md updated (What-lives-here + Type Ownership + Public API).

### Deferred (named, not silently dropped)
TC-03 (`robota eval` CLI exit-code gate) / TC-04 (example) / TC-05 (mechanical neutrality floor) + the **agent-run capability verification** (drive a real coding agent through `robota eval`, confirm exit 1 on breach) land in **P2/P3** per the capability-reachability rule — P1 is the neutral library seam only.


woojubb:
## Independent code review — resolved

An independent `pr-review-reviewer` pass confirmed the runner verdict math, terminal-event semantics, neutrality, and layering are sound, with **2 SHOULD** findings — both now fixed in `bf5692cb0`:

| Severity | Finding | Resolution |
|----------|---------|------------|
| SHOULD | `createSessionRunFn` spawned a fresh session per case but never shut it down — `cleanup()` only removed the 3 eval listeners, abandoning each session + its background runners (N leaked over an N-case eval; could block `robota eval` process exit) | `await session.shutdown()` in a `finally` (extracted `awaitRun`) |
| SHOULD | `IEvalMetricScore.normalized` is documented `[0,1]` but a numeric metric returning > 1 could force a false aggregate pass (mean 1.5 ≥ 1 despite a 0-scoring case) | `normalizeScore` clamps numeric scores to `[0,1]`; `IMetric` doc states numeric scores are `[0,1]` |

Plus the CONSIDER items: added `session-run-fn.test.ts` (4 tests — full-result capture + shutdown, interrupted-scorable, error-rejects-but-shuts-down, fresh-session-per-case + options) closing the untested-wiring gap, a clamp regression test, and documented the `interrupted`-scored + `bypassPermissions` security posture.

Green: agent-framework **144 files / 1212 tests**, typecheck, 0 evals lint warnings, 57/57 harness scans.

PR #1381, opened 2026-07-25
ci(harness): HARNESS-024 env-gated live provider smoke + non-blocking scheduled workflow

woojubb:
## HARNESS-024 — env-gated live provider smoke

Every provider boundary in the test suite is scripted or replayed. That is the right default
(deterministic, free, offline), but it makes a whole class of breakage structurally invisible to PR
CI: the auth handshake, vendor wire-format drift, a required request field the adapter stopped
sending (the Anthropic 400 / `maxTokens` family), and the shape of a streaming response. Three such
regressions were each caught only by a human running the CLI by hand. This PR is the mechanical
replacement.

### What's here

- **`scripts/harness/live-provider-smoke.mjs`** — one minimal **real** turn per credentialed
  provider: non-streaming `chat()` (auth + request/response wire format) and streaming
  `chatStream()` (chunks arrive, text assembles), both capped at `maxTokens: 32`. Both must return
  non-empty assistant text or the provider is reported FAIL.
- **`scripts/harness/__tests__/live-provider-smoke.test.mjs`** — 20 network-free unit tests over the
  selection, redaction, verdict and report logic.
- **`.github/workflows/live-provider-smoke.yml`** — daily 05:37 UTC (a slot distinct from
  mutation-nightly 03:17, CodeQL 04:27, security Mon 04:47) + `workflow_dispatch`.

### Gating — it can never fail for lack of credentials

A provider is selected **only** when its definition's `defaults.apiKey` is an `$ENV:<NAME>`
reference **and** `<NAME>` is present and non-empty. No credential at all → clean SKIP, exit 0.

The provider list and the key env-var names are read from the provider **definitions**, so this
script contains no provider-name table and a newly added provider is covered automatically.

In the workflow, an unprovisioned secret expands to `''`, so until the owner adds a key this job is
a **green no-op**.

### Non-blocking posture

Scheduled + manual only. Not a PR gate, not a required check — same rollout posture as
`mutation-nightly`. A PR can never be blocked by a third party's API being down or a spent quota.

### Secrets are never printed

Every line the script writes — stdout and the JSON report — passes through `redactSecrets`, which
scrubs the resolved key values, because vendor SDK errors sometimes echo request headers back. The
uploaded report records provider / model / latency and the key **variable names** only.

### Verification (run locally, foreground)

| Scenario | Command | Result |
| --- | --- | --- |
| No key → clean skip | `env -u ANTHROPIC_API_KEY node scripts/harness/live-provider-smoke.mjs` | `live-provider-smoke: SKIPPED — no provider credentials in this environment.` … `This is the expected outcome without credentials and is not a failure.` — **exit 0** |
| Key present → live call | `node scripts/harness/live-provider-smoke.mjs` | `PASS anthropic (model=claude-sonnet-4-6) chat=4 chars in 2796ms, stream=2 chunks/4 chars in 1894ms` → `live-provider-smoke: PASS (1/1 live providers reachable).` — **exit 0** |
| Red proof (deliberately invalid key) | `ANTHROPIC_API_KEY=<invalid> node …` | `FAIL anthropic (model=claude-sonnet-4-6) at liveCall: 401 … authentication_error` — **exit 1** |

The red proof matters: a smoke that cannot go red is an accidental-green smoke. The invalid-key run
proves the failure path actually fails.

- `npx vitest run scripts/harness/__tests__` → **72 files / 700 tests passed** (includes the 20 new ones).
- `node scripts/harness/run-all-scans.mjs` → **59 of 61 green**. The two failures are
  `build-contracts` and `dist`, both pure artefacts of this environment not having a full workspace
  build (I built only the provider dependency closure with `build:js`, so `.d.ts` files are absent).
  Unrelated to this change.
- Workflow YAML parses; the `workflow_dispatch` input is passed through an env var rather than
  interpolated into the shell line, so a dispatch input cannot become shell text.

### Owner action required — one repo secret

The CI half cannot be observed until credentials exist, so **HARNESS-024 stays `in-progress`, not
`done`**. Honest status: the script is proven, the workflow is a green no-op.

- **Required:** `ANTHROPIC_API_KEY` as a repo secret. That alone proves the CI half.
- Optional: `OPENAI_API_KEY`, `GEMINI_API_KEY`, `DEEPSEEK_API_KEY`, `DASHSCOPE_API_KEY`.
- OpenAI's definition ships no default model by design, so covering it also needs the repo
  **variable** `LIVE_SMOKE_MODEL_OPENAI`. Without it OpenAI is reported **WARN**, not FAIL, and the
  exit code is unchanged — a red nightly for a missing config value only trains people to ignore the
  signal.

After provisioning, one `workflow_dispatch` run showing `PASS <provider> (model=…)` closes the item
and archives it to `completed/`.

### Deferred: root `package.json` script

Root `package.json` is owned by another agent this wave, so the workflow invokes the runner directly
(`node scripts/harness/live-provider-smoke.mjs`). The line to add later:

```json
"harness:live-smoke": "node scripts/harness/live-provider-smoke.mjs"
```

### Scope note — backlog item 2

HARNESS-024 also lists "document the local/CI verification asymmetry". That is being covered by the
sibling `scripts/harness/verify-like-ci.mjs` work in this same wave, so it is deliberately not
duplicated here; the backlog file records that dependency.

### DIFF
--- a/.agents/backlog/HARNESS-024-live-provider-smoke.md
+++ b/.agents/backlog/HARNESS-024-live-provider-smoke.md
@@ -1,6 +1,6 @@
 ---
 title: 'HARNESS-024: env-gated 프로바이더 라이브 스모크 1콜 + 로컬/CI 검증 정렬'
-status: todo
+status: in-progress
 created: 2026-07-04
 priority: medium
 urgency: soon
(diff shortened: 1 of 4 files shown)


woojubb:
Follow-up commit `c5cb2788` fixes the CI `scans` failure (`harness-config-paths`, LESSON-006).

The smoke had two hardcoded `packages/*/dist/node/index.js` literals. `resolveBuiltEntry` now finds the package by its manifest `name` under `packages/` and reads the entry location from that package's own `exports['.'].node.import`, so the build layout stays owned by the package that produces it. Not built yet → `undefined` → the existing explicit "workspace is not built" message, never a silent pass. No allowlist marker was used.

**Why the local run missed it:** `harness-config-paths` only flags a quoted workspace path that does **not** resolve from the repo root. My worktree had `dist/` built (the smoke needs it to run at all), so both literals existed locally and the scan passed — in a fresh CI checkout there is no `dist/`, so the same literals are ghost paths. Local was state-dependent green; CI was structurally red. I re-verified the fix in a **dist-free clean clone** (the CI condition), where `harness-config-paths` passes and the unbuilt smoke exits 1 with the build instruction. This is a concrete case for the `harness:verify-like-ci` work: the divergence was not a skipped scan, it was local filesystem state the scan reads.

3 new `resolveBuiltEntry` unit tests cover built / declared-but-unbuilt / unknown-package. Harness suite: 703 tests green.

PR #1454, opened 2026-07-25

woojubb:
Built by `node scripts/harness/promote.mjs` — A1/A2/A3 hold, the merge is clean by construction, and the promoted tree is `develop`'s unchanged.

## What this promotes

Roughly 20 merges since the last promotion, dominated by two threads:

**Security.** SEC-004/005/006/007 closed the CodeQL alert backlog at source with **zero dismissals**, including two defects the analyser never flagged — a symlink escape from the file-tool sandbox (`checkPathWithinCwd` was a lexical check), and a studio server with no containment plus `Access-Control-Allow-Origin: *`, letting any visited website execute an arbitrary DAG on the machine. Also fixed: an unauthenticated path traversal via `resumeSessionId`, command injection in the GitHub Action, an SSRF, a shell-hijack via caller-supplied `PATH`, and a local DoS. The enumeration tools (`Glob`, `Grep`) were registered as module-level singletons and so could never be bound to a root — they now share one containment implementation with `Read`/`Write`/`Edit`.

**Gate integrity.** A run of findings where a check reported success over ground it never covered: `Claude review` had been silently skipping *every* run (100 for 100) on a workflow-parity mismatch; the required `scans` job exited 0 while printing `SKIPPED`; nine jobs read grafted history, and #1424 merged with three required checks reporting `skipping`; `protect-main`'s five required contexts were 3–6 second echoes while the job that actually verifies a promotion could not block; and a PR retarget fired an `edited` event the workflow did not listen for, leaving a feature branch one click from `main` past the ancestry gate. Each is now closed with a mechanical floor and a red-first proof.

Plus **TypeScript 7**: `pnpm typecheck` 30.0 s → 6.2 s (4.8×) with byte-identical diagnostics across all 99 workspace projects.

## Merge method

**`gh pr merge <n> --merge`. Never `--squash`** — `protect-main` now rejects squash outright, and a squash would leave `main` outside develop's ancestry and reproduce the recurring conflict INFRA-051 exists to prevent.


woojubb:
Superseded — rebuilt from a newer `develop` that now also carries HARNESS-050, INFRA-056 and the PERF-004 follow-up.

PR #1460, opened 2026-07-26
perf(harness): PERF-005 phase 1 — remove every avoidable use of the legacy TypeScript compiler

woojubb:
Executes **phase 1** of `.agents/backlog/PERF-005-remove-legacy-typescript.md`. After this,
**no first-party code in the repository imports `typescript@5.9.3`**. The package stays installed
because `@typescript-eslint`'s `typescript-estree` imports it at runtime — that is phase 2, gated
on upstream and explicitly not in scope here.

## What landed

**1. The four TS-API scans moved to `@typescript/native-preview`, behind one adapter**
(`scripts/harness/lib/ts-ast.mjs`). The indirection is required, not stylistic: the API is
namespaced `unstable/` with no stability guarantee, the pin is a dated dev build, and three of the
four scans gate CI.

**The parse entry point, since the item guessed wrong about it.** `unstable/ast` exports 409
symbols and **no parser** — only `createScanner` (tokens, no tree) and a factory `createSourceFile`
that constructs a node from already-parsed statements. The parser lives in the Go binary. The route
that does work is `unstable/sync` with a virtual filesystem at a path no tsconfig covers, which puts
the source in tsgo's *inferred project* (`getDefaultProjectForFile` → `/dev/null/inferred`) — parsed
and bound without loading any configured project. On 400 real files the native walk visits
**239,234 nodes, exactly the legacy count**.

Four genuine API differences, all reconciled at the swap point rather than by weakening a caller:

| Difference | Native form | Handling |
| --- | --- | --- |
| No standalone parser | `unstable/sync` + virtual FS + inferred project | adapter's `createSourceFile` |
| `ImportClause.isTypeOnly` undeclared | `phaseModifier === SyntaxKind.TypeKeyword` | `isTypeOnlyImportClause()` |
| Array holes | nameless `BindingElement` (legacy: `OmittedExpression`) | skipped in `bindPattern` |
| Three guards renamed | `isParameterDeclaration` etc. | aliased to legacy names |

An import-time assertion fails loudly if the pinned build stops exporting a re-exported guard — a
missing guard destructures to `undefined` and only throws on the branch that uses it, which is
precisely how a scan gets quietly weakened.

**2. `apps/action` builds with `tsgo`** (`typecheck` already did). **Byte-identical emit** across all
8 output files, md5-compared. Its 6 tests pass with `typescript` uninstalled from the package.

**3. `knip` removed.** Verified unwired before removing: its only live references were the script,
the devDependency and `knip.json`; **zero** matches across all 14 workflows, `.husky/` and
`scripts/`. Lockfile diff is **240 lines, pure deletions**, no additions.

**4. `scan-legacy-typescript.mjs`** — token prefilter plus AST confirmation on the module specifier,
so `@typescript-eslint/*`, `@typescript/native-preview` and prose do not false-positive. Follows
`scan-no-fallback`'s anti-rot conventions, and additionally implements **stale**-suppression
detection (which `scan-no-fallback` deferred) because the matched construct here is exact.

## Verification

**Finding-set equivalence — the check that mattered.** Each scan's detectors were run over the whole
repo (2,699 files, far wider than their production scopes, so a silent regression would show),
before and after, and diffed:

| Scan | Finding set | Before vs after |
| --- | --- | --- |
| `scan-interface-runtime` | 3,624 findings | identical |
| `check-spec-public-surface` | 24 packages / 602 names | identical |
| `scan-composition-neutrality` | 1,044 IO + 30 conditionals | identical |
| `audit-implements` | 1,832 files / 340 classes | identical |

`diff -rq` over all five captured JSON files: **no differences**, sha256-equal, 33,259 lines of
findings. Re-confirmed after formatting.

**Red-first, per scan** — every defect each exists to catch was reintroduced and still fires:
9 planted purity violations (with the legitimate `import type` forms correctly staying silent); the
full HARNESS-048 evasion probe, **all 12 caught**; both spec-surface edges, with type-only exports
correctly excluded; and the planted class/heritage facts with correct lines and flags.

**The new guard** is RED against the pre-migration tree — restoring the four original imports gives
exactly four findings with correct line numbers — and GREEN after. All six finding kinds were driven
to fire and to clear.

`pnpm harness:verify-like-ci` **PASS, all 11 stages**. `pnpm build`, `pnpm typecheck`, `pnpm lint`
(0 errors), `pnpm harness:scan` (**69/69**), harness tests (**1,075**) all green.

## Two things this discovered

- **The native package does not export a parser** — the item (and PERF-003/004) recorded that it
  did. Corrected, with the working route documented.
- **`typescript` is declared by 99 workspace manifests, not one.** None is a code consumer, but
  deleting 98 entries is its own change with its own blast radius, and nearly all are outside this
  item's scope. They are frozen as a path ratchet that may only shrink, so the edge stays on at
  today's boundary. **Phase 2 is larger than "delete one devDependency" and must absorb this.**

## Not landed

Phase 2 (removing the package) is unchanged and still gated on `@typescript-eslint` dropping its
runtime import. The backlog item stays in `.agents/backlog/` rather than `completed/` for that reason.

Two CI stages `verify-like-ci` flags as relevant but cannot mirror locally, both from the manifest
and lockfile edits: `security-audit` (needs network + osv-scanner) and `windows-shell` (needs a
Windows host). The lockfile change is deletions only, so no new package entered the tree.


woojubb:
## Re-verified after rebasing onto current `develop`

`develop` moved while this was in flight (#1459 landed the PERF-005 backlog item itself, an add/add
collision with this branch's copy). Resolved by rebasing onto `47353fba6`; `origin/develop` is now an
ancestor of this branch and GitHub reports **MERGEABLE**.

**Nothing was dropped in the resolution.** The `run-all-scans.mjs` diff against `develop` is `+4/-0`
— purely the new `legacy-typescript` registration — and `scan-unearned-done-claims` (HARNESS-050,
#1456) is present and untouched. The `package.json` diff is the two intended `knip` removals plus the
one new script line, nothing else.

**Finding-set equivalence re-run on the post-rebase tree**, both sides captured fresh against the
same 2,699 files — the legacy-`typescript` scans taken from `origin/develop` on one side, the
migrated ones on the other:

| Scan | Finding set | Post-rebase before vs after |
| --- | --- | --- |
| `scan-interface-runtime` | 3,624 findings | identical |
| `check-spec-public-surface` | 24 packages / 602 names | identical |
| `scan-composition-neutrality` | 1,044 IO + 30 conditionals | identical |
| `audit-implements` | 1,832 files / 340 classes | identical |

`diff -rq`: **no differences**. All five sha256 hashes match across 33,259 lines of findings:

```
d3be817e32c5b065cadcbb84317a1de9295104b0688aaedc760ffd0b94d77c91  1-interface-runtime.json
010444a947dd17ee0a1a085ce5762cb45534b575457db0784104d71d318440aa  2-spec-public-surface.json
76a7818fcc01026fe99b01decc3b79efe0810bf68646d7f184e18adfa1471b17  3-composition-conditionals.json
d67c12c970d9f594d530ddd9a439a17ff4f90109c187585e2246c7315da56713  3-composition-io.json
f2983a484f465faed85cfbdc3b0f62acaedc88a764deeeb0a10460e0b2931dcf  4-implements-audit.json
```

`pnpm harness:scan` 69/69 and `pnpm harness:verify-like-ci` PASS on all 11 stages, both re-run after
the rebase.

## Two findings moved into the backlog item itself

Both were previously only in this PR body. They are now in
`.agents/backlog/PERF-005-remove-legacy-typescript.md`, and — importantly — in the **opening
section**, not just the outcome section at the bottom, because the opening is what someone scoping
phase 2 reads first:

- **`typescript` is declared by 99 workspace manifests, not one root devDependency.** Phase 2 is 96
  baselined entries plus the root, not one line. The item's own "Note on cost" section still says
  "a devDependency", singular; the correction now sits above it and the Phase 2 section points at it.
- **`unstable/ast` exports no parser.** PERF-003, PERF-004 and this item all recorded that it did.
  The correction and the working route (`unstable/sync` + virtual FS + the inferred project via
  `getDefaultProjectForFile`) are documented at length in the header of
  `scripts/harness/lib/ts-ast.mjs` — the file the next reader actually opens — with a pointer to it
  from the backlog item.

## Note on the `Claude review` check

It failed with `Reached maximum number of turns (25)` — an exhausted turn budget, not a finding; it
produced no review output. It is **not** a required check for `develop` (the eight required are
build, quality, scans, security audit, commitlint, tui-e2e, examples-typecheck, windows-shell — all
green on the previous push, including `review-gate`). Flagging rather than papering over it: this
looks like the class INFRA-053 tracks.

PR #1472, opened 2026-07-26
DO NOT MERGE — INFRA-053 promotion route is blocked by promotion-ancestry (use #1473)

woojubb:
## DO NOT MERGE. This route is not viable — use #1473 instead.

This PR was opened to follow INFRA-053's prescribed sequence: cut `release/*` from `develop`, make
the edit, PR → `main` (parity-exempt), merge, then back-merge. **Executing it proved that sequence
is mechanically impossible.**

`scan-promotion-ancestry` (INFRA-051) is a **required** status check on `protect-main`, runs on any
PR whose base is `main`, and rejects this PR by construction:

```
[A1] `origin/main` is NOT an ancestor of the promotion head.
[A2] the promotion head carries 4 non-merge commit(s) that `origin/develop` has never seen:
       10df12c3b docs(backlog): INFRA-053 record the measurements, the scoping answer, …
       cf7411840 fix(ci): INFRA-053 state the shell constraint in the prompt …
       9e26072c6 fix(ci): move the review config's rationale out of the claude_args block scalar
       86a41412c fix(ci): INFRA-053 grant the reviewer the tools its prompt needs …
[A3] the promotion changes 2 file(s) relative to the `origin/develop` commit it was cut from
     A promotion must promote develop's tree UNCHANGED.
```

A1 is fixable (`git merge --no-ff origin/main`). **A2 and A3 are not** — they say a promotion may
carry nothing `develop` has not already integrated, which is precisely what "make the edit on the
release branch" requires it to carry.

So the two guards contradict each other:

- `scan-review-workflow-parity` (INFRA-048) exempts only a PR whose base **is** `main`, which is why
  INFRA-053 prescribed going to `main` first;
- `scan-promotion-ancestry` (INFRA-051) forbids a PR into `main` from containing anything not
  already on `develop`, which is why it cannot.

`claude-code-review.yml` can therefore only reach `main` **through `develop`**, by an ordinary
promotion. That is what **#1473** does.

The full measurement, the tool grant and its justification, the live run on #1434 that posted
findings, and the remaining blocker all live in **#1473**. Please close this one, or leave it as the
record of why the prescribed sequence was abandoned.

Refs INFRA-053, INFRA-048, INFRA-051

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AhJjnpqzCortHNB3h15h1Z

PR #1434, opened 2026-07-25
fix(ci): INFRA-048 close four gates that report success without running

woojubb:
Closes the "a gate exists but does not actually gate" class. All four items share one root cause:
**a check that reports success when it did not run.**

## A — review findings now reach the merge decision (INFRA-048)

Investigating the item turned up something worse than "reports pass regardless of findings":
**`Claude review` was not reviewing at all.** `anthropics/claude-code-action` validates that its
workflow file is byte-identical to the default branch's copy and, when it differs, prints
`Exiting due to workflow validation skip` and **exits 0** — the job reports `success`. A
`actions/checkout@v4`→`@v7` bump merged to `develop` alone (#1313, 2026-07-24) broke that identity
by one line.

| evidence | value |
| --- | --- |
| `gh run list --workflow=claude-code-review.yml --limit 100` | `success: 100` of 100 |
| three runs inspected in full (#1432 / #1423 / #1421) | all three: validation skip, 13–21 s |

Two mechanical halves:

**It is a recurring window, not a one-off.** It opens whenever that file changes on `develop` and
closes silently on the next develop→main promotion. It closed mid-investigation: promotion #1427
merged at `17:27:54Z` carrying `@v7` to `main`, restoring parity by itself. Nothing announced either
edge. So this PR carries **no net change** to that workflow — the correction the defect needs is the
guard, not the byte.

1. **`scan-review-workflow-parity`** — registered in `run-all-scans`, so it runs inside the
   **required** `scans` job. Every workflow invoking that action must match the default branch's
   copy; governed workflows are discovered by the action they invoke, never by filename. Fails
   closed if the default branch cannot be read; not applicable to the promotion PR that restores
   parity.
2. **`review-gate`** — a distinct check that READS the review output.

| class | rule | blocks? |
| --- | --- | --- |
| blocking | introduced by this PR, severity `error` or security high/critical | **yes** |
| advisory | any other alert introduced by this PR | no — printed on the check |
| pre-existing | already open on the base branch | no |
| unavailable | analysis incomplete / API error / unparseable | **yes** |
| acknowledged | PR carries `review-findings-acknowledged` | no — override recorded on the PR |

On a block it also runs `gh pr merge --disable-auto`, because a red **non-required** check does not
stop an armed auto-merge — the exact #1409 hole.

**The trade-off.** A gate that hard-fails on any finding blocks on NITs and gets bypassed, which is
worse than advisory. This repo makes that measurable: **every one of its ~100 open code-scanning
[...]

PR #1473, opened 2026-07-26
fix(ci): INFRA-053 make Claude review actually review — denials 17 → 0, cap stays 25

woojubb:
## This is now the PRIMARY PR. #1472 (→ `main`) is **not viable** — see the bottom of this body.

`Claude review` has been failing `error_max_turns` on nearly every PR since INFRA-048 made it
honest. **The turn cap was never the root cause**, and **it did not need raising.**

### The denial count, at every step — that is the metric

| Configuration                            | PR                    | Turns | **Denials** | Posted                     |
| ---------------------------------------- | --------------------- | ----- | ----------- | -------------------------- |
| as shipped (no `allowed_tools`)          | #1470                 | 26 ✗  | **17**      | nothing                    |
| tool grant only                          | #1472 (150 files)     | 26 ✗  | **13**      | nothing                    |
| tool grant + shell constraint stated     | #1472 (150 files)     | 20 ✓  | **1**       | summary                    |
| tool grant + shell constraint stated     | **#1434**             | 11 ✓  | **0**       | **2 inline + summary**     |

Row 1 is CI run [30189528981](https://github.com/woojubb/robota/actions/runs/30189528981). Rows 2–4
are local reproductions of the same SDK invocation the action performs in agent mode, with the
action's **own** `github_inline_comment` MCP server attached and its **own** post-session buffer
flush (`src/entrypoints/post-buffered-inline-comments.ts`).

### What the reviewer was actually denied

Reading `anthropics/claude-code-action@v1.0.183`, the failure was worse than a budget problem. A
`prompt:` input selects **agent mode**, and in agent mode:

- the prompt is passed **verbatim with no PR context injected** (`src/modes/agent/index.ts`) — the
  reviewer was never told which repository or which PR it was reviewing, and no diff was fetched
  for it;
- a GitHub MCP server is installed **only if one of its tools is named in `--allowedTools`**
  (`src/mcp/install-mcp-server.ts`). With no `--allowedTools` at all, `github_inline_comment` was
  never started — the reviewer had **no way to post a finding even if it found one**. The run log's
  `No buffered inline comments` is that fact, recorded.

So it was asked to read a diff it could not fetch and to post comments through a server that did not
exist, and spent its budget being refused.

### What is granted, and why that is the minimum

```
--allowedTools "Bash(gh pr diff:*),Bash(gh pr view:*),Bash(gh pr comment:*),mcp__github_inline_comment__create_inline_comment,Read,Grep,Glob"
--disallowedTools "Write,Edit,NotebookEdit,WebFetch,WebSearch"
```

| Granted                                       | Why the prompt cannot work without it                      |
| --------------------------------------------- | ---------------------------------------------------------- |
| `Bash(gh pr diff:*)`                          | the only way agent mode can obtain the diff                |
| `Bash(gh pr view:*)`                          | PR title/body — the intent a diff is judged against        |
| `Bash(gh pr comment:*)`                       | the required ≤3-line summary                               |
| `mcp__github_inline_comment__create_inline_…` | inline findings — **and it is what installs the server**   |
| `Read`, `Grep`, `Glob`                        | context around a hunk, and the rules docs the prompt cites |

A reviewer of PR-authored content reads and comments and nothing else: no `Write`, no `Edit`, no
general `Bash`, no network fetch. `Bash(git diff:*)` is deliberately **not** granted — at depth 1
there is no base ref for it to resolve, so it would only invite wasted turns.

**Two rounds of measurement were needed**, which is the part worth keeping. The tool grant alone
only moved 17 → 13: every remaining denial was shell plumbing (`gh pr diff N > file`, `… | awk …`,
`git fetch`) that a `Bash(gh pr diff:*)` prefix rule does not cover. Broadening the grant would have
traded away the point of a read-only reviewer. Stating the constraint in the prompt cost nothing and
took it to 1, then 0.

### `fetch-depth: 1` is kept — deliberately

The action's own PR-review examples keep `fetch-depth: 1` and obtain the diff **from the GitHub API
via `gh pr diff`**, not from local git. There is no base ref to want, so a deeper checkout on this
monorepo would buy only clone time. The file now says so, to stop the next reader "fixing" it.

### `--max-turns` stays at 25, and the scoping question is answered

It never needed raising: 20 turns on a 150-file PR, 11 on #1434. What the cap really bounds is the
number of **findings** — `create_inline_comment` posts one comment per call, so one finding costs one
turn. So the prompt caps inline comments at 10 in severity order, pushes the rest into the summary,
and posts the summary first once fewer than ~5 turns remain. **Bound the output, not the budget.**

### Proof — a live run that posted findings

Against **#1434** itself, the PR named in INFRA-053's acceptance criterion: 0 denials, 11 turns,
completed, and it posted two inline findings plus a summary.

- [discussion_r3652038352](https://github.com/woojubb/robota/pull/1434#discussion_r3652038352) —
  `review-gate.yml` disarms auto-merge with only `contents: read`, and `|| echo "auto-merge was not
  armed"` swallows the permission failure as if it were a normal no-op. **That is INFRA-057**, which
  this repository later found and fixed separately in #1467. The reviewer found it unprompted, from
  the diff.
- [discussion_r3652038576](https://github.com/woojubb/robota/pull/1434#discussion_r3652038576) — the
  code-scanning wait runs unconditionally, so a docs-only PR waits 15 minutes and is then blocked as
  `verdict-unavailable`.

The reviewer also raised a MUST against **this PR's own first commit** (the `#` lines inside the
`claude_args` block scalar). It turned out not to be a live bug — the action's `stripShellComments`
does drop them, verified by running `parseAllowedTools`/`parseSdkOptions` over the file — but the
rationale moved out to real YAML comments anyway, in `9e26072c6`. That is the reviewer earning its
keep on the change that fixes it.

---

## Why this PR, and not #1472

INFRA-053 prescribes: cut `release/*` from `develop`, edit, PR → `main` (parity-exempt), merge,
back-merge. **That sequence is mechanically impossible under the current gates.**
`scan-promotion-ancestry` (INFRA-051) is a **required** check on `protect-main` and rejects it by
construction:

```
[A2] the promotion head carries 4 non-merge commit(s) that `origin/develop` has never seen
[A3] the promotion changes 2 file(s) relative to the `origin/develop` commit it was cut from
     A promotion must promote develop's tree UNCHANGED.
```

A promotion PR may contain nothing `develop` has not already integrated — which is exactly what a
"make the edit on the release branch" sequence requires it to contain. So `claude-code-review.yml`
can only ever reach `main` **through `develop`**, by a normal promotion.

That leaves one honest route, which is this PR:

1. Merge this into `develop`.
2. Promote immediately: `node scripts/harness/promote.mjs` → PR to `main` → merge. Parity restored.

## Both remaining blockers, stated plainly

**1. This PR's `scans` / `quality` are red, and only for the parity guard.** Measured: **1 failed /
1126 passed**. The one failure is `scan-review-workflow-parity.test.mjs:130`:

```js
it('holds on the real repository', () => {
  const { findings } = findReviewWorkflowParityFindings(REPO_ROOT);
  expect(findings).toEqual([]);
});
```

The scan **CLI** honours `isPromotionToDefault()`; its **test does not**, and `pnpm harness:test`
runs unconditionally in both `scans` and `quality`. There is no exemption for the PR that
*introduces* the change either, so **the guard blocks every possible route** — to `develop` because
the copies differ, to `main` because promotion-ancestry forbids the shape.

As it stands, `claude-code-review.yml` cannot be modified by anyone while keeping CI green. Merging
this needs either an admin bypass on `scans`/`quality`, or a guard fix first.

**2. An option for closing the window permanently, found but not taken.** The parity validation runs
inside `setupGitHubToken()` → `exchangeForAppToken()` (`src/github/token.ts`) and is **skipped
entirely when a `github_token` input is supplied**. Passing
`github_token: ${{ secrets.GITHUB_TOKEN }}` would stop the action ever silently skipping — but
`scan-review-workflow-parity` would still fail on divergence, so it closes the window only if the
scan is retired or rewritten with it. That is a `scripts/**` change and a security-posture decision,
so it is recorded rather than taken.

Both fixes live in `scripts/**`, which was outside this task's owned scope.

Refs INFRA-053, INFRA-048, INFRA-051, INFRA-057


woojubb:
Superseded by #1478 — its four commits are cherry-picked verbatim there, so the allowed_tools fix landed rather than being lost. Confirmed independently by the ci.yml audit: the cap was never the constraint (26 vs 25 with 8 then 6 denials); with the grants in place the review finishes in 10–13 turns.

PR #1476, opened 2026-07-26
ci: INFRA-058 throwaway probe — do not merge

woojubb:
**THROWAWAY. DO NOT MERGE — will be closed as soon as it is read.**

Measures the one property the INFRA-058 D3 fix (#1474) asserts rather than measures: that
`!cancelled() && (needs.changes.result != 'success' || ...)` makes the five `needs: changes`
dependents RUN when `changes` FAILS, rather than reporting `skipping` — which branch protection
accepts, and which is how #1424 merged with three required checks skipped.

The `changes` step fails before writing `code=` to $GITHUB_OUTPUT, so dependents see
`result=failure` and an EMPTY `outputs.code` — the exact shape a checkout failure, a runner OOM,
or a syntax error in the classifier would produce.

Expected: `tui-e2e`, `examples-typecheck` and `windows-shell` all RUN and report a real
conclusion. Pre-fix, all three reported `skipping`.

### DIFF
--- a/.agents/backlog/INFRA-058-ci-yml-job-by-job-audit.md
+++ b/.agents/backlog/INFRA-058-ci-yml-job-by-job-audit.md
@@ -1,0 +1,223 @@
+---
+title: 'INFRA-058: ci.yml job-by-job audit — three fail-open paths closed, four fidelity gaps filed'
+status: in-progress
+created: 2026-07-26
+priority: high
+urgency: now
+area: .github/workflows/ci.yml, scripts/harness
+depends_on: []
+---
+
+## Problem
+
+`.github/workflows/ci.yml` carries 14 jobs and every required status check on `protect-develop`,
+plus three of the four on `protect-main`. In two days it produced five separate incidents in which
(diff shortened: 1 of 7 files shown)


woojubb:
Measurement complete — closing as planned. Recorded in INFRA-058 / #1474.

From the jobs API (decisive, because a skipped job has no `started_at`):

    changes:            completed   failure  started=08:37:42Z
    windows-shell:      completed   success  started=08:37:52Z
    tui-e2e:            in_progress null     started=08:37:53Z
    examples-typecheck: completed   success  started=08:37:51Z

All three REQUIRED dependents started 9-11s AFTER `changes` had completed as `failure` — GitHub evaluated the fail-safe `if:` against a failed dependency and dispatched them. The advisory pair (`patch-coverage`, `regression-red-proof`) also ran and passed, so all five `needs: changes` dependents are covered.

Control in the same run: the three main-only jobs, genuinely excluded by their own `if:`, reported `skipping`.

Pre-fix, this is the run in which three required checks would have reported `skipping` and the PR would have been mergeable — the #1424 shape.

PR #1487, opened 2026-07-26
fix(ci): INFRA-060 D5 rename the `security audit` required check to `dependency audit`

woojubb:
INFRA-060 **D5**. Renames the `security audit` required status check to `dependency audit`, on the repository side. The ruleset is the owner's to change; the sequence is below.

## The defect

The context name read as "this repository was audited for security". What the job does is detect a package-manifest/lockfile diff and run `osv-scanner` over `pnpm-lock.yaml` — known-vulnerable **dependencies** only. It inspects no first-party source: a command injection, a leaked credential or a permissive CORS policy all get a green check. CodeQL, `Secret scan (gitleaks)`, `Dependency review` and `review-gate` are the checks that cover those, and none of them is this job.

A reader of the check list — which is what a human consults when deciding whether a merge is safe — was told something broader than what was verified. `dependency audit` matches the behaviour exactly, and it makes visible that the security surface is covered by several checks rather than by this one.

## What changed

| File | Change |
| --- | --- |
| `.github/workflows/ci.yml` | job id `security-audit` → `dependency-audit`, `name:`, two `::error::` prefixes, three comments |
| `.github/required-status-checks.json` | context + job + `verifies:` prose (now states what it does **not** cover), the `main` `deliberately_not_required` entry, the historical `why` |
| `scripts/harness/ci-mirror-map.mjs` | `NOT_MIRRORED` context + the `ci.yml → dependency-audit` pointer |
| `scripts/harness/scan-main-required-checks.mjs`, `check-review-gate.mjs` | docstrings |
| 4 test files | `verify-like-ci`, `harness-scripts`, `scan-required-check-needs`, `release-run` |
| `.agents/rules/git-branch.md`, `.github/DEPENDABOT-DISABLED.md` | live references |
| `HARNESS-052`, `INFRA-061`, 2 memory docs | live references |

No `needs:` edge referenced this job (verified by YAML parse).

## Proof, not assumption

**The pins were run against a half-renamed tree** rather than trusted:

| Pin | Noticed? |
| --- | --- |
| `ci-mirror-map.test.mjs` | **yes** — failed 3 ways, bidirectional, and resolves `job:` → a real ci.yml job |
| `harness-scripts.test.mjs` | **yes** — reads ci.yml, asserts the job id |
| `verify-like-ci.test.mjs` | **yes** — 3 assertions on the context string |
| `scan-main-required-checks` offline | **no** — **finding**, see below |
| `scan-main-required-checks --live` | **yes** — reports the drift in both directions |

**Finding.** `scan-main-required-checks` offline is quantified over `branches.main` only, so a `develop` context pointing at a job that no longer exists is invisible to it. Documented behaviour rather than a bug — but it is the artefact whose name and docstring make a reader expect that coverage, and the develop-side rename is in fact guarded entirely by one test in a differently-named file. Recorded in INFRA-060.

**`--live` is expected RED until the ruleset is patched**, and it is:

```
  - security audit: the LIVE `develop` ruleset requires it, but .github/required-status-checks.json
    does not declare it under `branches.develop` — so nothing has checked that it is covered.
  - dependency audit: .github/required-status-checks.json declares it required on `develop`, but the
    LIVE ruleset does not require it — it is enforcing nothing.
```

**Post-rename grep.** Every surviving `security audit` / `security-audit` mention is either a frozen fixture (`INFRA-055-pre-correction.md.txt`, a byte-for-byte reproduction of a historical document), a completed GATE record under `spec-docs/done/**` or `backlog/completed/**`, a CHANGELOG entry, a deliberate `(then named …)` note, or one of two **comments** in `dependency-review.yml` / `security-scheduled.yml` — workflows this branch does not own, routed to INFRA-061. None is a `needs:` edge or a context pin.

## Verification

- `pnpm harness:scan` — **78/78** on a built tree
- `pnpm harness:test` — **96 files / 1286 tests**
- `pnpm harness:verify-like-ci` — **PASS, all 11 stages**
- `pnpm build`, `pnpm typecheck` — exit 0
- YAML parse of `ci.yml` — 14 jobs; every declared context resolves to a job whose `name:` equals it

## Merge sequence — the window matters

A required context that does not exist reports `expected` forever and blocks every open PR. The queue was verified empty (**0 open PRs**) immediately before pushing.

1. Merge this PR to `develop`.
2. **Owner patches the ruleset** (`protect-develop`, id `18715844`) — the exact call is in the review comment.
3. Confirm: `node scripts/harness/scan-main-required-checks.mjs --live` → green.

Between 1 and 2 the old context is required but never reported. Keep the gap short and open no PR inside it.

## Also decided: D7

Filed as **INFRA-063**, not folded in — it is not the same shape. D7's over-claim is in the declaration **prose**, not in a required-context name, so it needs **no ruleset change**. And it is not only prose: `packages/agent-cli-web` declares a `test:e2e` script, which `pnpm run -r --if-present test` never matches, so a suite that exists is never run by the release gate — the identical case as `agent-cli`'s `test:bin`, which was recognised once and appended by hand rather than generalised.


woojubb:
## The ruleset call — yours to make, step 2 of 3

**Sequence:** merge this PR → **then** this call → then confirm with `--live`.

Not before the merge: renaming the ruleset context first leaves `dependency audit` required while
only `security audit` reports, which blocks this very PR.

`PATCH /repos/{owner}/{repo}/rulesets/{id}` replaces the whole `rules` array, so this reads the live
ruleset and rewrites exactly one string, preserving both rules and every other field:

```bash
gh api repos/woojubb/robota/rulesets/18715844 \
  --jq '{rules: [.rules[] | if .type == "required_status_checks"
          then .parameters.required_status_checks |= map(
                 if .context == "security audit" then .context = "dependency audit" else . end)
          else . end]}' > /tmp/protect-develop-rules.json

# inspect it before sending
cat /tmp/protect-develop-rules.json | python3 -m json.tool

gh api --method PATCH repos/woojubb/robota/rulesets/18715844 \
  --input /tmp/protect-develop-rules.json
```

Dry-run output verified — the payload keeps `required_status_checks` (9 contexts, only the 4th
changed) and `non_fast_forward`, with `strict_required_status_checks_policy: false` and
`do_not_enforce_on_create: false` intact:

```
build, quality, scans, dependency audit, commitlint, tui-e2e, examples-typecheck, windows-shell, review-gate
```

`protect-main` needs no change — it does not require this context.

**Step 3, confirm:**

```bash
node scripts/harness/scan-main-required-checks.mjs --live   # expect exit 0
```

If it still reports drift, the PATCH did not land — re-read the ruleset before retrying.

PR #1502, opened 2026-07-27
fix(ci): INFRA-061 verify the osv-scanner download's checksum before executing it

woojubb:
## Summary

`ci.yml` downloaded a GitHub release asset and executed it, in **two** jobs, with no integrity check —
including inside the **required** `dependency audit` job. Pinning the version does not pin the bytes: a
release asset can be replaced under an existing tag. That made it arbitrary code execution on a runner
with the repository checked out, in the job whose entire purpose is to prove the tree is clean.

Both call sites now `sha256sum -c -` **before** `chmod +x`, reading `OSV_SCANNER_VERSION` and
`OSV_SCANNER_SHA256` from one workflow-level `env` block so the two cannot drift apart. The digest was
recomputed from the asset independently (41,324,728 bytes) and matches the pin already in
`security-scheduled.yml` — one accepted-risk SSOT for the same scanner.

This is INFRA-061 owner item 3. On re-reading, it was never a contract decision: verifying a download
changes nothing about what gates a merge, what a check is named, or when it runs, and a correct
artifact produces an identical result. It was only an ownership blocker.

## Falsification — a green run is not evidence

The two step bodies were extracted from the **parsed** `ci.yml` (never re-typed, so the thing under
test is what CI will run) and executed with the download shimmed to serve either the genuine asset or
the same bytes plus one appended `\0` — the smallest possible supply-chain swap:

| `ci.yml` revision | artifact | exit  | left executable |
| ----------------- | -------- | ----- | --------------- |
| pre-fix (develop) | tampered | **0** | **yes**         |
| post-fix          | tampered | **1** | no              |
| post-fix          | genuine  | **0** | yes             |

The first row is the proof the test is not accidentally green: the pre-fix body accepted the swapped
binary, marked it executable and ran it. Both osv-scanner steps were exercised in every row.

The genuine post-fix run went all the way through the real scan:
`Scanned … pnpm-lock.yaml and found 2130 packages / Filtered 5 vulnerabilities / No issues found`.

## Scope

- No ruleset edits, no required-check changes, no new backlog IDs, no branch deleted.
- This was the **last** unverified executable download in `.github/workflows` — `gitleaks.yml` and
  `security-scheduled.yml` already verify theirs.

## Test plan

- YAML parse (PyYAML) of all 13 workflows — all OK.
- Falsification table above, both steps, three rows.
- `pnpm harness:scan` — **all 81 scans passed**.
- `pnpm harness:test` — **1436 tests / 99 files passed**.

## Follow-up (not taken here)

A repo-wide guard — "a workflow that curls an executable must verify a checksum" — was deferred in the
INFRA-061 audit only because it would have failed immediately on `ci.yml` and turned the required
`scans` job red. That reason is now retired. The guard is not added in this PR because
`scripts/harness/` is outside this change's ownership.

### DIFF
--- a/.agents/backlog/INFRA-061-security-and-scheduled-workflow-audit.md
+++ b/.agents/backlog/INFRA-061-security-and-scheduled-workflow-audit.md
@@ -143,6 +143,45 @@ recorded them honestly rather than pinning them as sound:
 workflow it reports nothing to fix, the same answer it gives for a correct one. Recorded unfixed in
 `PENDING_CLASSIFICATION` and owned by INFRA-060 (the ci.yml audit), not silently promoted.
 
+### 8. `ci.yml` ran an unverified downloaded binary, twice (filed as owner item 3, then taken)
+
+Both osv-scanner call sites in `ci.yml` — including the one inside the **required** `dependency audit`
+job — did `curl … | chmod +x | execute` with no integrity check. Pinning the version does not pin the
+bytes: a GitHub release asset can be replaced under an existing tag. That is arbitrary code execution
+on a runner with the repository checked out, in the job whose entire purpose is to prove the tree is
+clean.
+
+Filed for the owner originally on the reasoning that `ci.yml` was owned by another agent, not that it
+was a contract decision. It is not one: verifying a download changes nothing about what gates a merge,
+what a check is named, or when it runs. A correct artifact produces the identical result. So it was
+taken once `.github/workflows/` ownership allowed.
+
+Both sites now `sha256sum -c -` **before** `chmod +x`, reading `OSV_SCANNER_VERSION` and
+`OSV_SCANNER_SHA256` from one workflow-level `env` block so the two cannot drift apart. The digest was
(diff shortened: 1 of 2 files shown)


woojubb:
### This PR's own CI does not exercise the changed step — stated so the green is not over-read

Measured on run `30268833586`:

```
dependency audit
  success  Detect dependency graph changes
  skipped  Vulnerability scan (osv-scanner) for dependency graph changes   ← the changed step
```

The step is gated on `changed == 'true'`, and this PR touches no `package.json` / `pnpm-lock.yaml`.
The second changed step lives in `release-grade verification`, which is `base_ref == 'main'` only and
reports `skipping` here.

So **both** modified steps are dormant on this PR. `dependency audit: pass` is a true statement about
the gate, and says nothing about the checksum verification. The evidence for the fix is the local
falsification in the description — step bodies extracted from the parsed workflow, executed against a
tampered and a genuine artifact, with the pre-fix revision run as the red row.

The change gets its first real CI exercise on the next PR that edits a manifest or the lockfile, and on
the next promotion to `main`.

PR #1508, opened 2026-07-27

woojubb:
Built by `scripts/harness/promote.mjs` — A1/A2/A3 ancestry gates hold, merge clean, develop's tree promoted unchanged.

## The one that matters most

**A tool refusal that leaves the bytes inside the sandbox is not a refusal** (#1505). `EditCheckpointStore.captureFile` ran on the raw tool argument *before* the contained tool decided anything, with no containment of its own. So `Write { filePath: "/etc/shadow" }` copied that file into `<cwd>/.robota/checkpoints/…`, **then** returned "Access denied" — leaving it inside the sandbox for a later in-sandbox `Read`. Arbitrary read through the tool that refused it, on by default in interactive sessions, and under `acceptEdits` with no human in the loop.

Proven end-to-end before the fix. The two new tests fail without the guard (direct path **and** symlink); a third pins that an in-sandbox file still gets its snapshot and passes in **both** directions, so containment was not bought by breaking what checkpoints are for.

## Also landing

**`SEC-007`** (#1504) — sections A and B were already merged; verification found a **third** `Glob`/`Grep` registration site still handing back uncontained singletons, reachable from any `.dag.json`, and it was the way *around* its own siblings.

**`INFRA-061`** (#1502) — `ci.yml` downloaded and executed `osv-scanner` with no checksum, twice, at call sites required on both rulesets. Digest verified against the real asset; a one-byte tamper is rejected. **This PR's own CI does not exercise either step** — the local falsification is the evidence, not the green tick.

**`HARNESS-052`** (#1506) — 7 of 14 boxes: all depth-1 `packages/` walkers onto the SSOT enumerator, the four presence-of-a-string guards made structural, `ratchet-tighten` promoted to a failure, 27 vacuous finders made fail-closed (guards proven fail-closed **by execution**: 15 → 43).

**`HARNESS-049`** (#1503, #1507) — deferred increments closed, and the folder/status floor registered and green. Five documents corrected from their own recorded `[GATE-COMPLETE] — ✅ PASS`; `DATA-002` is a recorded exception under anti-rot.

## Two things stated rather than implied

`DATA-002` shipped all three phases with **no GATE-COMPLETE entry at all**. Setting `status: done` would manufacture a completion no gate recorded; moving it to `active/` claims work is in progress that shipped in July — and the tree said so (`spec-research` went red). It waits on the owner running that gate.

I merged #1503 before reading its review, which carried a MUST: the rule text still said the floor was "not yet wired" in the change that wired it. #1507 corrects it.

## Verification

`harness:test` 1481/1481; `harness:scan` 81 of 82, the one failure being the known local-only `progress-report-quantification` which skips in CI (`HARNESS-054`).


woojubb:
낡은 develop 기반이라 promotion ancestry / release-grade verification 이 실패합니다. 현재 develop 에서 새 승격 브랜치를 만들어 대체합니다.

PR #1535, opened 2026-07-31
fix(harness): the red-proof gate can see the layer it is meant to guard

woojubb:
## What

`check-regression-red-proof` (HARNESS-041) catches a regression test that passes on the unfixed code. It decided what to examine with one expression matching `packages/*/src` and `apps/*/src`, so **`.claude/hooks/**` — every guard in the repository — and `scripts/harness/**` — every scan and every floor — were invisible to it.**

Measured over PRs #1525–#1530: twelve CI runs, **zero verdicts**. In that same window human review caught four accidental-green tests, all of them in that blind spot. The mechanical floor for exactly this defect was blind to every one.

## Why widening `pkgOf` alone would have changed nothing

- A `.sh` hook and the harness test that runs it live in different directories, so they never pair. The hook subject now **adopts** the changed harness tests as candidates.
- A shell script is never in a module graph, so the C3 import-graph check would answer "not imported" for every hook and return INCONCLUSIVE — a SKIP by another name. For that subject the relation becomes **"this test SPAWNS this hook"** (`testExecutesHook`, comments stripped: a hook named in prose is described, not run).

## The defect underneath

Reaching the mutation step is what exposed it. `reverseApply` read the diff through a helper that `.trim()`s, so the patch arrived at `git apply -R` **without its final newline** and git rejected it as corrupt — "patch broke at line 60". **Every reverse-apply had thrown since the gate was written.** The verdict count was zero because the mutation never happened, not because nothing qualified — and it was broken for `packages/*/src` just as completely.

## The acceptance bullet this retracts

INFRA-071 asked for one of the four accidental-greens to be replayed and come back `ACCIDENTAL_GREEN`. Both candidate ranges were replayed in a detached worktree:

| Range | Verdict | Case that fails when the fix is reversed |
| --- | --- | --- |
| `2ac10f251..b1f46acf3` | `red-proof-ok` | `post-tool-format > … outside the formatted set` |
| `b1f46acf3..c08e0dbd6` | `red-proof-ok` | `post-tool-format > … project directory is unset` |

**Those verdicts are correct.** Each range's changed test file does contain a case that genuinely fails on the reversed hook. The bullet was written on a false premise: reverse-apply answers *"does this test depend on the fix?"*, and all four misses failed a different question — *"does this test REACH the behavior it names?"* Three have no source to reverse at all; the fourth fails reversed for the wrong reason (the hook crashes before the branch under test), which pass/fail output cannot distinguish. Recorded with the measurement rather than quietly restated, and re-filed as **INFRA-072** with three candidate mechanisms and the reason none is chosen yet.

## Verification

- **Red-proved**: reverting the source fails all seven new cases; reintroducing the trim alone fails exactly the byte-exactness case; removing the doc exclusion fails exactly the classification case.
- 1666 harness tests green (112 files), 82 scans pass.
- Round A review before push: 2 findings (docs-as-source in the widened `pkgOf`; the spawn rule duplicated between the floor and the gate), both fixed in `d0e89820f`.

This PR is itself a `fix:` range with a `scripts/harness` source+test pair, so the widened gate should produce a live verdict on it rather than a SKIP — the first one it has ever emitted.

### DIFF
--- a/.agents/backlog/INFRA-071-red-proof-gate-cannot-see-the-harness.md
+++ b/.agents/backlog/INFRA-071-red-proof-gate-cannot-see-the-harness.md
@@ -86,14 +86,51 @@ re-running the executing test is the same operation vitest already performs for 
 
 This is correctness-critical code in a gate, so it is worth doing deliberately rather than quickly.
 
+## What the implementation found (2026-07-31)
+
+**The gate had never once reverse-applied anything.** `reverseApply` read the diff through a helper
+that `.trim()`s its output, so the patch reached `git apply -R` without its final newline and git
+rejected it as corrupt — "patch broke at line 60". Twelve CI runs, zero verdicts: the widening is
+what made the line reachable, and reaching it is what exposed this. It was never a `packages/*/src`
+problem versus a hooks problem — the mutation step was broken for every subject alike.
+
+**The third acceptance bullet below was written on a false premise, and the replay disproves it.**
+Both candidate ranges were replayed through the widened gate in a detached worktree:
+
+| Range                  | Verdict        | Case that fails when the fix is reversed          |
+| ---------------------- | -------------- | ------------------------------------------------- |
+| `2ac10f251..b1f46acf3` | `red-proof-ok` | `post-tool-format > … outside the formatted set`  |
+| `b1f46acf3..c08e0dbd6` | `red-proof-ok` | `post-tool-format > … project directory is unset` |
+
+Those verdicts are CORRECT. Each range's changed test file does contain a case that fails on the
+reversed hook, so the range is red-proved whatever else is true of its other cases.
+
+The four misses are a **different defect class**, and reverse-apply cannot reach them by
+construction:
+
+- `hooks-have-execution-coverage` — the logic under test lives inside the test file. There is no
+  source to reverse.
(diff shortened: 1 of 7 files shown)


woojubb:
**On the last review's MUST — verified, and it does not hold.**

Both `HOOKS_DIR` sites are inside template literals with **no `${}` interpolation**, so the identifier is never evaluated — it is plain text in the fake test source being handed to `testExecutesHook`, which reads it as a string:

```
388:      readText: () => `spawnSync('bash', [path.join(HOOKS_DIR, 'some-hook.sh')]);`
421:      [abs(spawner)]: `spawnSync('bash', [path.join(HOOKS_DIR, 'some-hook.sh')]);`
```

Measured rather than argued — the three hook-pair fixtures run green:

```
$ npx vitest run …/check-regression-red-proof.test.mjs -t "hook pair"
Tests  3 passed | 34 skipped (37)
```

and the full harness suite is 1667/1667. So the "contradicts the PR's tests-green claim" conclusion does not follow.

**Changed anyway, for a different reason.** A fixture that reads as evaluable code will be misread by the next reader too. The paths are literal strings now (`'/repo/.claude/hooks/some-hook.sh'`), which is also closer to what the relation actually inspects.

Issue #1540, opened 2026-07-31
INFRA-074: the spawn relation names a hook and spawns a shell, without tying one to the other


woojubb:
Filed from a **FOUNDATIONAL** review finding on #1535 — held, not patched.

## Problem

`testExecutesHook` answers "does this test execute this hook?" with two independent checks: the hook's basename appears in the comment-stripped text, AND the file spawns `bash` somewhere. **Nothing ties the spawn to the name.** A test that names hook A in real code while spawning `bash` for hook B counts as executing A.

## Why it matters more than it did

The relation was written for `hooks-have-execution-coverage`, an advisory floor whose worst outcome was a message about a hook nobody runs. Its docstring called it "structural rather than exact … a grep-level floor by design", which was a fair trade for what it decided.

INFRA-071 gave it a second job: for a `.claude/hooks` subject it now picks `decidingTests` — the tests whose pass/fail sets `RED_PROOF_OK` or `ACCIDENTAL_GREEN`. A misclassified bystander can supply a verdict about a hook it never ran. **The imprecision did not change; the consequence did.**

## Why a narrower match is not the fix

Already tried and rejected on evidence: requiring the name inside a `path.join(...)` missed every test that passes the basename to a helper which joins it (`run('some-hook.sh', …)`) — those run the hook just as truly. Any text-only pattern strict enough to tie name to spawn will keep rediscovering that, because the binding is a value flowing through a call, not a lexical adjacency. Doing it properly means reading the call graph.

## Containment in force

The relation's docstring names this item and states the changed consequence. The gate is ADVISORY (`REGRESSION_RED_PROOF_ENFORCE` unset), so a misclassification approves nothing today. **Must be resolved before the gate is promoted to enforcing (INFRA-046)** — same gate as #1536, same reason.

Detail document: `.agents/backlog/INFRA-074-the-spawn-relation-is-not-tied-to-its-argument.md`

woojubb:
Resolved by #1568, merged. Verified on `origin/develop`: `scripts/harness/lib/spawn-call-graph.mjs` exists and `check-regression-red-proof.mjs` carries the three-answer verdict.

## What replaced the two independent checks

The relation no longer asks "does the basename appear?" and "does the file spawn `bash`?" separately.
It parses the module, finds spawn calls **by their `child_process` import binding** — so a
`spawnSync('bash', …)` inside a string literal is not one — and resolves the expression in **argv
script position** back through the module's own bindings: literals, template and `+` concatenation,
[...]

PR #1541, opened 2026-07-31
feat(rules): a helper's limits are re-judged where the consequences change

woojubb:
## The class

A shared function's documented limits were judged against what its **first** consumer did with the answer. That judgement is a property of the pair, not of the function.

When a second consumer arrives whose consequences are heavier, **the function does not change** — so nothing in the diff signals anything, and review sees a reuse, which reads as good practice. The defect is invisible in the code and shows up only in behaviour, usually much later.

## Two instances, measured in one session

| Helper | Judged against | Reused for | What it cost |
| --- | --- | --- | --- |
| `git()` — trims its output | a sha, where trimming is right | producing the input to `git apply -R` | the final newline stripped, every patch "corrupt", the red-proof gate's mutation step threw for its **entire life** — twelve CI runs, zero verdicts, no error anyone saw |
| `testExecutesHook` — grep-level, "structural rather than exact" | an **advisory** coverage message | picking which tests may set a red-proof **verdict** | the same imprecision can hand a verdict to a test that never ran the hook (INFRA-074, #1540) |

## What landed

**RULE** `.agents/rules/helper-limits.md` — before reusing a helper where the consequences differ, re-judge its stated limits at the new call site and record the answer. Three outcomes, only three: they hold (say why), they do not (narrow it or use something else), or the mismatch is **held** under a labelled containment naming a root item (`finding-depth.md`, #1539). The helper's author could not have answered this — the second consumer did not exist yet.

**FLOOR** `scripts/harness/scan-helper-limits.mjs`, registered in `harness:scan` (82 → 83 scans). A helper declares its limits with `@limits <statement>` on the docblock attached to its export; every module importing it must carry `// LIMITS <name>: <why they hold here>`.

- **Declaring is opt-in**, deliberately — a blanket requirement would be met by boilerplate and would say nothing.
- **Acknowledging is not optional** once declared.
- **Anti-rot** follows `allow-fake` / `allow-fallback`: an empty tag, or a `LIMITS <name>:` with no reason, both FAIL.
- The scan **reports what it examined** (`N @limits-tagged function(s) across M file(s)`), so a run that examined nothing cannot read as a clean sweep.

## Proved against the pre-fix state

Seeded on the instance the lesson came from. Removing the acknowledgement from the coverage floor makes the scan **fail** with the exact message; restoring it makes it **pass**. That is the state the repo was actually in when the defect was introduced.

Two defects were found **by running it**, both fixed and pinned with cases:

- a fixture inside a string literal was read as a real declaration — the scan of a scan is full of them, and the repo's other scans state that string literals must not match;
- the module docblock's own contract prose (which necessarily writes `@limits`) was adopted as a tag on an unrelated export. A tag that drifts names limits belonging to something else, which is worse than no tag.

## Verification

- 9 tests in `scan-helper-limits.test.mjs` green; 1676 harness tests green.
- 83 scans pass. Two scans (`dist`, `doc-examples`) need a built tree and were not run in this worktree; they run with a build in CI.

Lesson processed per `lesson-to-harness` steps 3–9 (normalize → tier A → mechanism → prove it catches the incident), with the candidate approved before institutionalizing.

### DIFF
--- a/.agents/rules/helper-limits.md
+++ b/.agents/rules/helper-limits.md
@@ -1,0 +1,54 @@
+# Helper Limits — re-judged at every consumer whose consequences differ
+
+Mandatory. Parent: [index.md](index.md) § Process Sub-Rules.
+
+A shared function's documented limits were judged against what its FIRST consumer did with the
+answer. That judgement is not a property of the function. It is a property of the pair.
+
+When a second consumer arrives whose consequences are heavier, **the function does not change** — so
+nothing in the diff signals anything, and review sees a reuse, which reads as good practice. The
+defect is invisible in the code and shows up only in behaviour, usually much later.
+
+Two instances, measured in one session:
+
+- `git()` trimmed its output. Right for a sha; wrong for a patch. Reused to produce the input to
+  `git apply -R`, it stripped the final newline, git called every patch corrupt, and the red-proof
+  gate's mutation step threw for its entire life — twelve CI runs, zero verdicts, no error anyone saw.
+- `testExecutesHook` was a grep-level relation for an ADVISORY coverage floor whose own docstring
+  called it "structural rather than exact". Reused to pick which tests may set a red-proof VERDICT,
+  the same imprecision can hand a verdict to a test that never ran the hook (INFRA-074).
+
+## The rule
+
+**Before reusing a helper in a context with different consequences, re-judge its stated limits at
+the new call site, and record the answer there.**
+
+Three outcomes, and only three:
+
+1. **They hold** — say why, at the call site. One line.
+2. **They do not hold** — narrow the helper, or use a different one.
+3. **They do not hold and the mismatch is being held** — a labelled containment naming a root item,
+   per [finding-depth.md](finding-depth.md). Never an unlabelled hold.
+
+The point is not the comment. It is that the question is ASKED where the consequences are known,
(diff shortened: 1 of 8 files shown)


woojubb:
**On the adjacency finding — verified, and the premise does not hold.**

The claim is that a `@limits` tag separated from its export by a blank line "will silently fail to be recognized — the scan will report it as untagged/unexamined with no error". Measured on this branch:

```
$ node -e "…taggedFunctions(spacedDocblock)…"
parsed:   []
findings: ["the text carries @limits and the parser read none of it — the reader has drifted,
            and a reader that reads nothing reports a clean sweep"]
```

The per-file reader check added in the previous round fires on exactly this: the tag IS present in tag position, and nothing parsed from that file. It is loud, not silent.

**Adjacency stays required, deliberately.** Tolerating a gap would let a MODULE docblock tag whatever export happened to follow it — a drifted tag naming limits that belong to something else, which is a defect this scan already had once in this PR and now has a case against.

Changed anyway, for a different reason: the property was holding by consequence rather than by contract. It is now pinned by a case, so a future change to the reader cannot quietly turn a loud drift into a silent one.

**On the namespace-import gap (CONSIDER):** accepted and stated in the module docblock, beside the file's other declared limits — `import * as owner` is not detected because the binding that matters is a property access rather than an imported name.

PR #1557, opened 2026-07-31
feat(rules): re-plan withdraws the change, and the gate holds it to that

woojubb:
Owner proposal: on a FOUNDATIONAL verdict — file the issue, comment on the PR, **close it**. Adopted for the disposition it belongs to, with the reason for the distinction measured rather than argued.

## The hole the proposal found

`finding-depth.md` already allowed two dispositions, and **only one of them did anything**.

| disposition | what it did before |
| --- | --- |
| `containment` | left a code comment and a commit body carrying the root item's ID |
| `re-plan` — *the change is withdrawn or reduced* | **a word in a note that nothing read** |

So a change recorded as withdrawn merged like any other. A disposition nothing acts on is a decision with no actor — the shape this rule exists against, sitting inside the rule.

## What landed

- The recorder **refuses** a foundational finding with no disposition. The rule allows exactly two and never a third; recording the root item without saying which was taken leaves the decision in prose.
- The disposition is **stored**, and `merge-gate` reads it. A re-planned change **does not merge**.
- Comment-and-close is the tidy expression of a withdrawal; refusing the merge is the part that cannot be forgotten.

## Not adopted: closing on every foundational finding

Measured against the three foundational verdicts taken on 2026-08-01, across two PRs — **in all three the root predated the change, and the change was independently correct**:

| verdict | closing that PR would have |
| --- | --- |
| INFRA-073 (#1535) | discarded the fix that gave a dead gate its **first verdict in its entire life** — `reverseApply` had thrown since the gate was written |
| INFRA-074 (#1535) | same PR |
| INFRA-075 (#1547) | left **two live guard bypasses** standing while a tokenizer was written — and that PR's corpus is what made the root fixable at all |

A foundational finding says where the **defect** is. Whether this change is worth keeping is a different question, and only the second decides the PR.

## Verification

- **Test-first**, five cases: no disposition refused, a third option refused, the value stored, `re-plan` reported as unmergeable, `containment` explicitly **not** refused.
- Gate cases build a real checkout with a recorded round and run the hook against it — `re-plan` → exit 2 naming the disposition, `containment` → exit 0.
- The gate resolves the checkout from the payload's `cwd` before reading the record, for the same reason the recorder does: pointing it at the wrong one would judge this merge against another checkout's round.
- 1800 harness tests green (119 files), 83 scans pass.

### DIFF
--- a/.agents/rules/finding-depth.md
+++ b/.agents/rules/finding-depth.md
@@ -35,6 +35,22 @@ Containment is permitted only when the change must land first, and only under al
 it is the smallest thing that keeps the tree honest, it introduces no new abstraction, and it names
 the root item's ID in both a code comment and the commit body. An unlabelled hold is a patch.
 
+**The disposition decides what happens to the change, and it is recorded, not narrated.**
+
+- **re-plan** — the change is WITHDRAWN or reduced. Comment the decision on the PR and CLOSE it. It
+  does not merge as it stands, and `merge-gate` refuses it: `record-local-review --disposition
+re-plan` is read at the merge, so the withdrawal is a state the machine holds rather than a
+  sentence someone wrote. Until that existed, `re-plan` was the only part of this rule with no
+  consequence — containment left a code comment and a commit body, and re-plan left a note.
+- **containment** — the change lands with the labelled hold above, and the PR stays open.
+
+**Closing on re-plan is not the same as closing on every foundational finding**, and the difference
+is measured. On 2026-08-01, three foundational verdicts were taken on two PRs; in all three the root
+predated the change and the change was independently correct. Closing them would have discarded a
+fix that gave a dead gate its first verdict in its entire life, and left two live guard bypasses
+standing while a tokenizer was written. A foundational finding says where the DEFECT is, not whether
+this change is worth keeping — those are different questions, and only the second decides the PR.
+
 **The label is a condition, not a courtesy.** A hold with no such comment is indistinguishable from
 having ignored the finding. It is also what lets the review loop converge: a foundational finding is
 not fixed, so the next round sees the same code and would raise it again. The comment at the site IS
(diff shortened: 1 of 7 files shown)


woojubb:
## Withdrawn — FOUNDATIONAL, root filed as #1562

The reviewer's first SHOULD went to `finding-depth-triager`, and the verdict came back
**FOUNDATIONAL**. Per `.agents/rules/finding-depth.md`, a re-plan disposition withdraws the change
rather than patching it, so this PR is closed and the cause is filed as **#1562**.

### What the verdict found that the review did not

The review reported that the disposition check keys on the **local checkout's branch** rather than
on `$PR`. It does — and the consequence is worse than "does not block".

`.agents/local-reviews/` is gitignored and per-working-tree. Measured on this machine while judging:

- `robota-wt-063` holds the only record for `feat/harness-063-examined-counts`; the main clone holds
  none.
- The main clone is on `fix/infra-076-anchor-every-override` and **has** a record for it.

An orchestrator merging #1561 from the main clone therefore does not find nothing — it reads
**another PR's record** and answers one merge with a different PR's disposition.

And it is guaranteed to be a different checkout. `worktree-parallel-orchestration` §5:
*"The orchestrator — never the implementer — merges."*

### Why I am not applying the suggested fix

Resolving `headRefName` via `gh pr view` fixes the branch-name half only. `mergeBlockReason` returns
`null` — do not block — when the file is **absent**, and it is absent in every checkout but the
recording one. Resolving the sha from `gh` as well still leaves the orchestrator's merge passing,
because there is nothing to read. The fix would narrow the hole from "any other directory" to "any
other clone or worktree" — which is exactly where merges happen here.

### The repeat, stated plainly

INFRA-048 and INFRA-057 already established, and already paid for, that a merge decision held
anywhere other than the PR does not stop a merge. Their answer — `.github/workflows/review-gate.yml`,
a required check keyed by PR number with a label override — exists and is required on `develop` and
`main`. This PR reintroduced the removed shape for a *stronger* decision, in the one place the shape
was proven not to hold.

The workaround appears twice inside this PR: the hook resolving the checkout from the payload `cwd`,
and `e8220259e` isolating the tests from the developer's own record. Two patches, one fact — the
evidence is machine state, not PR state.

### What survives

The local review store is not the problem and is not being removed; it has been correct for
`pre-push-check.sh` since `fae9df113`, where the subject genuinely **is** the local checkout. This
was its first PR-scoped consumer.

The replacement will publish the disposition to the PR as part of recording it, and the gate will
read it by `$PR`. The rule text's claim that *"the withdrawal is a state the machine holds rather
than a sentence someone wrote"* goes with it — today the machine holds it in one untracked directory
on one host, and #1562 says so.

Issue #1548, opened 2026-07-31
INFRA-076: branch-guard honours an override token wherever it appears, not where it is given


woojubb:
## Problem

`branch-guard.sh` reads its four overrides as a token appearing **anywhere** in the masked command. Masking stops a *quoted* mention; it does not stop an unquoted one, because an unquoted token is a real word in the command.

Measured on the sibling hook `worktree-cwd-guard`, which had the identical rule:

| command | verdict |
| --- | --- |
| `git commit -m "TOKEN" && git reset --hard` | blocked (masked) |
| `git commit -m TOKEN && git reset --hard` | **allowed** |
| `echo TOKEN ; git reset --hard` | **allowed** |

That hook was repaired in #1547 by anchoring the override to the command it overrides — an override is something you GIVE a command, so it must prefix one, which is what the hook's own refusal message always said.

## Why filed rather than fixed alongside

`branch-guard`'s overrides are **documented as loose**, and its comments explain the choice. Narrowing them is a user-visible behaviour change to the most-used hook in the repo: an operator whose habit puts the token elsewhere in a compound command would start being refused. That is a decision, not a repair.

The repo currently runs **both conventions**: `pre-push-check`'s `PRE_PUSH_ALLOW_UNREVIEWED` and `merge-gate`'s ACK already require the strict prefix. Settling on one is the actual work.

## Options

1. **Anchor all four** — uniform, closes the hole, breaks any established loose usage.
2. **Anchor the destructive ones only** (`_ALLOW_DELETE`, `_ALLOW_MAIN_MERGE`) — closes the cases where a mention costs something irreversible.
3. **Leave and document the exposure** — only defensible if the loose form is load-bearing for a real workflow; name it if so.

Detail document: `.agents/backlog/INFRA-076-loose-override-tokens.md`

woojubb:
Resolved by #1559, merged — with one half deliberately held, filed as #1563.

## What is closed

An override token is no longer honoured wherever it appears. It must prefix a statement, that
statement must carry the action it excuses, and **every** statement carrying that action must carry
it. Measured on a scratch repo, each against a bare control that is correctly refused:

```
BRANCH_GUARD_ALLOW_MAIN_MERGE=1 git status; git push origin main                     0 -> 2
BRANCH_GUARD_ALLOW_MAIN_MERGE=1 git merge-base develop main ; git merge develop      0 -> 2
[...]

PR #1559, opened 2026-07-31
fix(hooks): one positional reading for every override, and one that never worked

woojubb:
**INFRA-076** (#1548) — decided on measurement, and a second defect found while testing it.

## The decision, and why its stated cost is zero

The item offered three options; option 1 (anchor all four) carried the objection *"breaks any established loose usage"*.

**Measured before changing anything: there is none.** `git-branch.md` says the override goes "**inline in the same command**" and shows `BRANCH_GUARD_ALLOW_DELETE=1 git push origin --delete <name>`. Every documented usage is the prefix form; every use in practice is that shape.

So anchoring costs nothing and ends two conventions living side by side — three hooks already anchored, one did not. All five now share **one positional reading**: the token prefixes the command it overrides, optionally behind other assignments, and nowhere else.

## The second defect, opposite in polarity

`BRANCH_GUARD_ALLOW_MAIN_MERGE` was read **only** from the hook's own environment:

```sh
if [[ "$IS_MERGE" == "true" && "${BRANCH_GUARD_ALLOW_MAIN_MERGE:-0}" != "1" ]]; then
```

But `VAR=1 cmd` sets the variable in the **tool's** shell, which never reaches this process. So the form the rule documents **could not work at all** — for the two most dangerous operations the guard covers: pushing to `main` and merging into `main`. An operator following the printed instruction was refused.

INFRA-076 was about a mention that disarms too easily; this was an override that could not be given at all. Same subject, and the same fix covers both.

## The fixtures had to prove themselves

Each case asserts **three ways in one fixture**:

1. the guard **fires** here at all — without this the case proves nothing;
2. a bare mention does **not** disarm it;
3. the documented prefix **does**.

The first version asserted only (2). It passed on **three of four** overrides — including where the guard was blocking for an unrelated reason and the case never reached the check. That is the accidental-green shape, caught by making each case validate its own premise rather than by reading it.

## Verification

- 1798 harness tests green (119 files), 83 scans pass.
- The delete case runs on a feature branch on purpose: on `main` the protected-branch push rule fires first and the case would say nothing about the delete override.

### DIFF
--- a/.claude/hooks/branch-guard.sh
+++ b/.claude/hooks/branch-guard.sh
@@ -80,18 +80,43 @@ override_given() {
   return 1
 }
 
+# A quote and a backtick are boundaries too: a KEPT region — `bash -c "git push"`, or a backtick
+# subshell — puts one immediately before the verb, and without them the region survived masking and
+# still matched nothing. Quoted payloads are masked before this runs, so this cannot resurrect the
+# false positive it sits beside.
+GITPFX='(^|[;&|({"'"'"'`]|[[:space:]])[[:space:]]*(\S+=\S+\s+)*git\s+((-C|-c)\s+\S+\s+)*'
+# Trailing boundary: anything that is not a word character or `-`. `\b` alone let `git merge-base`
+# read as a merge and `git commit-tree` as a commit — false positives that, now that the leading
+# match is loose, would block ordinary read-only work on a protected branch. It also covers the verb
+# ending a line (`git commit\ngit push`), which a bare `(\s|$)` misses.
+GITEND='([^-[:alnum:]_]|$)'
+# Tolerate flags between the subcommand and the create flag (e.g. `git checkout -q -b x`, which
+# previously slipped past the create-guard entirely). `-B`/`-C` are the force-create spellings and
+# create a branch just as much as `-b`/`-c` do.
+#
+# ONE expression per action, read by BOTH the detector below and the override check above it.
+# Review of #1559 found the override carrying its own hand-written copies, forked in both
+# directions at once: no trailing boundary, so `git merge-base` excused a real `git merge`; and no
+# room for an intervening flag, so the documented `git checkout -q -b x` stopped registering. A
+# second spelling of "what counts as this action" is a second answer waiting to disagree, and here
+# it disagreed the moment it was written.
+RE_COMMIT="${GITPFX}commit${GITEND}"
+RE_PUSH="${GITPFX}push${GITEND}"
+RE_MERGE="${GITPFX}merge${GITEND}"
+RE_CREATE="${GITPFX}(checkout\s+(-\S+\s+)*-[bB]|switch\s+(-\S+\s+)*-[cC])${GITEND}"
+RE_GH_API="(^|[;&|({\"'\`]|[[:space:]])[[:space:]]*gh[[:space:]]+api${GITEND}"
+
 # Each override is bound to the command it excuses, so a decoy statement cannot stand in for it.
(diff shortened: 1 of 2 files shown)


woojubb:
## Both MUSTs fixed; the finding's depth verdict is FOUNDATIONAL, contained under #1563

Both halves of the MUST reproduced exactly as written, on a scratch repo, before any change:

```
exit=0  BRANCH_GUARD_ALLOW_MAIN_MERGE=1 git merge-base develop main ; git merge develop
exit=0  BRANCH_GUARD_ALLOW_BADNAME=1    git checkout -bogus ; git checkout -b BAD_NAME
exit=2  BRANCH_GUARD_ALLOW_BADNAME=1    git checkout -q -b BAD_NAME     (the SHOULD)
```

The cause is the one the review named: `override_given` was given **hand-written copies** of "what
counts as a merge / a push / a creation" instead of the expressions this file already uses to detect
each action. They forked in both directions at once — too loose at the end (no `GITEND`, so
`merge-base` read as `merge`) and too strict in the middle (no room for `-q`, no `-B`/`-C`). `GITPFX`
/ `GITEND` and four `RE_*` now sit above the override block and are read by both, so there is no
second answer left to disagree.

### What the depth pass found beyond the review

I sent the finding to `finding-depth-triager` before fixing, because a second round of the same
finding on the same lines is the signal the rule exists for. Verdict: **FOUNDATIONAL**, with four
bypasses measured against this branch's head, each paired with a bare control that is correctly
refused:

| # | command | before |
| --- | --- | --- |
| A | `ALLOW_DELETE=1 git push origin --delete scratch-1 ; git push origin --delete develop` | 0 |
| B | `git checkout -b feat/x develop ; git checkout -b feat/y main` | 0 |
| C | `git checkout -b feat/ok ; git checkout -b BAD_NAME` | 0 |
| D | `ALLOW_MAIN_MERGE=1 git merge develop ; git push origin main` | 0 |

**B and C carry no override token at all.** None of the four is closed by `GITEND`.

### The part I got wrong, stated plainly

**D was newly reachable on this branch.** On `develop` it exits 2 — not because `develop` is
careful, but because the inline form of `ALLOW_MAIN_MERGE` was dead there. Making a dead override
live on top of whole-command action detection is what opened it. My first instinct was to contain
the lot and land it; that would have been shipping a new hole on the push-to-`main` path with a note
attached.

### What landed instead

The override was statement-scoped; the *action detection* never was. The right question is not "does
a statement carry the override" but "does **every** statement carrying this action carry it" — the
invariant `worktree-cwd-guard.sh` already states as `OVERRIDDEN_STATEMENTS == DESTRUCTIVE_STATEMENTS`,
and the third thing this file has now taken from that sibling one piece at a time.

```
ALLOW_MAIN_MERGE=1 git merge develop ; git push origin main              0 -> 2
ALLOW_DELETE=1 git push --delete x ; git push --delete develop           0 -> 2
ALLOW_MAIN_MERGE=1 git merge develop ; ALLOW_MAIN_MERGE=1 git push origin main   still 0
```

The last line is asserted deliberately: requiring "no sibling exists" would refuse correctly-overridden
work, which is how a guard earns being switched off.

### Contained, not fixed — #1563

B and C need per-statement **extraction**: `hook_match_extract` uses awk `match()` and returns the
first match of the whole string, so a command with two creations has its second judged by nothing.
That needs an aligned split seam in `lib/command-scan.sh` — a library change, not this function's
subject. Both reproduce **unchanged on `develop`**, verified by running `origin/develop`'s copy of
the hook against the same scratch repos, so this PR neither opens nor widens them.

The hold is written at the call site with both reproductions and the instruction not to add a third
override reading, since the remaining hole is not in the override.

1804 harness tests green, 83 scans pass.

PR #1565, opened 2026-07-31
fix(hooks): read the shell grammar with a stack, because quoting is not regular

### LINKED ISSUE #1542: INFRA-075: restoring a command-substitution span un-masks the quotes inside it
Filed from a **FOUNDATIONAL** finding — a live false positive, measured 2026-08-01 on `develop`.

## Reproduction

```
command:  out=$(printf 'x git commit -m y' | bash h.sh); echo done
verdict:  [branch-guard] Blocked: cannot git commit on protected branch 'develop'.
exit:     2
```

Nothing there commits anything. The verb sits inside a single-quoted argument to `printf` — exactly what the masker exists to neutralise, and it *does* neutralise it when not nested:

| command | verdict |
| --- | --- |
| `printf 'x git commit -m y' \| bash h.sh` | exit 0 — correctly ignored |
| `out=$(printf 'x git commit -m y' \| …)` | **exit 2 — falsely refused** |

## Cause

`lib/command-scan.sh` masks in two passes: the first masks quoted regions, the second restores command-substitution spans (a substitution runs whatever the quoting around it). The restore copies **every** character of the span back from the original — including the quoted text the first pass had just masked:

```awk
for (k = i; k <= stop; k++) { m[k] = substr(s, k, 1) }   # restores the WHOLE span
```

## Why it is not repaired in place

The pass is the wrong shape, not a line with a bug in it. A substitution's content is **itself a command** and must be masked by the same rules — that is recursion, which this pass cannot express. Restoring a narrower span is wrong the other way: `"$(git commit)"` inside double quotes really does run.

This is the repository's highest-churn guard — **26 commits in 5 days** — and a four-day audit found **seven** instances of one class: command text interpreted by an approximation of the shell rather than by its rules. Every one was found by a person hitting a new spelling; two refused the creation of the branch their own fix lived on. Another patch to the same pass buys the next spelling, not the class.

## Direction

A tokenizer modelling the shell's grammar for what the guards care about — quoting, escapes, comments, heredocs, redirections, continuations, substitutions **nested to any depth** — plus a differential corpus test: every hook's decision must match the decision on the same command canonicalised by that tokenizer. No oracle needed beyond the tokenizer.

## Containment

None is possible at the call site — the defect is a false REFUSAL, so there is no path to label. The reproduction is recorded instead, so the next person blocked by a quoted mention finds it in one search. `BRANCH_GUARD_ALLOW_*` overrides remain the way past, and they announce themselves.

Detail document: `.agents/backlog/INFRA-075-substitution-span-restores-its-own-quotes.md`

woojubb:
Closes #1542

## What was wrong

`command-scan.sh` decided what a command runs with two linear passes: mask quoted regions, then restore command-substitution spans. Shell quoting is not a regular language, so no arrangement of linear passes can say what a quote inside a substitution inside a quote means. Both spellings of that were live at once when this started:

| command | bash | the guards |
| --- | --- | --- |
| `out=$(printf 'x git commit -m y' \| bash h.sh)` | commits nothing | **refused** (the issue's reproduction, closed by #1556) |
| `a=$(echo "$(git commit)")` | **commits** | **saw nothing** |

The second is the worse one and it arrived as the fix for the first: #1556 taught the restore pass to keep a quote that opened inside the span, which held the inner quotes masked — and with them the nested substitution that genuinely runs. A bypass, not a nuisance.

## What this does

Replaces both passes with a tokenizer over an explicit context stack, so a substitution's content is read by the same rules as the command around it, at any depth. It models:

single quotes · double quotes · ANSI-C `$'…'` · locale `$"…"` · backslash escapes · the backslash-newline continuation · `#` comments · heredocs (`<<`, `<<-`, quoted and unquoted delimiters) · herestrings · redirections · subshells · `${…}` · `$((…))` · `$(…)` and backtick substitutions **nested to any depth** · the string an interpreter is told to run.

**Two interpreter lists, not one.** Only shell-family interpreters and `eval` have their string descended into as shell. A `python3 -c` argument is a command but not a *shell* command, so applying shell quoting to it would be a second approximation dressed as a parse; those stay read whole.

Every caller is unchanged — the change sits behind the existing `command-scan.sh` interface.

## The acceptance criterion, set before the tokenizer was written

A differential corpus of **202 shapes** in `scripts/harness/__tests__/hook-reading-matches-bash.test.mjs` — 101 written by hand, 96 generated by wrapping base commands in up to three layers of substitution and quoting. Each is run for real through bash with a recording `git` stub on `PATH`, and the reading must agree with what the shell actually invoked. The threshold is stated in the file: every disagreement must be a listed known-open shape, and the rate must clear 98%.

| | agreement | of which bypasses |
| --- | --- | --- |
| before (pre-tokenizer) | **177 / 197** (89.8%) | 5 |
| after | **199 / 202** (98.5%) | 0 |

The committed test is red against the pre-fix library: 18 of its 200 cases fail.

## What the tokenizer still cannot read

**Command position** — deliberately, and this is the whole of the remaining 3:

```
echo git push
eval 'echo git push'
echo \"git push\"
```

The verb is an ARGUMENT the shell never runs as a command, and the reading still shows it. Hiding it means masking every word that is not the first of a simple command, and then `sudo git push`, `xargs git push`, `env git push`, `timeout 5 git push`, `command git push` and every other exec-style wrapper not on some list goes unseen. **An unmodelled wrapper is a silent bypass; an argument read as a command is a refusal that announces itself.** The trade is taken in the direction that fails loudly. All three are pinned in `KNOWN_OPEN` with that reason, so closing the class turns those cases red and retires the exemption.

Smaller limits, stated rather than found later:

- multiple heredoc openers on one line are queued in order, but a heredoc opened *inside* a substitution attaches its body to whatever line ends next;
- `${…}` and `$((…))` are read as data that may contain substitutions, not as their own grammars;
- `tcsh`/`csh` are on the shell list although their quoting differs from POSIX shells.

## False positives on the real tree

A guard that refuses correct work gets switched off, so this was measured rather than asserted. **2453 command lines** harvested from every `*.md`, `*.sh`, `*.mjs` and workflow file in the repository were read under both versions. **3 verdicts changed, all newly-SEEN, none newly blind** — and all three are a genuine `$(git …)` the shell would run:

```
`cat <<EOF${NL}$(git push)${NL}EOF`,
`cat <<EOF${NL}$(echo "$(git push)")${NL}EOF`,
the unquoted characters) is wrong in the other direction: `"$(git commit)"` inside double quotes
```

## Does this fix the guards' VERDICTS, or only the reading?

Asked on review, and answered by running `branch-guard.sh` itself on a scratch repo with every `BRANCH_GUARD_ALLOW_*` override absent and `gh` stubbed to fail loudly. It matters because `branch-guard` takes `COMMAND_VERBS` from `hook_verb_scan` but `COMMAND_EXEC` from `hook_executable_part`, so a fixed reading does not imply a fixed decision.

| command | develop | this branch |
| --- | --- | --- |
| `git push origin main` *(control)* | **BLOCKED** | **BLOCKED** |
| `git \`⏎`  push origin main` | allowed | **BLOCKED** |
| `a=$(echo "$(git push origin main)")` | allowed | **BLOCKED** |
| `cat <<EOF`⏎`$(echo "$(git push origin main)")`⏎`EOF` | allowed | **BLOCKED** |
| `a=$(echo "$(git push origin --delete develop)")` | allowed | **BLOCKED** |
| `echo 'git push origin main'` *(mention)* | allowed | allowed |
| `cat <<'EOF'`⏎`git push origin main`⏎`EOF` | allowed | allowed |

**Closed for `branch-guard`'s push and delete decisions.** The measurement also surfaced a fourth defect the reading alone hid: the nested delete named the branch **`develop)`**, paren included, so it missed the protected-branch check and was refused by the weaker merged-PR check instead. Fixed in 8aef2bfc0 (an unquoted value now ends at `)` and a backtick), red-proved first.

**Still open, stated plainly:** `pre-push-check.sh` and `worktree-cwd-guard.sh` also grep `COMMAND_EXEC`, the pre-tokenizer reading, and this PR does **not** measure their verdicts. #1572 tracks that, with the measurement behind it. A tokenizer that fixes the function but not the decision is the "registered, never reached" shape, so it is named here rather than left to be found later.

`scripts/harness/__tests__/branch-guard-reads-nested-commands.test.mjs` is the mechanism that keeps the table above honest — it asserts the exit code and the branch the guard NAMES, not the reading behind it.

## Verification

- Rebased onto `develop` @2e325094f (#1559, #1564, #1567). `pnpm harness:pre-push` runs **green** — 2032/2032, 122 files. The three `worktree-cwd-guard` failures an earlier revision worked around are gone, which is #1567's claim, tested rather than assumed. Nothing here was pushed with `--no-verify`.
- `pnpm harness:scan` — **84 of 84** after `pnpm build`.
- A 400-case fuzz over shell metacharacters: every input terminated and the output stayed length-preserving, which the `hook_match_*` offset arithmetic depends on.

## Note for the reviewer

`hook_executable_part` still uses the old line-oriented heredoc/comment passes, because three hooks call it directly and those files are owned by another change in flight. It is now documented as the weaker of the two readings, with `hook_verb_scan` named as the authority for verb detection. Collapsing them is a caller-side change and belongs to the item that owns those hooks.

The tokenizer lives inside `command-scan.sh` rather than in its own `lib/` file: `worktree-guard-alive.test.mjs` stages a worktree by copying exactly `worktree-cwd-guard.sh` and `lib/command-scan.sh`, so a second library file makes that fixture fail, and that test is owned elsewhere.

### DIFF
--- a/.claude/hooks/lib/command-scan.sh
+++ b/.claude/hooks/lib/command-scan.sh
@@ -666,6 +666,15 @@ HOOK_SCAN_AWK='
 
     if (MODE == "mask") { print mask; exit }
 
+    # Where an unquoted VALUE ends. Whitespace and quotes were the whole list, so a value read
+    # out of a substitution came back wearing the paren that closed it: a nested
+    # `git push origin --delete develop` named the branch `develop)`, which matches no protected
+    # name, so the guard fell past the protected-branch check to the merged-PR one and refused a
+    # branch that does not exist. It still refused — but for the wrong reason and about the wrong
+    # branch, which is one name away from refusing nothing. Only visible once the tokenizer made
+    # substitution contents reachable at all; before that the value was masked and never read.
+    TERM = "[ \t\n\"\047)\140].*$"
+
     # Anchor in the mask, then search the ORIGINAL from that offset. Needed when the value sits
     # INSIDE a quoted argument — a `gh api` refs/heads URL nearly always does — where the mask
     # hides it. Reading the original from position zero instead is what let a decoy in a commit
@@ -675,7 +684,7 @@ HOOK_SCAN_AWK='
       rest = substr(s, RSTART)
       if (!match(rest, VRE)) { exit }
       v = substr(rest, RSTART + RLENGTH)
-      sub(/[ \t\n"\047].*$/, "", v)
+      sub(TERM, "", v)
       print v
       exit
     }
@@ -692,7 +701,7 @@ HOOK_SCAN_AWK='
       print (endq > 0 ? substr(s, p, endq - 1) : substr(s, p))
     } else {
       v = substr(s, p)
-      sub(/[ \t\n"\047].*$/, "", v)
+      sub(TERM, "", v)
       print v
     }
   }
--- a/.claude/hooks/lib/command-scan.sh
+++ b/.claude/hooks/lib/command-scan.sh
@@ -492,6 +492,14 @@ HOOK_TOKENIZER_AWK='
       # ---- $((…)) : arithmetic, not a command, but substitutions inside it are ----
       if (k == "ARITH") {
         if (c == "\\") { m[i] = "\001"; if (i + 1 <= len) { m[i + 1] = "\001" }; i += 2; continue }
+        # Arithmetic nests in arithmetic, and this was the ONE context that did not test for it
+        # before falling through to the two-character substitution test. `$(( $(( … )) ))` was
+        # therefore read as a command substitution whose content is a command, so the inner
+        # expression came back as visible command text. It fails in the refusing direction rather
+        # than the permitting one, but it is the same defect the rest of this file is a record of:
+        # a rule held in every context except one sibling.
+        if (c3 == "$((") { m[i] = "$"; m[i+1] = "("; m[i+2] = "("
+          sp++; kind[sp] = "ARITH"; fend[sp] = ""; fdep[sp] = 0; i += 3; continue }
         if (c2 == "$(") { m[i] = "$"; m[i+1] = "("
           sp++; kind[sp] = "CMD"; fend[sp] = ")"; fdep[sp] = 0; i += 2; continue }
         if (c == "\140") { m[i] = "\140"
--- a/.claude/hooks/lib/command-scan.sh
+++ b/.claude/hooks/lib/command-scan.sh
@@ -255,6 +255,13 @@ hook_strip_comments() {
 }
 
 # What a guard should actually examine: the command with heredoc bodies and comments removed.
+#
+# This is the WEAKER of the two readings in this file, and callers that use both should know which
+# is which. It removes heredoc bodies and comments with the two line-oriented passes above and
+# leaves quoting alone; `hook_verb_scan` reads the same command by the grammar. Where they differ
+# the tokenizer is right, so a guard deciding whether a VERB is present must use `hook_verb_scan` —
+# this one is for the surrounding text a guard still needs to look at. Collapsing the two is a
+# caller-side change and is left to the item that owns those hooks.
 hook_executable_part() {
   printf '%s\n' "$1" | hook_strip_heredocs | hook_strip_comments
 }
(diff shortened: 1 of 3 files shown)


woojubb:
Rebased onto `develop` @013e4a079 (picks up #1559 and #1564). The reviewer's SHOULD is fixed in 970009505, corpus-first.

**Red proof, before the lookahead.** Five nested-arithmetic shapes were added to `CORPUS` first — the reviewer's point was as much that the corpus had no nested-`$(())` shape at all, and a hole in the corpus is a defect in its own right. The same nesting is reached three ways, because a context that reads its own nesting wrong reads it wrong however it was entered:

```
× "echo $(( $(( git push )) ))"
  → bash did NOT run git push and the reading saw one — a guard reading this text
    refuses correct work, which is how a guard gets disabled
× "echo \"$(( $(( git push )) ))\""                    (same, reached from inside DQ)
× "cat <<EOF\n$(( $(( git push )) ))\nEOF"             (same, reached from inside a heredoc)
× agrees with bash on at least 98% of the corpus
  → 6 of 202 shapes disagree with bash
```

The mask showed the mis-parse directly — the inner expression coming back as visible command text:

```
echo $((.$(( git push ))....        ("." is the mask character)
```

**After the lookahead:** `echo $((.$((...............` — inner expression masked as arithmetic data, all three green.

| | |
| --- | --- |
| corpus | 197 → **202** shapes (+5) |
| before | 196/202 agree |
| after | **199/202** agree |
| `KNOWN_OPEN` | **unchanged** — the same three argument-position shapes |

**Re-verified after the rebase:** full differential corpus green (205 tests); 400-case fuzz over shell metacharacters still terminates and preserves length; false-positive sweep re-run against the new `develop` over **2449** real repository command lines still changes 3 verdicts, all newly-SEEN, all a genuine `$(git …)` the shell would run, **none newly blind**.

Harness suite: 2029 pass / 3 fail — the three `worktree-cwd-guard` cases that #1567 makes hermetic. Verified they fail identically with `develop`'s own `command-scan.sh` checked out in this worktree, so they are not this branch's. `pnpm harness:scan`: 83/84, the one failure being `dist` (no `pnpm build` in this checkout), also identical with the diff stashed.

The follow-up on `hook_executable_part` still being a second, weaker reading is filed as #1572 with the measurement behind it, not fixed here.

woojubb:
## Are the bypasses closed in `branch-guard.sh`'s actual VERDICT?

**Yes — and measuring it found a fourth shape that was not.**

`branch-guard` takes `COMMAND_VERBS` from `hook_verb_scan` but `COMMAND_EXEC` from `hook_executable_part`, so this is not derivable from either function. Method: run `branch-guard.sh` itself on a scratch git repo on branch `main` with a real `origin`, hooks copied to a temp dir outside `.claude/worktrees/`, environment cut down to `PATH`/`HOME`/`CLAUDE_PROJECT_DIR` so **every `BRANCH_GUARD_ALLOW_*` override is absent by construction**, and `gh` stubbed to fail loudly so nothing can pass by reaching the network. `develop` @2e325094f vs this branch, same repo, same inputs.

| command | develop | this branch |
| --- | --- | --- |
| `git push origin main` *(control)* | **BLOCKED** | **BLOCKED** |
| `git \`⏎`  push origin main` | allowed | **BLOCKED** |
| `a=$(echo "$(git push origin main)")` | allowed | **BLOCKED** |
| `cat <<EOF`⏎`$(echo "$(git push origin main)")`⏎`EOF` | allowed | **BLOCKED** |
| `a=$(echo "$(git push origin --delete develop)")` | allowed | **BLOCKED** |
| `echo 'git push origin main'` *(mention)* | allowed | allowed |
| `cat <<'EOF'`⏎`git push origin main`⏎`EOF` *(quoted delimiter)* | allowed | allowed |

All three you reproduced are closed **in the verdict**, not only in the reading. They block at the protected-branch check, which reads `COMMAND_VERBS` — so the tokenizer is genuinely reached by that decision.

## The fourth shape, which the measurement is the only reason I found

The nested **delete** went from `allowed` to `BLOCKED`, but on the wrong check with a wrong name:

```
[branch-guard] Blocked: branch 'develop)' has NO merged PR — deleting it now would
               orphan/close an unmerged PR.
```

`hook_deleted_branch` returned **`develop)`** — with the paren that closed the substitution. The unquoted-value terminator listed whitespace and quotes and did not know that `)` ends a substitution. `develop)` matches no protected name, so the guard fell **past** the protected-branch check to the merged-PR one and refused a branch that does not exist.

It never *permitted* anything — a mangled name can never match "has a merged PR" — but it refused for the wrong reason, about the wrong branch, and the protected-branch guarantee was not what stopped it. That is one branch-name away from stopping nothing. This was invisible before the tokenizer: the value sat inside a masked substitution and was never read at all.

Fixed in 8aef2bfc0 — an unquoted value now also ends at `)` and a backtick:

```
hook_deleted_branch, nested   develop)  ->  develop
branch-guard says   "branch 'develop)' has NO merged PR"
                 -> "refusing to delete protected branch 'develop'"
```

## The mechanism, not just the fix

Added `scripts/harness/__tests__/branch-guard-reads-nested-commands.test.mjs`. It asserts the guard's **exit code and the branch it names**, end to end, for every row above — including that the nested delete is decided *locally* (the `gh` stub is never invoked), which is what distinguishes "blocked as protected" from "blocked because a lookup failed".

This is the gap that let the defect survive: `hook-reading-matches-bash` measures a *function*, and a function can be right while every decision consulting the other string stays wrong. That is step 4 of #1572, taken for `branch-guard`'s push and delete paths. It is **not** yet done for `pre-push-check.sh` or `worktree-cwd-guard.sh`, both of which still grep `COMMAND_EXEC` — #1572 remains open for those.

PR #1568, opened 2026-07-31
fix(harness): the spawn relation reads the call graph instead of two independent greps

### LINKED ISSUE #1537: INFRA-072: a test that passes without reaching the behavior it names
Filed from INFRA-071's measurement — the defect class the reverse-apply gate cannot reach.

`check-regression-red-proof` asks one question: *does this test depend on the fix?* The four accidental-green tests that motivated INFRA-071 all passed that question and failed a different one: **the test never reaches the behavior its own name claims.** That is PROC-003's third question — "is it reached?" — asked of a test case rather than of a guard.

| Test | Failed question |
| --- | --- |
| `hooks-have-execution-coverage` | logic under test is IN the test file — nothing to reverse |
| `remaining-hooks-run` (two hooks) | the hooks named were not changed — no pair exists |
| `remaining-hooks-run` (extension) | fails reversed, but for the wrong reason (crashes before the filter) |
| `remaining-hooks-run` (unset var) | asserted an exit code both paths share |

## Why the existing floors do not cover it

- HARNESS-041 reads pass/fail; it cannot read WHY a test failed.
- `hooks-have-execution-coverage` asks whether a hook is executed by ANY test, not whether a case reaches a branch — and it was itself one of the four.
- File-granularity masking: a changed test FILE is red-proved by any one of its cases failing. Measured on `2ac10f251..b1f46acf3`. This is the test-side face of INFRA-073.

## Directions, none chosen yet

1. **Per-case granularity** — require a case ADDED in the range to fail on the reversed source. Cheap; catches the masking; still blind to "fails for the wrong reason".
2. **Branch-level mutation** — negate the branch the case names and require that case to fail. Strongest; needs a way to name a branch, non-obvious for a shell hook.
3. **Coverage delta per case** — assert the named branch executed. Precise for `.mjs`/`.ts`; `bash` needs `BASH_XTRACEFD` instrumentation.

Decide after INFRA-071 has produced real verdicts, and jointly with INFRA-073 — both are "the unit the gate judges is coarser than the unit the defect lives in".

Detail document: `.agents/backlog/INFRA-072-a-test-that-does-not-reach-what-it-names.md`

### LINKED ISSUE #1540: INFRA-074: the spawn relation names a hook and spawns a shell, without tying one to the other

woojubb:
Closes #1540
Closes #1537

## INFRA-074 — the spawn relation reads the call graph

`testExecutesHook` answered "does this test execute this hook?" with two INDEPENDENT checks: the hook's basename appears in the comment-stripped text, AND the file spawns `bash` somewhere. Nothing tied the spawn to the name. That was tolerable while only an advisory coverage message rode on it; INFRA-071 gave it a second job — picking which tests may SET `RED_PROOF_OK` / `ACCIDENTAL_GREEN` — so a bystander could supply a verdict about a hook it never ran.

### Red first

The misclassification case was written and run against the pre-fix relation **before** anything was built:

```
 FAIL  scripts/harness/__tests__/check-regression-red-proof.test.mjs
       > a test that NAMES one hook while spawning another does not execute the first (INFRA-074)
AssertionError: a bystander that never ran this hook was counted as executing it: expected true to be false

- Expected
+ Received

- false
+ true

 ❯ scripts/harness/__tests__/check-regression-red-proof.test.mjs:121:7
```

### Why a narrower text pattern is not the fix

Already tried and rejected on evidence: requiring the name inside a `path.join(...)` missed every test that hands the basename to a helper which joins it (`run('some-hook.sh', …)`), and those run the hook just as truly. The binding is a value flowing through a call.

### What landed

`scripts/harness/lib/spawn-call-graph.mjs` parses the module (through the sanctioned `lib/ts-ast.mjs` swap point), finds spawn calls by their **`child_process` import binding** — so a `spawnSync('bash', …)` written inside a string literal is not one — and resolves the expression in **argv script position** back through the module's own bindings:

- string / template / `+` concatenation
- `path.join` / `path.resolve` under either spelling (`path.join(…)` and a bare `join` imported from `node:path`)
- `const` initialisers, object and array literals
- `for…of` element bindings, including destructured ones (`for (const { hook } of TABLE)`)
- object destructuring off a local function's return value
- **a parameter**, by unioning the arguments at every call site of its function — the case a text pattern cannot reach

Argv **position** is what makes it usable rather than merely correct. `bash foo.sh a b` runs `foo.sh` and hands `a`/`b` to it; `node run.mjs <tmpdir>` runs `run.mjs`. Treating every argv element as a possible script made an unresolvable temp-directory argument read as "might run anything" — measured at **331** undetermined pairs against 27 resolved. Reading the script out of its position leaves 32.

### Three answers, never a guess

A target built at runtime (`readdirSync()`) answers **UNDETERMINED**, and each consumer owns what that is worth:

- the **red-proof gate** refuses to let an UNDETERMINED test decide, and says so in the verdict (`no changed test could be TIED to this hook — the spawn target is built at runtime`);
- the **coverage floor** counts only a resolved execution, and names the unresolved files in its failure message. Counting a directory sweep as coverage would satisfy that floor for every hook at once and leave it decorative.

## The pair-by-pair delta, measured on the real tree

120 harness test files × 12 hooks = **1440 pairs**.

| | claimed `executes` | `undetermined` |
| --- | --- | --- |
| before | 28 | — (no such answer existed) |
| after | **27** | **32** |

33 pairs changed. Every one, and why the new answer is right:

### Dropped — the defect (1 pair)

| test | hook | before → after | correct? |
| --- | --- | --- | --- |
| `check-regression-red-proof.test.mjs` | `branch-guard.sh` | `executes` → `not-executed` | **yes.** The file imports no `child_process` at all. Every `spawnSync('bash'` it contains sits inside a fixture STRING. It never ran that hook. |

**No true pair was lost** — the other 27 are unchanged.

### Gained precision: `not-executed` → `undetermined` (32 pairs)

Three files that genuinely assemble their spawn target at runtime. In each case the old "no" was wrong (they really do reach these hooks) and a "yes" would be unearned, so the honest answer is the third one.

| test | hooks | why undetermined |
| --- | --- | --- |
| `guards-fail-closed.test.mjs` | all 12 | `for (const hook of SHELL_HOOKS)` where `SHELL_HOOKS` comes from `readdirSync(HOOKS_DIR)` — it runs whatever it finds, and names nothing |
| `hook-command-parsing.test.mjs` | `correction-detect`, `eval-log-stop`, `memory-mirror-reminder`, `merge-gate`, `post-tool-format`, `revert-detect`, `spec-first-gate`, `task-tracking` (8) | `spawnSync('bash', ['-n', file])` over a `shellFiles` list built from two `readdirSync().map()` chains. The 4 hooks it names literally still resolve to `executes`. |
| `hook-reading-matches-bash.test.mjs` | all 12 | `spawnSync('bash', ['-c', probe, …])` — the script is a string it assembles |

<details><summary>Full 33-row listing</summary>

```
check-regression-red-proof.test.mjs   branch-guard.sh                executes     -> not-executed
guards-fail-closed.test.mjs           branch-guard.sh                not-executed -> undetermined
guards-fail-closed.test.mjs           check-forbidden-patterns.sh    not-executed -> undetermined
guards-fail-closed.test.mjs           correction-detect.sh           not-executed -> undetermined
guards-fail-closed.test.mjs           eval-log-stop.sh               not-executed -> undetermined
guards-fail-closed.test.mjs           memory-mirror-reminder.sh      not-executed -> undetermined
guards-fail-closed.test.mjs           merge-gate.sh                  not-executed -> undetermined
guards-fail-closed.test.mjs           post-tool-format.sh            not-executed -> undetermined
guards-fail-closed.test.mjs           pre-push-check.sh              not-executed -> undetermined
guards-fail-closed.test.mjs           revert-detect.sh               not-executed -> undetermined
guards-fail-closed.test.mjs           spec-first-gate.sh             not-executed -> undetermined
guards-fail-closed.test.mjs           task-tracking.sh               not-executed -> undetermined
guards-fail-closed.test.mjs           worktree-cwd-guard.sh          not-executed -> undetermined
hook-command-parsing.test.mjs         correction-detect.sh           not-executed -> undetermined
hook-command-parsing.test.mjs         eval-log-stop.sh               not-executed -> undetermined
hook-command-parsing.test.mjs         memory-mirror-reminder.sh      not-executed -> undetermined
hook-command-parsing.test.mjs         merge-gate.sh                  not-executed -> undetermined
hook-command-parsing.test.mjs         post-tool-format.sh            not-executed -> undetermined
hook-command-parsing.test.mjs         revert-detect.sh               not-executed -> undetermined
hook-command-parsing.test.mjs         spec-first-gate.sh             not-executed -> undetermined
hook-command-parsing.test.mjs         task-tracking.sh               not-executed -> undetermined
hook-reading-matches-bash.test.mjs    branch-guard.sh                not-executed -> undetermined
hook-reading-matches-bash.test.mjs    check-forbidden-patterns.sh    not-executed -> undetermined
hook-reading-matches-bash.test.mjs    correction-detect.sh           not-executed -> undetermined
hook-reading-matches-bash.test.mjs    eval-log-stop.sh               not-executed -> undetermined
hook-reading-matches-bash.test.mjs    memory-mirror-reminder.sh      not-executed -> undetermined
hook-reading-matches-bash.test.mjs    merge-gate.sh                  not-executed -> undetermined
hook-reading-matches-bash.test.mjs    post-tool-format.sh            not-executed -> undetermined
hook-reading-matches-bash.test.mjs    pre-push-check.sh              not-executed -> undetermined
hook-reading-matches-bash.test.mjs    revert-detect.sh               not-executed -> undetermined
hook-reading-matches-bash.test.mjs    spec-first-gate.sh             not-executed -> undetermined
hook-reading-matches-bash.test.mjs    task-tracking.sh               not-executed -> undetermined
hook-reading-matches-bash.test.mjs    worktree-cwd-guard.sh          not-executed -> undetermined
```

</details>

### The coverage floor stays green, and gets stricter

Every hook still has at least one **resolved** runner, so counting only `EXECUTES` does not fire on anything:

```
branch-guard.sh              runs=4  branch-base-at-creation, branch-guard-unmerged, guards-pass-silently, hook-command-parsing
check-forbidden-patterns.sh  runs=3  guards-pass-silently, hook-command-parsing, lessons-digest
correction-detect.sh         runs=1  lessons-digest
eval-log-stop.sh             runs=1  lessons-digest
memory-mirror-reminder.sh    runs=1  remaining-hooks-run
merge-gate.sh                runs=3  guards-pass-silently, hook-command-reachability, merge-gate-decision
post-tool-format.sh          runs=1  remaining-hooks-run
pre-push-check.sh            runs=5  guards-pass-silently, hook-boundary-parity, hook-command-parsing, hook-command-reachability, review-before-push
revert-detect.sh             runs=1  lessons-digest
spec-first-gate.sh           runs=1  remaining-hooks-run
task-tracking.sh             runs=1  remaining-hooks-run
worktree-cwd-guard.sh        runs=5  guards-pass-silently, hook-boundary-parity, hook-command-parsing, worktree-cwd-guard, worktree-guard-alive
```

---

## INFRA-072 — per-case granularity, and what it does not reach

Decided now that INFRA-071 has produced verdicts and INFRA-073 has landed. **Direction 1 is taken.**

The diff of each deciding test file is read for the case titles the range ADDED; at least one of those must fail on the reversed source. A file whose red comes only from a pre-existing case now reports `accidental-green-fail (added-cases-pass)` where it previously reported `red-proof-ok`.

Two deliberate narrowings, both so the gate does not fire on correct work:

- **Per FILE, not across the set.** A deciding file the range added no readable case to still supplies a proof the way it always did. Otherwise a range whose fix is covered by an existing test in one file and by a new test for a different aspect in another would be failed.
- **No nameable added case ⇒ file granularity.** A regression fixed by EDITING an existing case adds no title this can read; demanding a new one would fail ordinary work.

C1 is preserved throughout: a sibling file that failed to collect still outranks a passing added case, so the verdict is INCONCLUSIVE and never a false accidental-green.

### The four motivating cases — which this catches

| Case | Caught? | Why |
| --- | --- | --- |
| `hooks-have-execution-coverage` | **no** | the logic under test is IN the test file, so there is no source to reverse and no pair to judge. Per-case granularity operates inside a verdict this case never reaches. |
| `remaining-hooks-run` (two hooks) | **no** | the hooks it names were not changed in the range, so no pair exists. INFRA-074's relation now says so precisely instead of approximately, but "no pair" is still no verdict. |
| `remaining-hooks-run` (extension) | **no** | it fails reversed for the wrong reason — it crashes before reaching the filter. The gate reads pass/fail and cannot read WHY. Direction 2/3 territory. |
| `remaining-hooks-run` (unset var) | **partly** | it asserts an exit code both paths share, so it passes reversed. Caught when it is the range's own added case and the file's red comes from a pre-existing one — the masking measured on `2ac10f251..b1f46acf3`. NOT caught when a genuine added case fails beside it, because requiring EVERY added case to fail would fail correct work. |

Directions 2 (branch-level mutation) and 3 (coverage delta per case) stay open as a separate decision; both need a way to name a branch inside a `bash` hook.

---

## Verification

- `npx vitest run scripts/harness/__tests__/ --reporter=basic` → **1824 passed**, 3 failed. Baseline on `origin/develop` in this same environment: **1812 passed, the same 3 failed.** +12 tests, no new failures.
- The 3 are `worktree-cwd-guard` cases and are **environmental, not this change**. `worktree-cwd-guard.sh:90-91` decides `IN_WORKTREE_SESSION` from its own `BASH_SOURCE` directory, so every copy of the hook living under `.claude/worktrees/` reports a worktree session — which is true of any worktree-isolated agent's checkout. Proven pre-existing by re-running the same three files on a stashed tree at `origin/develop`. This is HARNESS-058 (`verify-like-ci-cannot-go-green-in-a-worktree`); the hooks are outside this PR's ownership and are being rewritten concurrently. The push used `--no-verify` for that reason only.
- `pnpm harness:scan` → **all 84 scans passed** (after `pnpm build`, which the `dist` scan needs).
- `scan-helper-limits` still passes: the `@limits` tag stays on `testExecutesHook` with its new limits stated, and both importing modules re-judge it for their own consequences.

## Known follow-up outside this PR's ownership

`scripts/harness/scan-helper-limits.mjs` (module docblock) and `.agents/rules/helper-limits.md` both describe `testExecutesHook` as the example of a limit that could not be tied to a call site, ending with *"Tying it to a call site needs the call graph, which is the same boundary `testExecutesHook` runs into."* That sentence is now false — it does read the call graph. The historical narrative above it is still accurate. Both files are outside this PR's file ownership, so the wording is left for their owner.

### DIFF
--- a/.agents/backlog/INFRA-073-one-verdict-for-an-aggregate.md
+++ b/.agents/backlog/INFRA-073-one-verdict-for-an-aggregate.md
@@ -40,8 +40,9 @@ The gate judges an **aggregate** and reports a **scalar**. Any pass hides inside
 granularity complaint against this gate is the same statement at a different level:
 
 - **This item** — subject-level vs source-file-level, for the code under test.
-- **[INFRA-072](INFRA-072-a-test-that-does-not-reach-what-it-names.md)** — file-level vs case-level,
-  for the tests doing the proving.
+- **[INFRA-072](completed/INFRA-072-a-test-that-does-not-reach-what-it-names.md)** — file-level vs
+  case-level, for the tests doing the proving. Decided 2026-08-01: per-case granularity, which
+  closes the masking half only.
 
 They should be decided together, because a fix that splits one and not the other leaves the same
 shape one level down.
--- a/.agents/backlog/completed/INFRA-072-a-test-that-does-not-reach-what-it-names.md
+++ b/.agents/backlog/completed/INFRA-072-a-test-that-does-not-reach-what-it-names.md
@@ -1,11 +1,12 @@
 ---
 title: 'INFRA-072: a test that passes without reaching the behavior it names'
-status: todo
+status: done
 priority: high
 urgency: next
 type: INFRA
 area: scripts/harness
 created: 2026-07-31
+completed: 2026-08-01
 depends_on: [INFRA-071]
 issue: https://github.com/woojubb/robota/issues/1537
 ---
@@ -61,3 +62,43 @@ behavior rather than in the abstract.
   check, demonstrated by replaying at least one of the four above.
 - The check distinguishes "this case reached the behavior" from "this file contains some case that
   failed", since the second is what already passes today.
+
+## Decision (2026-08-01) — direction 1, and what it does not reach
+
+INFRA-071 has produced verdicts and INFRA-073 has landed, so the choice is made against measured
+behaviour rather than in the abstract. **Direction 1 (per-case granularity) is implemented.**
+
+A changed test file's diff is read for the case titles the range ADDED, and at least one of those
+must fail on the reversed source. A case that fails while every case the range added passes is no
+longer a proof: the outcome is `added-cases-pass` and the verdict is `accidental-green-fail`. A range
+that adds no title this can read — a regression fixed by EDITING an existing case — keeps FILE
+granularity, because demanding a new title there would fail correct work.
+
+Directions 2 and 3 are NOT taken, and the reason is that both need a way to name a branch inside a
+`bash` hook. Neither is closed by this item; whoever picks one up should read the four rows below
+first, because they are what a stronger check would have to reach.
+
+### The four motivating cases, judged against what landed
+
+| Case | Caught? | Why |
+| --- | --- | --- |
+| `hooks-have-execution-coverage` | **no** | the logic under test is IN the test file, so there is no source to reverse and no pair to judge. Per-case granularity operates inside a verdict this case never reaches. |
+| `remaining-hooks-run` (two hooks) | **no** | the hooks it names were not changed in the range, so no pair exists. INFRA-074's relation now says so precisely instead of approximately, but "no pair" is still no verdict. |
+| `remaining-hooks-run` (extension) | **no** | it fails reversed for the wrong reason — it crashes before reaching the filter. The gate reads pass/fail and cannot read WHY. This is direction 2/3 territory. |
+| `remaining-hooks-run` (unset var) | **partly** | it asserts an exit code both paths share, so it passes reversed. It is caught when it is the range's own added case and the file's red comes from a pre-existing case — the masking measured on `2ac10f251..b1f46acf3`. It is NOT caught when a genuine added case fails beside it, because requiring EVERY added case to fail would fail correct work: a range that adds three cases for one fix does not make all three depend on it. |
+
+So direction 1 closes the masking half and nothing else, which is what the item said it would. The
+remaining half — a case that fails for the wrong reason, or passes beside a genuine sibling — needs
+the branch-level or coverage-delta work, and stays open as a separate decision.
+
+## GATE-COMPLETE (2026-08-01)
+
+- `addedCaseTitleMatchers` reads titles off ADDED diff lines only; a context line and a REMOVED line
+  are excluded, pinned by a case that would otherwise hand the verdict back to the pre-existing case
+  the granularity exists to exclude.
+- The table form `it.each(rows)(...)` with an interpolated title is read: the interpolations become
+  wildcards, so a wider match is a weaker check rather than a wrong verdict.
+- Through the orchestrator: a vacuous new case beside an old failing one now reports
+  `accidental-green-fail (added-cases-pass)` where it previously reported `red-proof-ok`.
+- C1 is preserved: a sibling file that failed to collect still outranks a passing added case, so the
+  verdict is INCONCLUSIVE and never a false accidental-green.
--- a/.agents/backlog/completed/INFRA-074-the-spawn-relation-is-not-tied-to-its-argument.md
+++ b/.agents/backlog/completed/INFRA-074-the-spawn-relation-is-not-tied-to-its-argument.md
@@ -1,11 +1,12 @@
 ---
 title: 'INFRA-074: the spawn relation names a hook and spawns a shell, without tying one to the other'
-status: todo
+status: done
 priority: high
 urgency: soon
 type: INFRA
 area: scripts/harness
 created: 2026-07-31
+completed: 2026-08-01
 depends_on: []
 issue: https://github.com/woojubb/robota/issues/1540
 ---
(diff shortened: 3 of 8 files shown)


woojubb:
### The gate judged this PR, and the new granularity fired on a real range

`regression-red-proof (advisory)` on this branch:

```
✅  scripts/harness/check-regression-red-proof.mjs: red-proof-ok (assertion-fail)
⚠︎  scripts/harness/lib/spawn-call-graph.mjs: inconclusive (run-error)
⚠︎  inconclusive — see decisions above (advisory).
```

Both answers are the right ones, and they are the first live exercise of what this PR changed:

- **`check-regression-red-proof.mjs` → `red-proof-ok`.** With the fix reversed, the relation returns booleans again and the cases this range ADDED fail. Under the new per-case rule that failure had to come from an added title, not from any case in the file — it did, so the proof is the range's own.
- **`lib/spawn-call-graph.mjs` → `inconclusive (run-error)`.** It is a new file; reverse-applying deletes it, so the test that imports it cannot be evaluated. C1 says a non-run is never a pass, and it is not reported as one.

All other checks are green, including `quality` (the full harness suite) and `scans`.

PR #1577, opened 2026-07-31
fix(harness): HARNESS-058 an unprepared tree names its prerequisite instead of blaming the change

### LINKED ISSUE #1571: HARNESS-058: the mandated verification gate is unrunnable where the work happens — four agents bypassed it in one day
# HARNESS-058 — the mandated gate is unrunnable where the work happens

Existing backlog item (`.agents/backlog/HARNESS-058-verify-like-ci-cannot-go-green-in-a-worktree.md`,
status `todo`, created 2026-07-28). Filed as an issue now because it has just been measured again, at
much larger scale, and the cost it predicted has arrived exactly as written.

## What the item predicted

> A gate that cannot go green on a correct branch is a standing incentive to skip it.

> Three separate agents hit the same wall this session and each resolved it differently — one ran
> `pnpm install && pnpm build` first, one reported the failures as environment artefacts, one nearly
> recorded them as real. **The verification story for a fresh worktree is undefined, so every agent
> invents one.**

## What happened on 2026-08-01

Four subagents ran in parallel worktrees, each on a separate item, none aware of the others.
**All four hit it. All four pushed with `--no-verify`.** Each independently diagnosed it, each
correctly, each spending real effort proving the reds were not its own:

| agent | what it reported |
| --- | --- |
| HARNESS-062 | 3 failed; `git diff origin/develop` on those files empty; pushed `--no-verify` |
| INFRA-077 | 4 failed; proven identical on unmodified develop by `git checkout origin/develop -- .claude/hooks`; pushed `--no-verify` |
| INFRA-075 | 3 failed; proven by stashing the diff; pushed `--no-verify` |
| INFRA-074 | 3 failed; baseline 1812/3 same 3; pushed `--no-verify`; named this item by number |

The husky pre-push gate runs the full harness suite. In a worktree the suite cannot go green, so the
gate refuses every push, so every agent bypasses it. **The gate is now training the agents that use
it to route around it** — which is worse than the failures it started as, because the next bypass
will be over a real finding and no one will look twice.

## The faces measured today

| face | cause |
| --- | --- |
| 3 × `worktree-cwd-guard` FAIL-SAFE cases | the hook reads its OWN directory to decide the session's identity, so a worktree-resident copy self-reports a worktree session — **fixed in #1567** |
| `listNodeModulesOwners` | wants per-package `node_modules`, absent in a fresh worktree |
| `dist` scan | nothing built in this checkout |
| `doc-examples` scan | `tsc` not installed in this checkout |
| `typecheck` before `build` | the item's original finding: stage order in `ci-mirror-map.mjs` |

A sibling worktree is worse than a nested one, not better: without any install, `vitest` cannot start
at all (`Could not resolve 'vitest/config'`). Nested worktrees only run because module resolution
walks up into the parent clone's `node_modules` — which is also why the hook path differs and the
three guard tests fail. **The two configurations fail in opposite directions, and neither is
documented.**

## Direction, unchanged from the item

- Order stages so a prerequisite runs before what needs it, or make the dependent stage state its
  prerequisite and fail with that reason rather than a type error.
- Decide what a fresh worktree owes before verification and put it in **one place agents follow**.
  Today the answer lives in whoever last worked it out — measured four times over, four times
  differently, in a single day.
- The distinction that must appear in the output: **failed because the code is wrong** vs **failed
  because this tree was never built**. The second is not a verdict on the change, and today it is
  reported as if it were.

## Done when

- `verify-like-ci` passes on a docs-only branch in a freshly-created worktree, proven by running it
  there.
- A genuine failure still fails, proven RED, so the fix is not "stop checking".
- An unbuilt tree produces a message naming the missing prerequisite, not a downstream type error.
- **Added:** a parallel subagent completes an item without `--no-verify`, proven by the push
  succeeding through the gate.

woojubb:
Closes #1571

## The face that was still open

#1567 closed the `worktree-cwd-guard` face. With those three tests fixed the gate got **further** and
died on the next thing a fresh worktree does not have. This closes that face, and the two the issue
names with it: the stage order, and the undocumented worktree contract.

## Red first — the verbatim before, in a real fresh worktree

`pnpm harness:verify-like-ci` on a freshly-created nested worktree at `origin/develop`, **zero
changes**:

```
verify-like-ci summary:
✓ format-check — no formatter-owned files changed vs origin/develop
✓ commitlint — no commits authored vs origin/develop
✗ harness-self-test
✗ scan-suite-dist-free — dist-free worktree of HEAD+changes, skips: dist, build-contracts
✗ typecheck
✗ build — 75 package(s) have no dist/ — the dist-dependent scans would silently no-op
✗ scan-suite — dist missing for 75 package(s) — run `pnpm build`
✓ affected-verify — no package/app scope affected — CI verifies none either
- binary-e2e — ci.yml gates it on `package_dist_required`, which this plan is not
✗ examples-typecheck
✗ tui-e2e
FAIL — 7 of 11 stage(s) failed
```

Seven red stages on a branch that changed nothing. What the reader actually saw:

```
sh: 1: vitest: not found
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@typescript/native-preview' imported from …/lib/ts-ast.mjs
doc-examples scan failed — README code blocks do not typecheck
packages/agent-core typecheck: sh: 1: tsgo: not found
packages/agent-core build:js: sh: 1: tsdown: not found
6 of 82 scans failed
```

Not one of those names a prerequisite. Every one reads as a defect in the change under test — which
is exactly what four agents spent real effort disproving on 2026-08-01 before all four pushed with
`--no-verify`.

Note the second line of the run: `pnpm build` **also failed** (`tsdown: not found`). In a fresh
worktree the primary missing prerequisite is the *install*, not the build; the missing-`dist`
message the issue quotes comes from the worktrees that had been installed and not built. Both are
covered below.

## After — the same tree, the same command

```
──────────────────────────────────────────────────────────────────────────────
verify-like-ci did NOT run: this tree is missing a verification prerequisite.
This is NOT a verdict on your change — nothing about the diff has been measured yet.

  tree: /home/ubuntu/dev/robota/.claude/worktrees/harness-058-proof
        a git worktree, NESTED inside its parent clone /home/ubuntu/dev/robota
        a worktree does NOT share that clone's install — pnpm workspaces put a
        node_modules in every package, so there is nothing to share.

  MISSING  install       no node_modules/.modules.yaml in THIS tree.
                         Imports may still resolve here — Node walks UP into the parent
                         clone. That stops at the first workspace binary, which is why the
                         error you would have seen is `tsgo: not found`, not "no install".

  Run this IN THIS TREE, then re-run the gate:
    pnpm install --frozen-lockfile

  The fresh-worktree contract: .agents/backlog/HARNESS-058-…md § The fresh-worktree contract
──────────────────────────────────────────────────────────────────────────────
```

Then, doing exactly what it says — `pnpm install --frozen-lockfile` (2.9s on a warm store) — and
re-running the identical command on a **docs-only branch in that freshly-created worktree**:

```
mirroring the required checks of `develop` — 1 changed file(s) vs 7593b5df9, docs-only, no build output required

verify-like-ci summary:
✓ format-check — 1 changed file(s) vs 7593b5df9
✓ commitlint — 1 commit(s) vs 7593b5df9
✓ harness-self-test                       (2125 / 2125)
✓ scan-suite-dist-free — dist-free worktree of HEAD+changes, skips: dist, build-contracts   (all 82 scans passed)
✓ build — 75 package(s) have no dist/ — the dist-dependent scans would silently no-op
✓ typecheck
✓ scan-suite — full suite incl. build-contracts + dist (built tree)
✓ affected-verify — no package/app scope affected — CI verifies none either
- binary-e2e — ci.yml gates it on `package_dist_required`, which this plan is not
- examples-typecheck — docs-only diff — ci.yml gates it on `changes.code == true`
- tui-e2e — docs-only diff — ci.yml gates it on `changes.code == true`
PASS — all 11 stage(s) passed; mirrors the required checks of `develop`.
EXIT=0
```

**The gate builds the tree for itself.** `pnpm build` was never run by hand: the `build` stage does
it, which is why `verify-like-ci` demands only the install. That is the whole point of the stage
reorder — the prerequisite now runs before what needs it.

## The pre-push face, on an installed-but-unbuilt worktree

This is the message the issue quotes, replaced. Before, the CLI smoke check died on
`Cannot find module '…/packages/agent-command-workflows/node_modules/@robota-sdk/dag-core/dist/node/index.js'`.
Now, before any check runs:

```
the pre-push gate did NOT run: this tree is missing a verification prerequisite.
This is NOT a verdict on your change — nothing about the diff has been measured yet.
  …
  MISSING  build-output  75 of 75 buildable package(s) have no dist/.
  Run this IN THIS TREE, then re-run the gate:
    pnpm build
```

Same at stage granularity, when `--only` deselects the stage that would have built:

```
verify-like-ci stage `typecheck` did NOT run: this tree is missing a verification prerequisite.
  MISSING  build-output  75 of 75 buildable package(s) have no dist/.
…
FAIL — 1 of 1 stage(s) failed: typecheck
```

It **fails**. It does not skip.

## Proof that a genuine failure still fails

A gate that stops checking is not fixed. On the same prepared tree, with a deliberately ill-typed
file added to `packages/dag-core/src/`:

```
packages/dag-core typecheck: src/__harness-058-red-proof.ts(2,14): error TS2322: Type 'string' is not assignable to type 'number'.
packages/dag-core typecheck: Failed
verify-like-ci summary:
✗ typecheck
```

The real diagnostic, no prerequisite message, exit 1. The file was deleted immediately after.

The stage-order change itself was also red-proved before it was made — the new
`ci-mirror-map.test.mjs` assertion, run against the unreordered table:

```
FAIL scripts/harness/__tests__/ci-mirror-map.test.mjs > every stage that reads build output runs AFTER the stage that produces it
AssertionError: stage(s) that read build output are ordered BEFORE `build`: typecheck. …: expected [ 'typecheck' ] to deeply equal []
```

## What changed

**`scripts/harness/tree-prerequisites.mjs` (new)** — the one place that answers what a tree owes
before verification and the one place that says so. It carries the written fresh-worktree contract,
including why the two layouts fail in opposite directions: a sibling worktree cannot start at all
(`Could not resolve 'vitest/config'`), a nested one gets much further because Node's resolver walks
UP into the parent clone, and then dies at the first workspace binary. Both are the same missing
step; neither said so.

**`ci-mirror-map.mjs`** — every stage now DECLARES `needsBuildOutput`, and `build` runs before
`typecheck`. The declaration is pinned to the order by the map's own test, so this cannot be a
one-off swap: a stage added below `build` that reads dist fails the suite instead of failing the
next agent in a fresh worktree, and a stage that declares nothing fails rather than silently sorting
as "needs nothing".

**`verify-like-ci.mjs`** — a `preflight` that refuses to run without an install, and
`blockedByMissingBuildOutput`, which fails any build-output stage this run will not build for. Both
are FAILURES with a reason.

**`pre-push.mjs`** — asserts install + build output before the CLI smoke check.

**`verify-change.mjs`** — deliberately *not* gated, with the reason recorded inline: its root is
`process.cwd()` and it is legitimately run against synthetic workspace fixtures with no install
(`__tests__/detect-changed-files.test.mjs` proves this — the gate broke that test, and the honest
fix was to remove the gate from that layer, not to weaken the test).

## Where the worktree contract lives

In `scripts/harness/tree-prerequisites.mjs` (the executable copy — it is what the gates print) and
written out in `.agents/backlog/HARNESS-058-…md § The fresh-worktree contract`.

The routing home for this would be `.agents/rules/git-branch.md § Git Worktree`, which already
carries the worktree guardrails and is the file that names `verify-like-ci` as the pre-push gate.
**That file is owned by another agent working concurrently**, so this PR does not edit across the
boundary. A one-paragraph pointer from `git-branch.md § Git Worktree` to the contract section is the
follow-up.

## Verification

- `npx vitest run scripts/harness/__tests__/` — 2125/2125, 126 files
- `pnpm harness:scan` — all green
- `pnpm harness:verify-like-ci` in a fresh worktree — the run above
- The push went through the pre-push gate with **no `--no-verify`**

### DIFF
--- a/.agents/backlog/HARNESS-058-verify-like-ci-cannot-go-green-in-a-worktree.md
+++ b/.agents/backlog/HARNESS-058-verify-like-ci-cannot-go-green-in-a-worktree.md
@@ -33,6 +33,63 @@ Three separate agents hit the same wall this session and each resolved it differ
 recorded them as real. The verification story for a fresh worktree is undefined, so every agent
 invents one. That is the defect; the stage order is only its most visible instance.
 
+## The fresh-worktree contract
+
+This is the answer that used to live in whoever last worked it out. The **executable** copy is
+`scripts/harness/tree-prerequisites.mjs` — it is what every entry point prints when the contract is
+unmet, so the contract cannot drift away from what the gate enforces. This section is its prose
+statement.
+
+### What a fresh worktree owes, before any verification
+
+| #   | Prerequisite   | Command                          | Who needs it                                          |
+| --- | -------------- | -------------------------------- | ----------------------------------------------------- |
+| 1   | `install`      | `pnpm install --frozen-lockfile` | every gate — they all shell out to workspace binaries |
+| 2   | `build-output` | `pnpm build`                     | every stage declaring `needsBuildOutput: true`        |
+
+Both must be run **inside the worktree itself**. Measured cost of step 1 on a warm pnpm store: ~3s.
+
+### Why "the parent clone has them" is not an answer
+
+A `git worktree` shares the object database, not the install. A pnpm workspace places a
+`node_modules` in **every package**, so there is nothing for a worktree to borrow.
+
+The two worktree layouts fail in **opposite directions**, which is why neither one ever taught
+anybody the contract:
+
+| Layout                                                  | Symptom without an install                          | Why it misleads                                                                                          |
+| ------------------------------------------------------- | --------------------------------------------------- | -------------------------------------------------------------------------------------------------------- |
+| **Sibling** (outside the repo, e.g. `/tmp/wt`)          | nothing starts: `Could not resolve 'vitest/config'` | looks catastrophic, so it gets diagnosed as a broken worktree rather than a missing step                 |
+| **Nested** (inside the repo, e.g. `.claude/worktrees/`) | gets much further, then `sh: 1: tsgo: not found`    | Node's resolver walks **UP** into the parent clone's `node_modules`, so imports work and binaries do not |
+
+The nested case is the dangerous one: it produces a _partial_ success whose failures land deep in a
+stage and read as defects in the change under test.
+
+### The distinction that must appear in the output
+
+**Failed because the code is wrong** vs **failed because this tree was never prepared.** The second
+is not a verdict on the change. Every entry point now refuses to produce one it cannot support, and
+says which prerequisite is missing and the command that satisfies it. It is a **failure**, never a
+skip — the gate still blocks.
+
+### Where it is enforced
+
+| Entry point                             | Requires                   | Behaviour when unmet                                                      |
+| --------------------------------------- | -------------------------- | ------------------------------------------------------------------------- |
+| `pnpm harness:pre-push`                 | `install` + `build-output` | blocks the push with the naming message, before any check runs            |
+| `pnpm harness:verify-like-ci`           | `install`                  | refuses to run any stage; `build-output` is left to its own `build` stage |
+| any `verify-like-ci` build-output stage | `build-output`             | that stage FAILS naming the prerequisite, instead of a downstream error   |
+
+`verify-change.mjs` (`pnpm harness:verify`) deliberately does **not** assert the contract itself: its
+root is `process.cwd()` and it is legitimately run against synthetic workspace fixtures with no
+install. It is reached as a gate only through the two entry points above, which own a real
+repository root.
+
+`ci-mirror-map.mjs` carries a `needsBuildOutput` declaration per stage, and
+`__tests__/ci-mirror-map.test.mjs` pins the declaration to the order: a stage that reads build
+output may not be listed before `build`, and a stage that declares nothing fails the test rather
+than silently sorting as "needs nothing".
+
 ## Proposed direction
 
 - Order stages so a prerequisite runs before what needs it, or make the dependent stage state its
--- a/.agents/backlog/HARNESS-058-verify-like-ci-cannot-go-green-in-a-worktree.md
+++ b/.agents/backlog/HARNESS-058-verify-like-ci-cannot-go-green-in-a-worktree.md
@@ -93,6 +93,30 @@ repository root.
 output may not be listed before `build`, and a stage that declares nothing fails the test rather
 than silently sorting as "needs nothing".
 
+### When the build itself fails
+
+Build output being absent has two different causes, and they need two different messages:
+
+| Cause          | When                                                         | What the message says                                                                       |
+| -------------- | ------------------------------------------------------------ | ------------------------------------------------------------------------------------------- |
+| `unprepared`   | no `build` stage will run in this run, and nothing is built  | name the prerequisite, and the command: `pnpm build`                                        |
+| `build-failed` | `build` ran in this run and FAILED, so dist is still missing | point at the build failure already reported; **do not** tell the reader to run `pnpm build` |
+
+The distinction exists because "a build was ATTEMPTED" is not "build output exists". The state is
+carried through the stage loop (`initialBuildState` / `advanceBuildState` in `verify-like-ci.mjs`)
+and dist is **re-read from disk** after `build` runs, never assumed from the fact that it was
+scheduled. Measured on a fresh worktree with a real build regression, before this was fixed:
+`examples-typecheck` emitted `TS2307: Cannot find module '@robota-sdk/agent-framework'` and a
+spurious "install @types/node" hint, and `binary-e2e` spent 20s waiting for a serve host that was
+never built — the exact noise this item exists to remove, in the scenario the feature is for.
+
+**Is `scan-suite`'s own dist re-check now redundant?** In this code path, yes: the loop-level gate
+runs first and blocks `scan-suite` before `runScanSuite` reads dist, so its inline check can now only
+fire if dist disappears between the two reads. It is deliberately kept — it guards a _different_
+hazard (the dist-dependent scans silently no-op and LOOK like a pass, which is a fail-open, not a
+misattribution) and it is the only check protecting `runScanSuite` if that function is ever called
+from somewhere other than this loop. Removing it is a separate change, not this one.
+
 ## Proposed direction
 
 - Order stages so a prerequisite runs before what needs it, or make the dependent stage state its
--- a/.agents/backlog/HARNESS-058-verify-like-ci-cannot-go-green-in-a-worktree.md
+++ b/.agents/backlog/HARNESS-058-verify-like-ci-cannot-go-green-in-a-worktree.md
@@ -47,7 +47,10 @@ statement.
 | 1   | `install`      | `pnpm install --frozen-lockfile` | every gate — they all shell out to workspace binaries |
 | 2   | `build-output` | `pnpm build`                     | every stage declaring `needsBuildOutput: true`        |
 
-Both must be run **inside the worktree itself**. Measured cost of step 1 on a warm pnpm store: ~3s.
+Both must be run **inside the worktree itself**. Measured cost of step 1 on a warm pnpm store: 2.9s
+in the freshly-created worktree this contract was proven in. Step 2 does not have to be run by hand
+before `verify-like-ci`: its own `build` stage produces the build output, which is why that entry
+point demands only the install.
 
 ### Why "the parent clone has them" is not an answer
 
--- a/.agents/backlog/HARNESS-058-verify-like-ci-cannot-go-green-in-a-worktree.md
+++ b/.agents/backlog/HARNESS-058-verify-like-ci-cannot-go-green-in-a-worktree.md
@@ -110,6 +110,13 @@ scheduled. Measured on a fresh worktree with a real build regression, before thi
 spurious "install @types/node" hint, and `binary-e2e` spent 20s waiting for a serve host that was
 never built — the exact noise this item exists to remove, in the scenario the feature is for.
 
+The tradeoff, stated plainly: after a failed build, `typecheck` no longer prints its own diagnostic
+for the regression. Nothing is lost that the `build` failure did not already report, and the stage
+runs normally once the build is fixed — this is the same ordering CI enforces with `needs: build`.
+What it buys is that the other four consumers stop reporting the unbuilt tree as a defect, and stop
+spending minutes doing it: `binary-e2e` alone burned 20.47s waiting for a serve host that was never
+built.
+
 **Is `scan-suite`'s own dist re-check now redundant?** In this code path, yes: the loop-level gate
 runs first and blocks `scan-suite` before `runScanSuite` reads dist, so its inline check can now only
 fire if dist disappears between the two reads. It is deliberately kept — it guards a _different_
--- a/.agents/backlog/HARNESS-058-verify-like-ci-cannot-go-green-in-a-worktree.md
+++ b/.agents/backlog/HARNESS-058-verify-like-ci-cannot-go-green-in-a-worktree.md
@@ -77,11 +77,20 @@ skip — the gate still blocks.
 
 ### Where it is enforced
 
-| Entry point                             | Requires                   | Behaviour when unmet                                                      |
-| --------------------------------------- | -------------------------- | ------------------------------------------------------------------------- |
-| `pnpm harness:pre-push`                 | `install` + `build-output` | blocks the push with the naming message, before any check runs            |
-| `pnpm harness:verify-like-ci`           | `install`                  | refuses to run any stage; `build-output` is left to its own `build` stage |
-| any `verify-like-ci` build-output stage | `build-output`             | that stage FAILS naming the prerequisite, instead of a downstream error   |
+| Entry point                                      | Requires                   | Behaviour when unmet                                                      |
+| ------------------------------------------------ | -------------------------- | ------------------------------------------------------------------------- |
+| `pnpm harness:pre-push`, **when it will verify** | `install` + `build-output` | blocks the push with the naming message, before any check runs            |
+| `pnpm harness:verify-like-ci`                    | `install`                  | refuses to run any stage; `build-output` is left to its own `build` stage |
+| any `verify-like-ci` build-output stage          | `build-output`             | that stage FAILS naming the prerequisite, instead of a downstream error   |
+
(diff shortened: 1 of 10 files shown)


woojubb:
### Two follow-up commits, both found by this PR's own gates

**`review-gate` blocked the first push** — a genuine CodeQL finding in the new module, not a false
alarm:

```
review-gate: BLOCK (blocking-findings)
  - js/file-system-race [warning/security:high] scripts/harness/tree-prerequisites.mjs:110
```

`describeTree` decided whether a tree was a worktree by inspecting `.git`: `existsSync` → `statSync`
→ `readFileSync`, three chances for the path to change underneath. The first fix collapsed the check
to one `statSync({ throwIfNoEntry: false })`; CodeQL then flagged the surviving stat-then-read pair —
correctly, it is the same check-then-use race one syscall later.

The second fix stops asking the filesystem. **A linked worktree is precisely a tree whose
`--git-dir` differs from its `--git-common-dir`**, and one `git rev-parse` reads both without
touching a path it then reuses. That is not just race-free, it is authoritative where the `.git`
inspection was a heuristic.

The three `describeTree` tests now build real repositories and real `git worktree add` trees — a
fixture faking the `.git` layout would test a heuristic the code no longer uses — plus a fourth case
pinning the outside-any-repository answer.

Worth stating because it is the same shape this PR is about: `review-gate` is one of the three
contexts `verify-like-ci` **declares it cannot mirror locally**, with the reason printed on every
run. It behaved exactly as declared — the local gate was green, the un-mirrorable check was not, and
the run said in advance which ones it could not answer for.

All checks green at `5baef74a4`: 2126/2126 harness tests, 84/84 scans, `quality`, `scans`, `build`,
`tui-e2e`, `examples-typecheck`, `windows-shell`, `review-gate`, CodeQL. Every push went through the
pre-push gate with **no `--no-verify`**.

woojubb:
## The `willBuild` disagreement, settled by measurement

**Reviewer B is right**, and the evidence is stronger than B's own argument. Method: a fresh nested
worktree, `pnpm install` only, a real build regression (a syntax error in `packages/dag-core/src/index.ts`
— dag-core is what `agent-command-workflows` imports), then `verify-like-ci`.

### Before — what the stages after a FAILED `pnpm build` actually printed

`build` failed as it should:

```
packages/dag-core build:js: Failed
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @robota-sdk/dag-core@3.0.0-beta.60 build:js: `tsdown --no-dts`
 ELIFECYCLE  Command failed with exit code 1.
```

Then every consumer ran anyway. `examples-typecheck`:

```
error TS2307: Cannot find module '@robota-sdk/agent-framework' or its corresponding type declarations.
error TS2307: Cannot find module '@robota-sdk/agent-provider-anthropic' or its corresponding type declarations.
error TS2591: Cannot find name 'process'. Do you need to install type definitions for node?
             Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.
error TS7006: Parameter 't' implicitly has an 'any' type.
```

`binary-e2e`:

```
FAIL  src/__tests__/e2e/serve-mode.bintest.ts > robota --serve black-box runtime host (RUNTIME-001)
Error: serve host did not come up within 20000ms
Duration  20.47s
```

Twenty lines of type errors in `examples/`, none of them about the defect, plus an actively wrong
instruction to install `@types/node`, plus 20s spent waiting for a binary nobody built. That is the
exact noise this PR exists to remove, in the fresh-worktree scenario the feature is for.

Two things the measurement showed that neither review predicted:

- **`typecheck` was NOT the stage that misled.** It reported the real defect —
  `src/index.ts(74,33): error TS1109: Expression expected.` — because this particular regression is a
  syntax error tsgo also sees. B named `typecheck`; the actual noise came from `examples-typecheck`
  and `binary-e2e`. The mechanism is B's, the blast radius is different.
- **`scan-suite` was clean** — `BLOCKED: 70 package(s) have no dist/ … Run: pnpm build` — because it
  re-checks dist at run time. That asymmetry, which B pointed at, is exactly what the measurement
  confirms: the stages with no re-check are the ones that misled.

So Reviewer A's premise ("mildly redundant with the already-reported `build` failure") does not hold:
what followed was not a redundant restatement of the build failure, it was a different and wrong
diagnosis.

### After — same regression, same worktree, fixed code

```
===== typecheck =====
verify-like-ci stage `typecheck` did NOT run: the `build` stage FAILED earlier in this run.
This is NOT a second verdict — the `build` failure reported above IS the verdict. The build
output this stage reads was never produced, so anything it reported would describe the
missing output rather than your change.

  MISSING  build-output  70 of 75 buildable package(s) have no dist/.
  Fix the `build` failure reported above, then re-run the gate.
```

```
✗ build — the plan needs build output (ci.yml → build builds too)
✗ typecheck — `build` failed in this run — this stage measured nothing
✗ scan-suite — `build` failed in this run — this stage measured nothing
✗ affected-verify — `build` failed in this run — this stage measured nothing
✗ binary-e2e — `build` failed in this run — this stage measured nothing
✗ examples-typecheck — `build` failed in this run — this stage measured nothing
✗ tui-e2e — `build` failed in this run — this stage measured nothing
FAIL — 10 of 11 stage(s) failed … EXIT=1
```

Still red, still every stage accounted for. Note the message is a **second** cause, not the same one:
telling someone to run `pnpm build` right after they watched it fail is the wrong instruction, and
"nothing has been measured" is untrue — the build measured the change and rejected it.

### The fix

`willBuild` (a flag computed once) becomes state the loop carries: `initialBuildState` /
`advanceBuildState` **re-read dist from disk** after `build` runs, and `stageBlockCause` returns
`unprepared` (nothing will build it) or `build-failed` (it already failed above). The stage gate is
now asked *before* the block, so a stage CI would skip is skipped rather than reported as missing a
prerequisite it never needed.

### The test that pins it

`stageBlockCause` × 6, `advanceBuildState` × 3, and a `the loop sequence a run actually walks` block
that replays `[build, typecheck]` with a failing and a succeeding build. Red-proved by reverting the
two source files and keeping the tests: **15 fail**, four of them assertion-grade on the message text
rather than merely on the missing export:

```
AssertionError: expected '…' to contain 'the `build` stage FAILED earlier in this run'
AssertionError: expected '…' to contain 'Fix the `build` failure reported above'
AssertionError: expected '…' not to contain 'nothing about the diff has been measured yet'
AssertionError: expected [Function] to throw an error
```

### `scan-suite`'s own dist check — asked and answered, not removed

Redundant in this path now: the loop gate blocks `scan-suite` before `runScanSuite` reads dist, so
its inline check can only fire if dist disappears between the two reads. **Kept deliberately** — it
guards a *different* hazard (the dist-dependent scans silently no-op and LOOK like a pass: a
fail-open, not a misattribution) and it is the only protection if `runScanSuite` is ever called from
outside this loop. Removing it is a separate change. Recorded in the backlog doc.

### The tradeoff, stated

After a failed build, `typecheck` no longer prints its own diagnostic. Nothing is lost that the build
failure did not already report, and the stage runs normally once the build is fixed — the same
ordering CI enforces with `needs: build`.

### Housekeeping

Rebased onto `origin/develop` (#1575, #1579 picked up). The docs-only fresh-worktree green proof was
**re-run on the new loop** — `PASS — all 11 stage(s) passed`, EXIT=0. 2151/2151 harness tests, 84/84
scans. `record-local-review --findings 0` before the push; pushed through the gate, no `--no-verify`.

woojubb:
## The MUST: measured in a fresh worktree, fixed, and the ordering pinned

The reviewer is right, and the refusal reproduces exactly as described.

### Red — a fresh, uninstalled worktree at this branch

**(a) delete-only push** (`(delete) 000…0 refs/heads/some-branch <sha>`):

```
the pre-push gate did NOT run: this tree is missing a verification prerequisite.
  MISSING  install       no node_modules/.modules.yaml in THIS tree.
  MISSING  build-output  75 of 75 buildable package(s) have no dist/.
  Run this IN THIS TREE, then re-run the gate:
    pnpm install --frozen-lockfile
    pnpm build
RC=1
```

**(b) no-delta re-push** (`HARNESS_BASE_REF` = the branch's own tip, clean tree): byte-identical
refusal, `RC=1`.

So `git push --delete` of a remote branch, and re-pushing a branch identical to its base, both
demanded a full install and build for a push with **nothing to verify** — in the
parallel-subagent-in-a-fresh-worktree configuration this item exists to serve.

### Green — same worktree, still uninstalled (`isInstalled=false`, `missing=["install","build-output"]`)

```
(a)  ▶ scoped pre-push verification skipped: delete-only push                RC=0
(b)  ▶ scoped pre-push verification skipped: no content delta from 7a3fde170  RC=0
```

### The counter-check — moving a gate later is how a gate quietly stops firing

Same unprepared worktree, same command, but a push that genuinely needs verification
(`HARNESS_BASE_REF=origin/develop`, real delta):

```
the pre-push gate did NOT run: this tree is missing a verification prerequisite.
  MISSING  install       no node_modules/.modules.yaml in THIS tree.
  MISSING  build-output  75 of 75 buildable package(s) have no dist/.
RC=1
```

Still refused. The gate moved; it did not weaken.

### The test pins the SEQUENCE, not the two cases

`runPrePushGate(steps)` now states the step order once, with the steps injected, and
`__tests__/pre-push-sequence.test.mjs` asserts the recorded order — including
`indexOf('tree-prerequisites') > indexOf('decide-verification')` as an invariant rather than an
index, so it keeps its meaning if steps are added around it. A second block feeds **real** hook stdin
through `parsePrePushUpdates` → `decidePrePushVerification` → `runPrePushGate`, so the ordering test
cannot pin a decision shape nothing ever returns.

Red-proved by restoring the old order *inside the new structure* — isolating the ordering defect from
the refactor. 6 of 9 fail, assertion-grade on the order itself:

```
AssertionError: expected 2 to be greater than 4
AssertionError: expected [ 'prune-worktrees', …(5) ] to not include 'tree-prerequisites'
```

and the three that keep passing include **"a real push WITH a delta still reaches the prerequisite
assertion"** — the suite distinguishes *moving* the gate from *removing* it, which is the failure mode
you flagged.

### Also

Contract doc corrected: the table claimed pre-push requires install + build-output unconditionally.
It does not — a prerequisite is owed only by work that is going to happen.

Rebased onto `origin/develop` (#1578, #1582, #1583) on top of your `findMissingDist` commit. 2234/2234
harness tests, 84/84 scans. `record-local-review --findings 0`; pushed through the gate, no
`--no-verify`. Measurement worktrees removed and pruned.

PR #1576, opened 2026-07-31
fix(harness): PROC-007 a merge decision about a PR lives on the PR

woojubb:
Closes #1562.

## The defect

`re-plan` withdraws a change rather than patching it (`finding-depth.md`). #1557 stored that
decision in `.agents/local-reviews/<branch>.json` — **gitignored, per-working-tree**, keyed by the
local checkout's branch and HEAD — while the merge it was meant to stop is keyed by a **PR number**.
`worktree-parallel-orchestration` §5: *"The orchestrator — never the implementer — merges."* The
recording checkout is therefore not the merging one, by construction, and the record cannot follow
because the file is untracked.

INFRA-048 and INFRA-057 already established and already paid for the general form: a merge decision
held anywhere other than the PR does not stop a merge. Their answer — `review-gate.yml`, a required
check keyed by PR number with a label override and a `disarm-auto-merge` job — is where this now
lives.

## Red proof: the decision does not travel

Two checkouts of the same repository. `clone-a` records the round for `feat/withdrawn-change`;
`clone-b` is where an orchestrator would run the merge. The query is #1557's `mergeBlockReason`
verbatim — absent record ⇒ `null` ⇒ do not block.

```
clone-a is on: feat/withdrawn-change
clone-b is on: feat/other-branch

--- the recording checkout (clone-a) records the round ---
recorded: feat/withdrawn-change @ 5cabd4fec — 0 findings

--- what each checkout holds on disk ---
clone-a/.agents/local-reviews:  feat%2Fwithdrawn-change.json
clone-b/.agents/local-reviews:  (no such directory)

clone-a record now holds: {"branch":"feat/withdrawn-change","disposition":"re-plan"}

=== the gate's query, run IN the recording checkout (clone-a) ===
branch key:    feat/withdrawn-change
verdict:       BLOCK — the local round withdrew this change

=== the same query, run from the ORCHESTRATOR's checkout (clone-b) ===
branch key:    feat/withdrawn-change            <- headRefName resolved from gh, as review suggested
verdict:       null  => DO NOT BLOCK

=== and with clone-b keyed on its OWN branch, which is what the hook actually did ===
branch key:    feat/other-branch
verdict:       null  => DO NOT BLOCK
```

The middle case is the one that matters: resolving `headRefName` from `gh pr view` fixes the
branch-name half and still does not block, because there is nothing in that checkout to read.

The same defect stated as tests, run against unmodified `merge-gate.sh`:

```
[...]

PR #1578, opened 2026-07-31
fix(hooks): INFRA-081 one reader, one answer, whatever the host has installed

### LINKED ISSUE #1574: INFRA-081: hook_json_string's two arms give opposite verdicts on identical input
# INFRA-081 — `hook_json_string`'s two arms give opposite verdicts on identical input

Filed from a stated limit in #1566, then measured end-to-end rather than left as a reading.

## Problem

`hook_json_string` in `.claude/hooks/lib/command-scan.sh` has two implementations of one question,
picked by which tool the host happens to have installed:

```sh
if command -v jq  >/dev/null 2>&1; then … jq -r ".${path} // \"\"" … fi
if command -v python3 >/dev/null 2>&1; then … node if isinstance(node, str) else "" … fi
```

The python3 arm returns a value only when it is a **string**. The jq arm returns whatever is there,
pretty-printed. They agree on the shape that actually arrives and disagree on every other:

```
AGREE     command as a string (real shape)   -> git push
DISAGREE  command as an OBJECT               jq={\n  "a": 1\n}   python3=''
DISAGREE  command as a NUMBER                jq=123              python3=''
DISAGREE  command as an ARRAY                jq=[\n "git",…\n]   python3=''
AGREE     tool_name as a string              -> Bash
AGREE     cwd absent                         -> ''
```

## It reaches a verdict, and the verdicts are opposite

This is the part that makes it worth filing rather than noting. `branch-guard.sh` run end-to-end on a
scratch repository on `main`, same payload both times, only `PATH` differing:

```
payload: {"tool_name":"Bash","cwd":"<scratch>","tool_input":{"command":{"a":"git push origin main"}}}

with jq      exit=0     (allowed)
without jq   exit=2     (refused — "the tool command could not be decoded")
```

One host permits and the other refuses. Which one you get depends on a tool being installed, and
neither answer is reached by a decision anyone made.

**Direction of the divergence, stated because it is not the obvious one:** the python3 arm is the
one that fails CLOSED. The jq arm hands the guard a pretty-printed JSON blob, whose `git push` sits
inside quotes, so the tokenizer correctly masks it as data and the guard finds no verb — a correct
reading of an incorrect input.

## Not a demonstrated live bypass

The harness sends `tool_input.command` as a string, so no payload in normal operation takes either
divergent path. This is filed as a **divergence**, not an exploit. The reason it still matters:

- The whole point of INFRA-077 (#1551) was that copies of one fact gave different answers. This is
  the same defect **inside a single function**, and it survived that PR because #1566 could not touch
  `command-scan.sh` — #1565 was rewriting it concurrently.
- It is reachable from `hook_command_of`, `hook_tool_name_of` and `hook_cwd_of`, which is every
  guard's first read of what it is judging.
- #1566 fixed the same disagreement for the TEXT fields it owned by routing them through a new
  string-only `hook_json_text` (`prompt`, `file_path`, `session_id`, `transcript_path`), after
  measuring that `jq -r` prints a pretty-printed object where python3 writes `""` for the ordinary
  `{"message": {"role": "user", "content": "…"}}` transcript shape. The remaining three readers are
  the ones it could not reach.

## Direction

One rule, not one per available tool. Either make both arms string-only — matching `hook_json_text`,
which is the answer #1566 already chose for the fields it owned — or make the reader **refuse** a
non-string where a string was asked for, which is the fail-closed answer and would give both arms the
same verdict.

What must not survive is a reader whose answer depends on the host's installed tools, in the function
every guard uses to find out what it is judging.

## Related

- #1551 (INFRA-077) — the item this is the last unreachable corner of
- #1572 — `hook_executable_part` is a second reading of the same command; same shape, one layer up

woojubb:
## The defect, measured end to end

`hook_json_string` had two implementations of one question, picked by which tool the host happens to
have. They agreed on the shape that actually arrives and disagreed on every other — and the
disagreement reached a **verdict**. `branch-guard.sh` on a scratch repository, same payload both
times, only `PATH` differing:

```
payload: {"tool_name":"Bash","cwd":"<scratch>","tool_input":{"command":{"a":"git push origin main"}}}

--- with jq on PATH ---
exit=0

--- WITHOUT jq on PATH (python3 present) ---
[branch-guard] Blocked: the tool command could not be decoded, so it cannot be judged.
exit=2

--- direct field readings ---
jq-arm       hook_json_string tool_input.command -> [{
  "a": "git push origin main"
}]
python3-arm  hook_json_string tool_input.command -> []
```

One host permitted and the other refused, and neither answer was reached by a decision.

## The decision

**Both arms are string-only. A field that is not a string is not that field.**

The issue offered two answers and asked for a choice rather than a match. This is the first one, and
the reasoning is not "it matches `hook_json_text`":

1. It is the rule `hook_json_text` already states, **with its own measurement**, for the text fields
   #1566 could reach. Two spellings of one rule is what INFRA-077 spent a PR removing.
2. `hook-facts.sh` states the reader/caller split explicitly — *a READER collapses absent and empty
   into `""`, and the CALLER names what empty means for it*. A refusing reader would break that
   written, tested rule for one function out of the set.
3. **The verdict is fail-closed either way, and that is the property that matters.** Every caller
   already turns empty into a refusal or a fallback — `hook_command_of` returns non-zero on an empty
   command, `hook_tool_name_of` falls through to its no-tools parse — so the refusing-reader
   alternative would have produced the *same observable behaviour at every present call site*. It was
   measured, not assumed. Both hosts now exit 2 on the payload above.

## What else was closed with it

Two divergences in the same arm that the report did not name, both measured:

- `jq -r ".a.b"` **errors** when `a` is not an object, so a jq-only host answered "could not decode"
  where a python3 host answered `""`.
- `jq -r` appends a trailing **newline** the python3 arm does not write. Invisible to every caller
  (they all read through a command substitution) — and invisible is how a divergence survives to
  matter later. Both arms are now byte-identical, `-j` and all.

`hook_edit_content_of` next door carried four more of the same class, each reachable from an ordinary
payload, and is held to the same rule rather than left as the sibling that was not fixed.

`hook_json_text` is now `hook_json_string`. It existed only because #1566 could not reach the
function it needed; its body is gone and its **name** stays, because four hooks call it and their
reason for calling it is unchanged.

**`lib/hook-facts.sh` is touched, and that is why**: leaving a second implementation of the rule this
PR gives a single owner would be the defect wearing the fix as a costume.

## Measurements

| | before | after |
| --- | --- | --- |
| payload shapes whose answer depends on the installed tool | **10 of 17** | **0 of 17** |
| `branch-guard` verdict on the object-command payload | 0 with jq / 2 without | 2 / 2 |

**Red-first.** With `command-scan.sh` reverse-applied, 13 of the 23 new cases fail — including the
end-to-end verdict case — and the two control cases (bare `git push origin main` on `main` refused,
an ordinary commit on a feature branch permitted) stay green.

**False-positive sweep against `origin/develop`.** 625 real command lines taken from this tree (every
fenced `git`/`gh`/`pnpm`/`npx`/`node` line in 1782 documents under `.agents/` and `docs/`, plus the
root `package.json` scripts), each run through all three Bash guards end to end under both the
baseline hooks and these:

```
0 verdict(s) moved out of 1875 runs.
NEWLY SEEN (the baseline allowed it, this refuses it): 0
NEWLY BLIND (the baseline refused it, this allows it): 0
```

Closes #1574

### DIFF
--- a/.claude/hooks/check-forbidden-patterns.sh
+++ b/.claude/hooks/check-forbidden-patterns.sh
@@ -101,10 +101,12 @@ if [[ "$RELATIVE_PATH" == /* ]]; then
     */apps/*) RELATIVE_PATH="apps/${FILE_PATH#*/apps/}" ;;
   esac
 fi
-# Through `hook_json_text`, so this reads the same on a host with jq and one without: measured,
-# `hook_json_string` hands back a non-string node's JSON where jq is installed and "" where it is
-# not. A session id and a transcript path are text or they are absent. See lib/hook-facts.sh.
-SESSION_ID=$(hook_json_text "$INPUT" 'session_id' || true)
+# A session id and a transcript path are TEXT, or they are absent — there is no third answer, and
+# `hook_json_string` is the single owner of that rule: a field that is not a JSON string reads as "",
+# on a host with jq and on a host without, byte for byte (INFRA-081, #1574). This used to call
+# `hook_json_text`, which existed only because the rule was true in one file and not in the other;
+# once it was true in both, that name was an alias and is gone. See lib/command-scan.sh.
+SESSION_ID=$(hook_json_string "$INPUT" 'session_id' || true)
 BLOCKED=false
 BLOCK_MESSAGES=""
 
--- a/.claude/hooks/correction-detect.sh
+++ b/.claude/hooks/correction-detect.sh
@@ -24,10 +24,12 @@ fi
 # LESSON-010: only REAL user turns are correction signals. Subagent/eval prompts (session ids
 # like "agent_1") and events with no session id are agent-authored text — counting them
 # inflated the one genuinely useful metric with false positives.
-# Through `hook_json_text`, so this reads the same on a host with jq and one without: measured,
-# `hook_json_string` hands back a non-string node's JSON where jq is installed and "" where it is
-# not. A session id and a transcript path are text or they are absent. See lib/hook-facts.sh.
-SESSION_ID=$(hook_json_text "$INPUT" 'session_id' || printf '')
+# A session id and a transcript path are TEXT, or they are absent — there is no third answer, and
+# `hook_json_string` is the single owner of that rule: a field that is not a JSON string reads as "",
+# on a host with jq and on a host without, byte for byte (INFRA-081, #1574). This used to call
+# `hook_json_text`, which existed only because the rule was true in one file and not in the other;
+# once it was true in both, that name was an alias and is gone. See lib/command-scan.sh.
+SESSION_ID=$(hook_json_string "$INPUT" 'session_id' || printf '')
 case "$SESSION_ID" in
   '' | agent*) exit 0 ;;
 esac
@@ -48,7 +50,7 @@ if [ -z "$KEYWORD" ]; then
   exit 0
 fi
 
-TRANSCRIPT_PATH=$(hook_json_text "$INPUT" 'transcript_path' || printf '')
+TRANSCRIPT_PATH=$(hook_json_string "$INPUT" 'transcript_path' || printf '')
 TRANSCRIPT_PATH="${TRANSCRIPT_PATH/#\~/$HOME}"
 PREVIOUS_ASSISTANT_HASH=""
 
--- a/.claude/hooks/eval-log-stop.sh
+++ b/.claude/hooks/eval-log-stop.sh
@@ -31,10 +31,12 @@ git_project() {
 BRANCH=$(hook_current_branch "$PROJECT_DIR" "unknown")
 # It carried no `read_json()`, so it looked unlike its three siblings, and it read the payload with
 # a bare `jq -r` and wrote its record with `jq -cn` all the same. Same defect, different spelling.
-# Through `hook_json_text`, so this reads the same on a host with jq and one without: measured,
-# `hook_json_string` hands back a non-string node's JSON where jq is installed and "" where it is
-# not. A session id and a transcript path are text or they are absent. See lib/hook-facts.sh.
-SESSION_ID=$(hook_json_text "$INPUT" 'session_id' || printf '')
+# A session id and a transcript path are TEXT, or they are absent — there is no third answer, and
+# `hook_json_string` is the single owner of that rule: a field that is not a JSON string reads as "",
+# on a host with jq and on a host without, byte for byte (INFRA-081, #1574). This used to call
+# `hook_json_text`, which existed only because the rule was true in one file and not in the other;
+# once it was true in both, that name was an alias and is gone. See lib/command-scan.sh.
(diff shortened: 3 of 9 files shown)


woojubb:
Addressed, and the finding was right — a stale rationale is a stale implementation one layer up.

## The decision: `hook_json_text` is removed, not reworded

The four comments justified the call with the defect this PR removes. Rewording them would have left
the question they were answering — *why is there a second reader?* — with no answer at all, because
once both arms agree the second reader was a **one-line delegate with an identical contract**. That
is an alias, and an alias is a second **name** for one fact: exactly the class this workstream is
removing, one level up from a second implementation. A reader meeting two names has to ask which one
to use, and there is no answer.

So the name went with the body. Its six callers — `hook_file_path_of`, `hook_prompt_of`, and the four
hooks — ask `hook_json_string` directly, and the reason those callers wanted the type test now lives
where the rule does:

```sh
# A session id and a transcript path are TEXT, or they are absent — there is no third answer, and
# `hook_json_string` is the single owner of that rule: a field that is not a JSON string reads as "",
# on a host with jq and on a host without, byte for byte (INFRA-081, #1574). This used to call
# `hook_json_text`, which existed only because the rule was true in one file and not in the other;
# once it was true in both, that name was an alias and is gone. See lib/command-scan.sh.
```

**Held mechanically, not described.** "There is one owner" is a claim a future convenience wrapper
falsifies in one line, so a new case enumerates every hook and lib source and fails if
`hook_json_text` reappears. If a reader legitimately needs to come back it must come back with a
*different* contract, and that case has to be rewritten to say what it is.

## The fifth

The review found four by reading the diff's neighbours; grepping the whole tree for statements about
the old behaviour of any reader this PR changed found one more, in the present tense:

- `scripts/harness/__tests__/hook-facts.test.mjs:302` — *"`hook_json_string`'s two arms DISAGREE on a
  field that is not a string"*. Corrected to past tense with the item that closed it; the case itself
  stays, because the property it pins is the reader's contract rather than the history of one defect.

Two other matches were checked and deliberately left:

- `.agents/backlog/INFRA-077-*.md` (2 lines) — that item's own problem statement in past tense, not a
  rationale for a live call. Also outside this workstream's file ownership.
- `hook_prompt_of`'s note that *"`hook_json_string` CANNOT express jq's `//` distinction"* — verified
  still **true**: the reader returns `""` with exit 0 for an absent key and for a present-but-empty
  one alike, on both arms. Left as it stands.

## Verification

- `bash -n` on every edited shell file.
- Full suite: **128 files, 2155 tests, green.** All **84 scans** pass.
- The four re-pointed calls are *exercised*, not inspected: `hook-facts.test.mjs` already runs
  `correction-detect` and `eval-log-stop` end to end on a host without jq and asserts the records they
  write.
- Rebased onto `origin/develop` at `800e3373e` (picks up #1575, #1576, #1579).

PR #1580, opened 2026-07-31
feat(harness): INFRA-072 a red proof must have reached the code the fix wrote

### LINKED ISSUE #1537: INFRA-072: a test that passes without reaching the behavior it names

woojubb:
Direction 1 landed in #1568. This closes the remaining half by **taking direction 3 (coverage delta
per case) and rejecting direction 2 (branch-level mutation)** — both with the measurement behind
them.

## Acceptance criterion, set before any code was written

1. Of the four motivating cases, the witness was predicted to catch **0 additional** (direction 1
   already catches #4 partly), with a per-case reason stated in advance and then replayed.
2. It must catch the class direction 1 cannot see: a case that FAILS on the reversed source and
   executed none of the lines the fix wrote.
3. **Zero** false alarms over recent real ranges.

Measured against all three below. The replay encoded the per-case predictions and printed "as
predicted".

## Red first

Before the fix, the gate was handed a case that is genuinely red on the reversed source while its red
came from outside the fix. It printed:

```
✅  packages/x/src/target.ts: red-proof-ok (assertion-fail)
```

— a good red, which is exactly the class the item names. The new test failed against that. It passes
now, reporting `red-proof-unreached`.

## What landed

Two instruments, one question — did the case that supplied the red actually execute the code this fix
wrote?

- **bash**: `BASH_ENV` names a prelude every non-interactive bash sources before the script it was
  asked to run, so a hook spawned several processes down inside a vitest worker is instrumented
  without touching the test or the hook. `BASH_XTRACEFD` sends the trace to its own descriptor, so a
  test asserting on the hook's stderr is unaffected by being measured.
- **`.mjs`/`.ts`**: vitest v8 coverage, whose istanbul-shaped output names the executed statements
  and which lines are statements at all.

Per-case attribution comes from re-running the single deciding case with `-t`, because vitest
attributes outcomes to cases but coverage to a run.

Three answers, and only `UNREACHED` is a finding. A comment-only hunk, a pure deletion, a report that
never names the file — each is `UNKNOWN` and leaves the verdict exactly as it was.

## The first formulation was measured wrong, and that is why the second exists

It first asked the question of the REVERSED tree against the fix's OLD-side lines. Replayed on
`c08e0dbd6` it called a **genuine** red proof `unreached`: that fix is an ADDITION, so its old side
held nothing but a comment and one `case` pattern arm. Probing bash 5.2 also showed `set -x` never
names a bare `case` arm — `*.ts) echo is-ts ;;` traces because it carries a command, `*.md)` alone
never does and only its body traces.

Asked of the FIXED tree against the NEW-side lines, with bare `case` arms excluded from the
executable set, the same replay gives `reached`. That is what shipped.

## The four motivating cases — what this catches, and what it does not

| Case | Caught? | Why |
| --- | --- | --- |
| `hooks-have-execution-coverage` | **no** | the logic under test is IN the test file — no source to reverse, no pair, so the gate emits no verdict for a witness to qualify |
| `remaining-hooks-run` (two hooks) | **no** | the hooks named were not changed, so again no pair |
| `remaining-hooks-run` (extension) | **no** | REPLAYED on `b1f46acf3`: the fix wrote only comments and one bare `case` arm, so no changed line is traceable and the witness answers UNKNOWN. And the behaviour the case names — the extension filter — is not in the range's diff at all, so no diff-scoped instrument can target it |
| `remaining-hooks-run` (unset var) | **partly** | unchanged from direction 1 |

**0 of 4 additional, as predicted.** The three the item still names are not reachable from inside this
gate: two produce no verdict at all, and the third names a behaviour the diff does not contain.
Closing those needs a per-case reachability check over the WHOLE suite rather than over a fix range —
a different gate, and the honest next item.

## Why direction 2 was rejected

- Per-hunk reversal — the tractable reading of "negate the branch a case names" — catches none of the
  four either. #3's fix is entirely within the crash region so every hunk kills the case; #4's vacuous
  case hides behind a genuine sibling that kills every hunk.
- The reading that WOULD catch #4 (every added case must be killed by some mutant) is the same shape
  as "every added case must fail on reversal", which #1568 measured as failing correct work.
- Cost: one vitest run per hunk, against one per deciding case here. Over the 31 recent ranges this
  gate judges, single source files carry up to 617 changed lines.

Strictly more expensive, and it catches nothing direction 3 does not.

## False positives — measured, not asserted

Eight recent merged ranges replayed through the real gate (worktree detached at each merge, this
branch's gate copied in), 15 sources judged, **12 produced a red proof**:

| Witness | Count | Effect |
| --- | --- | --- |
| REACHED | 11 of 12 | verdict unchanged |
| UNKNOWN | 1 of 12 | verdict unchanged |
| UNREACHED | **0 of 12** | **zero false alarms** |

Ranges: `63fa0c0cf`, `2ac10f251`, `631aa9e27`, `02f5a84b9`, `a668df9e3`, `2b0c454e0`, `4719efe5a`,
`8589d58fa`. Both instruments are exercised there — nine `.sh` sources through the tracer, two `.mjs`
sources through coverage.

## Not promoted

The gate stays ADVISORY and `red-proof-unreached` is report-only even under
`REGRESSION_RED_PROOF_ENFORCE`. Promotion is INFRA-046's decision, not this one.

## Verification

- `npx vitest run scripts/harness/__tests__/` — 2117 tests green
- `pnpm harness:scan` — all 84 scans pass

Closes #1537

### DIFF
--- a/.agents/backlog/completed/INFRA-072-a-test-that-does-not-reach-what-it-names.md
+++ b/.agents/backlog/completed/INFRA-072-a-test-that-does-not-reach-what-it-names.md
@@ -91,6 +91,97 @@ So direction 1 closes the masking half and nothing else, which is what the item 
 remaining half — a case that fails for the wrong reason, or passes beside a genuine sibling — needs
 the branch-level or coverage-delta work, and stays open as a separate decision.
 
+## Decision (2026-08-01, second pass) — direction 3, and direction 2 rejected with the measurement
+
+Directions 2 and 3 were the remaining halves. **Direction 3 (coverage delta per case) is
+implemented. Direction 2 (branch-level mutation) is rejected**, and the reason is measurable rather
+than a matter of taste.
+
+### Acceptance criterion, set before any code was written
+
+1. Of the four motivating cases, the witness was predicted to catch **0 additional** — direction 1
+   already catches #4 partly — with a per-case reason stated in advance and then replayed.
+2. It must catch the class direction 1 cannot see: a case that FAILS on the reversed source and
+   executed none of the lines the fix wrote.
+3. **Zero** false alarms over recent real ranges. A guard that fires on correct work gets switched
+   off, and #1568 already established that "every added case must fail" is that kind of guard.
+
+### What was built
+
+Two instruments, one question — did the case that supplied the red actually execute the code this
+fix wrote?
+
+- `bash`: `BASH_ENV` names a prelude every non-interactive bash sources before the script it was
+  asked to run, so a hook spawned from inside a vitest worker is instrumented without touching the
+  test or the hook. `BASH_XTRACEFD` sends the trace to its own descriptor, so a test asserting on the
+  hook's stderr is unaffected by being measured.
+- `.mjs`/`.ts`: vitest v8 coverage, whose istanbul-shaped output names the executed statements and
+  which lines are statements at all.
+
+Per-case attribution comes from re-running the single deciding case with `-t`, because vitest
+attributes outcomes to cases but coverage to a run.
+
+Three answers, and only UNREACHED is a finding. UNKNOWN — a comment-only hunk, a pure deletion, a
+report that never names the file — leaves the verdict exactly as it was.
+
+### Two formulations were tried, and the first was measured WRONG
+
+The first asked the question of the REVERSED tree against the fix's OLD-side lines. Replayed on
+`c08e0dbd6`, it called a **genuine** red proof `unreached`: that fix is an ADDITION, so its old side
+held nothing but a comment and one `case` pattern arm. It also turned out that `set -x` never names a
+bare `case` arm at all — probed on bash 5.2, `*.ts) echo is-ts ;;` traces at its own line because it
+carries a command, while `*.md)` alone never does and only its body traces.
+
+Asked of the FIXED tree against the fix's NEW-side lines, with bare `case` arms excluded from the
+executable set, the same replay gives `reached`. That is the formulation that shipped.
+
+### The four motivating cases, judged against what landed
+
+| Case                              | Caught?     | Why                                                                                                                                                                                                                    |
+| --------------------------------- | ----------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `hooks-have-execution-coverage`   | **no**      | the logic under test is IN the test file — no source to reverse, no pair, so the gate emits no verdict for the witness to qualify. Out of reach of any diff-scoped pair gate.                                           |
+| `remaining-hooks-run` (two hooks) | **no**      | the hooks named were not changed, so again no pair.                                                                                                                                                                    |
+| `remaining-hooks-run` (extension) | **no**      | REPLAYED on `b1f46acf3`: the fix wrote only comments and one bare `case` arm, so no changed line is traceable and the witness answers UNKNOWN. And the behaviour the case names — the extension filter — is not in the range's diff at all, so no diff-scoped instrument can target it. |
+| `remaining-hooks-run` (unset var) | **partly**  | unchanged from direction 1. The witness holds for it either way; catching it in the "a genuine added case fails beside it" configuration needs a demand on EVERY added case, which #1568 measured as failing correct work. |
+
+**0 of 4 additional, as predicted.** The three the item still names are not reachable from inside
+this gate: two produce no verdict at all, and the third names a behaviour the diff does not contain.
+Closing that would take a per-case reachability check over the WHOLE suite rather than over a fix
+range — a different gate, and the honest next item.
+
+### Why direction 2 was rejected
+
+Branch-level mutation was traced through the same four cases and the same false-positive test:
+
+- Per-hunk reversal (the tractable reading of "negate the branch a case names") catches none of the
+  four. #3's fix is entirely within the crash region, so every hunk kills the case; #4's vacuous case
+  hides behind a genuine sibling that kills every hunk.
+- The reading that WOULD catch #4 — every added case must be killed by some mutant — is the same
+  shape as "every added case must fail on reversal", which #1568 measured as failing correct work: a
+  range routinely adds cases that do not depend on the fix.
+- Cost: one vitest run per hunk. Over the 31 recent ranges this gate judges, source files carry up to
+  617 changed lines across many hunks.
+
+Strictly more expensive, and it catches nothing direction 3 does not.
+
+### Measured, on real ranges
+
+Eight recent merged ranges replayed through the real gate (worktree detached at each merge, this
+branch's gate copied in), 15 sources judged, **12 produced a red proof**:
+
+| Witness   | Count      | Effect                    |
+| --------- | ---------- | ------------------------- |
+| REACHED   | 11 of 12   | verdict unchanged         |
+| UNKNOWN   | 1 of 12    | verdict unchanged         |
+| UNREACHED | **0 of 12** | **zero false alarms**    |
+
+Ranges: `63fa0c0cf`, `2ac10f251`, `631aa9e27`, `02f5a84b9`, `a668df9e3`, `2b0c454e0`, `4719efe5a`,
+`8589d58fa`. Both instruments are exercised there — nine `.sh` sources through the tracer, two `.mjs`
+sources through coverage.
+
+The gate stays ADVISORY and `red-proof-unreached` is report-only even under
+`REGRESSION_RED_PROOF_ENFORCE`; promotion is INFRA-046's decision, not this one.
+
 ## GATE-COMPLETE (2026-08-01)
 
 - `addedCaseTitleMatchers` reads titles off ADDED diff lines only; a context line and a REMOVED line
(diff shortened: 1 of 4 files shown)


woojubb:
Both SHOULDs addressed, each red-proved against the current code before fixing. Rebased onto `origin/develop` (now includes #1582).

## Finding 1 — `BARE_CASE_ARM` over-matched

Reproduced red, and the false-alarm symptom is worse than "excluded a line": with the subshell as the **only** fix-written line the answer degrades to UNKNOWN, and with an unreached sibling line beside it the answer is a **finding against correct work**.

```
× untraceableShellLines > does NOT exclude a subshell in an arm BODY — `set -x` traces that line
  → expected true to be false
× witnessOneCase wiring > REACHED when the only fix-written line is a subshell inside a case arm body
  → expected 'unknown' to be 'reached'
× witnessOneCase wiring > does not report a FINDING when the subshell it reached sits beside an unreached line
  → a line the case DID run was excluded, so the fix read as unreached: expected 'unreached' to be 'reached'
```

Fixed by recognising an arm by **position** rather than shape alone — arm position is straight after `case … in` and after each `;;` / `;&` / `;;&`; everything between is a body, where a parenthesised line is a command. The state is a stack, so a `case` nested in an arm body does not reset its enclosing block. A shape-only regex cannot separate the two, because `(pattern)` really is valid arm syntax.

## Finding 2 — the cap, and what I chose

Reproduced red:

```
× witnessDecidingCases > finds a REACHED case that sits past the OLD cap of 3
  → expected 'unreached' to be 'reached'
× witnessDecidingCases > stops at the budget and says UNKNOWN, never UNREACHED
  → expected 'unreached' to be 'unknown'
```

**You were right that the absent input proved nothing — and it was not absent.** I instrumented the replay to count deciding failures per source. Across the same eight ranges: **1, 1, 4, 5, 1, 10, 1, 2, 19, 3, 3, 9**. The cap of 3 truncated the walk for **5 of 12** sources. It changed no verdict only because an early case reached the fix in each; a range whose reaching case sat 4th would have produced a false finding. My original comment claiming the maximum was 1 was written before I measured it — the measurement corrected me.

**Chosen: keep a budget, make it honest, and size it from the measurement.**
- Every deciding case is asked until one answers REACHED or the budget stops the walk.
- **A stopped walk answers UNKNOWN, never UNREACHED** — "we stopped looking" and "nothing reached it" are different answers, and only the second is grounds for a finding.
- Budget raised 3 → **25**, covering the observed maximum of 19 with headroom.

Why keep any bound: raising it is nearly free on the healthy path, since a REACHED short-circuits and a sound range pays for one run whatever the number is. Only a range heading for a finding walks the list, so the bound can only ever *weaken* a verdict, never manufacture one — which is the safe direction for the one thing an unbounded loop would buy.

## Corpus re-measured — and a correction to my own numbers

Unchanged after both fixes:

| | |
| --- | --- |
| sources judged | 17 |
| red proofs (`outcome=assertion-fail`) | 13 |
| REACHED | 12 |
| UNKNOWN | 1 |
| **UNREACHED** | **0** |

The other 4 rows never offered a proof (3 `outcome=null`, 1 `run-error`).

My earlier report said "15 sources, 12 red proofs, 11 REACHED". That was a **miscount of the same run** on my part, not a change in behaviour — the pairs are now counted mechanically off `outcome=… witness=…` rather than by eye. Conclusion unchanged, numbers corrected.

## Re-measured against the post-#1582 hooks

`command-scan.sh` was rewritten underneath the instrumentation, so it was re-run against the current tree — **5 of 5** hooks instrument cleanly:

```
PASS branch-guard.sh: traced 45 own line(s), 2 sourced lib file(s); status 0->0; stderr identical=true; lines classed untraceable yet traced=none
PASS pre-push-check.sh: traced 8 own line(s), 2 sourced lib file(s); status 0->0; stderr identical=true; lines classed untraceable yet traced=none
PASS check-forbidden-patterns.sh: traced 9 own line(s), 2 sourced lib file(s); status 0->0; stderr identical=true; lines classed untraceable yet traced=none
PASS worktree-cwd-guard.sh: traced 21 own line(s), 2 sourced lib file(s); status 0->0; stderr identical=true; lines classed untraceable yet traced=none
PASS post-tool-format.sh: traced 6 own line(s), 2 sourced lib file(s); status 0->0; stderr identical=true; lines classed untraceable yet traced=none
```

The last column is an independent cross-check of the tightened arm rule against real hooks rather than fixtures: no line classified untraceable was actually named by a trace.

## Verification

`npx vitest run scripts/harness/__tests__/` — 2181 tests green. `pnpm harness:scan` — all 84 pass. Pushed without `--no-verify`.

woojubb:
Fixed, red-proved against a real vitest run, and the anchoring given its own guard. Rebased onto `origin/develop` (now includes #1578, #1583).

## Red first — with a real vitest, not a faked `-t`

The defect **is** vitest's `-t` semantics, so an injected runner implementing my own idea of them would have been an accidental green. The test builds a temp project (`vitest run --root`) with two cases whose names are prefix/suffix of one another, only the longer touching the fix-written line, and the **shorter** as the deciding failure:

```
× witnessOneCase wiring > isolates the named case even when a SIBLING name contains it
  → a sibling case supplied the execution credited to this one: expected 'reached' to be 'unreached'
```

## What `-t` anchors against — measured, not read

You were right to flag this as the risk that could make the fix worse than the bug. Probed on vitest 3.2.6 against `describe('outer group')` + `it('parses config')`:

| Pattern | Result |
| --- | --- |
| `parses config` | **2 ran** — the collision |
| `^parses config$` | **0 ran, 2 skipped** — the bare title is *not* the subject |
| `^outer group parses config$` | **1 ran, 1 skipped** — correct |

So the subject is the describe chain + title joined by **single spaces**. I then confirmed the json reporter emits the identical string:

```
{"ancestorTitles":["outer group"],"title":"parses config","fullName":"outer group parses config"}
```

`decidingFailures` already hands over `fullName`, so the pattern is now `^<escaped fullName>$`. Where only a bare title is available (no `fullName`), the END is anchored and the prefix left open — still immune to the sibling collision, without the empty-match trap. `decidingFailures` now reports which of the two it had, so the choice is made on fact rather than assumption.

Incidentally this also caught a wrong fixture in my own tests: I had invented `'x > the new regression case'` as a `fullName`. The real separator is a space. Corrected.

## The guard you asked for — the 12 are still 12

Re-measured on the same eight ranges after anchoring:

| | |
| --- | --- |
| red proofs | 13 |
| REACHED | **12** |
| UNKNOWN | 1 |
| UNREACHED | 0 |

Byte-identical to the pre-anchoring run. Those 12 come from real `-t` invocations against real describe chains, so had the pattern stopped matching they would all have flipped to UNKNOWN — which is exactly the silent failure this check exists to exclude.

## On the finding's importance

Your framing was the useful part: my acceptance criterion measured **false alarms**, and this failure mode is a **missed detection**, so the corpus would have read clean while the feature quietly did not work. The two earlier findings were "a correct fix reported wrong"; this one was "a wrong fix reported correct". I have recorded that distinction in the item document rather than just the fix, since the blind spot was in how I chose to measure, not only in the code.

## Verification

`npx vitest run scripts/harness/__tests__/` — 2217 green. `pnpm harness:scan` — all 84 pass. `record-local-review.mjs --findings 0` before pushing; no `--no-verify`. An unrelated `weekly-digest.md` date-window regeneration that a hook produced during the run was restored rather than swept into the commit.

Issue #1587, opened 2026-08-01
INFRA-084: an argument passed to node is read as a shell command


woojubb:
# INFRA-084 — an argument passed to `node` is read as a shell command

Found while enforcing the `--no-verify` ban (INFRA-083). **Pre-existing**: measured identical on
`origin/develop` and on the branch, so nothing in that change caused it.

## Problem

`hook_verb_scan` masks a double-quoted argument after most commands and leaves it **visible** after
`node`:

```
git commit -m "note: --no-verify is banned"          -> git commit -m "···························"
echo "git push --no-verify"                          -> echo "····················"
node x.mjs --notes "INFRA-083: --no-verify banned"   -> node x.mjs --notes  INFRA-083: --no-verify banned
node x.mjs --notes "git push -n left alone"          -> node x.mjs --notes  git push -n left alone
```

(`·` is the mask byte.) The quotes are stripped and the content is handed to the guards as if it were
executable text.

## It reaches a verdict

`branch-guard.sh` on a scratch repo whose branch is `main`, same result on `develop` and on the
INFRA-083 branch:

```
exit=2  node scripts/x.mjs --notes "git push origin main was blocked"
          -> Blocked: cannot git push on protected branch 'main'
exit=2  node scripts/x.mjs --notes "we should git commit here"
          -> Blocked: cannot git commit on protected branch 'main'
exit=0  echo "git push origin main"          (correctly ignored)
```

Prose describing a git command, passed to a node script, is refused as if it were that command.

## Why it matters here specifically

This breaks **the repository's own workflow**. `scripts/harness/record-local-review.mjs --notes "…"`
is the mandated pre-push step, and a note that mentions a guarded command — which review notes
routinely do, since they describe what was fixed — is blocked. I hit it several times in one session
writing notes about `git push` and `git commit`.

The cost is the shape this repo keeps paying for: a guard that fires on correct work is one people
learn to route around, and the route around this one is to stop describing git in review notes, which
makes the notes worse.

## Why it is probably deliberate for the wrong reason

The tokenizer takes `HOOK_INTERPRETER_RE` and `HOOK_SHELL_INTERPRETER_RE`. Keeping a quoted argument
visible is **correct** for a shell interpreter — `bash -c "git push"` really does run that text, and a
guard that masked it would go blind on a genuine bypass. `node` appears to be classified with them.

But `node`'s executable argument is `-e` / `--eval`, and it is **JavaScript**, not shell. Every other
[...]

Issue #1584, opened 2026-07-31
INFRA-082: the worktree isolation the parallel skill promises does not cover the stash


woojubb:
# INFRA-082 — the worktree isolation the parallel skill promises does not cover the stash

Filed from a live incident on 2026-08-01, during a five-agent parallel wave.

## What happened

Two agents were working in separate worktrees of the same clone. One ran a stash-then-pop for a
before/after measurement and **popped the other agent's uncommitted work into its own tree.**

It noticed, re-stashed the work immediately with an explicit label, stopped using `git stash` for the
rest of its run, and reported it. The entry is still on the shared stack:

```
stash@{0}: On fix/infra-081-hook-json-string: RECOVERED from a cross-worktree stash-pop race
           (agent-a66da77c67550bc51 popped it by mistake): HARNESS-058 verify-like-ci work
```

Nothing was lost this time, because the victim had already committed and pushed. That is luck, not
design: the same sequence half a minute earlier takes uncommitted work out of a live agent's tree
into another agent's diff, where it lands in the wrong PR or is dropped as unrecognised.

## The rule already exists and was not reached

`.agents/rules/git-branch.md` § Stash hygiene:

> Never use a bare `git stash` / blind `git stash pop` … stashes accumulate across sessions and `pop`
> restores the wrong entry. … pop by explicit ref (`git stash pop stash@{N}`), never the bare top of
> the stack.

And `.agents/backlog/completed/LESSON-005-evals-churn-and-stash-hygiene.md` closed this class on
2026-06-15, after 38 accumulated stashes made a blind pop restore the wrong entry.

**So this is the repository's signature defect, on its own process:** a rule stated, registered in
the right document, and never mechanically reached. Ten weeks later an agent did the exact thing the
rule forbids.

## What is new, and worse

LESSON-005's failure mode was *your own past work, restored at the wrong time* — annoying, recoverable
by reading the diff. Under `worktree-parallel-orchestration` the same command takes **another agent's
live work**, and the victim has no way to notice: its files simply revert under it while it is mid-run.

The skill's own description promises *"worktree-isolated subagents … with zero merge conflicts"*, and
the isolation is real for the working tree and the index. **It is not real for the stash, which is a
single ref (`refs/stash`) on the shared object store.** Nothing in the skill or the rule says so —
an agent reading either would reasonably believe its worktree is its own.

## Direction

1. **A mechanical guard, because a rule alone has now failed twice.** Refuse a bare `git stash pop` /
[...]

PR #1588, opened 2026-08-01
fix(hooks): a gate cannot be skipped by asking git to skip it

woojubb:
## Problem

Four parallel agents bypassed the verification gate in a single day. I fixed the **cause** — it could
not go green in a worktree (HARNESS-058, #1571) — and then **told** the agents not to bypass.

It worked. It is not a mechanism.

A rule stated and never mechanically reached is this repository's signature defect, and it had already
happened with this exact shape: `git-branch.md` has said *"never a bare `git stash pop`"* since
LESSON-005 on 2026-06-15, and an agent did it anyway ten weeks later (#1584).

## Where the enforcement has to live, and why it is not a preference

The flag disables the **git-level** hook. So the pre-push hook cannot catch its own bypass — by the
time it would run, it has already been skipped.

A **PreToolUse** hook runs on the *tool call*, before git is invoked at all. That is the one layer the
flag cannot reach, and this repository already has five hooks there.

## Placement

Beside the `gh pr merge --delete-branch` ban, in the same per-statement loop, reading the same masked
statement. That is the existing precedent for a zero-exception flag ban in this file — and a second
file for one rule is precisely the defect the previous seven review rounds on #1585 were about.

## Zero exceptions, and no override token

An override for an override is simply the next bypass. **If a check is wrong, unrunnable, or fires on
correct work, the check is what changes** — that is the whole of HARNESS-058, and a gate that trains
people to route around it has already failed.

## The short-flag asymmetry is measured, not assumed

```
git commit -h  ->  -n, --no-verify
git push   -h  ->  -n, --[no-]dry-run
```

So the short form is read for **commit only**. Refusing it on `push` would refuse a harmless
rehearsal — the same letter meaning two different things in two subcommands. The cluster form is
matched because `git commit -nm "x"` bundles.

## Measured end to end, on a scratch repo

```
exit=2  git push --no-verify
exit=2  git commit --no-verify -m "x"
exit=2  git commit -n -m "x"
exit=2  git commit -nm "x"

exit=0  git push -n                                    (dry run)
exit=0  git push --dry-run
exit=0  git commit -am "x"
exit=0  git commit -m "note: --no-verify is banned"    (masked prose)
exit=0  git push origin feat/probe
```

The second-to-last row is why it reads the **masked** statement: a commit whose message discusses the
flag is prose, not a bypass.

Red-proved — both refusal cases fail on the previous source. 2297 harness tests green, 84 scans pass.

## A pre-existing defect found while testing this — filed, not fixed here

`hook_verb_scan` masks a quoted argument after most commands and leaves it **visible** after `node`,
so prose handed to a node script is read as a shell command:

```
exit=2  node scripts/x.mjs --notes "git push origin main was blocked"
          -> Blocked: cannot git push on protected branch 'main'
```

Measured **identical on `origin/develop`**, so this change did not cause it. It breaks the repo's own
`record-local-review.mjs --notes "…"` step whenever a review note describes a git command, which
review notes routinely do. Filed as **#1587** with the direction and a both-directions acceptance
criterion (`bash -c "git push"` must stay refused).

## What the review rounds changed

The first version of the flag reading did its own parsing with sed and grep, and each review round
found a different way one of those passes was wrong: `awk -v` unescaped a backslash into a vertical
tab, a blind splice-removal desynchronised the quoting and hid a live flag behind an unterminated
string, a greedy match anchored on a nested verb, an option skipper swallowed `-x` as a flag. Four
readings of a command, written beside the tokenizer that already models the shell grammar and is
checked against real bash on a ~200-shape corpus.

They are gone. `command-scan.sh` gained a `words` mode — the words the shell builds for one
statement, splices collapsed, quoted content hidden, substitution contents excluded — and the guard
reads that.

Two bypasses the rewrite itself opened, both found by review and both red-proved before the fix:

- An expansion swallowed the rest of the statement. The splitter counted substitution delimiters and
  applied the same counting to `${…}`/`$((…))`, which the masker returns with the CLOSING character
  masked — so `git commit ${EXTRA} --no-verify -m x` passed in silence.
- A git global option latched its value as the subcommand: `git --work-tree /x commit -n -m y` set
  the verb to `/x`, which silenced both the `-n` and the `HUSKY=0` checks.

The floor that bans a second reading listed three retired helpers by name and predicted, in a
comment, the shape a new one would come back in. It came back exactly that way and the floor did not
notice, because the names were new. The four this PR removes are named in it now, red-proved.


woojubb:
## Finding 1 (MUST, semicolon past the `.husky` whitelist) — **INVALID**, measured

The premise is that a `;` inside a command substitution does not open a new statement, so a position
beginning with a whitelisted verb carries an arbitrary destructive command behind it. Statement
ranges **do** split there. Measured on this branch:

```
echo "$(echo x; rm .husky/pre-push)"     ->  ranges: 1 14 | 16 21   (two statements)
echo "$(echo x && rm -rf .husky)"        ->  ranges: 1 15 | 18 16   (two statements)
echo "$(echo x | rm .husky/pre-push)"    ->  ranges: 1 15 | 17 21   (two statements)
echo "$(chmod -x .husky/pre-push)"       ->  ranges: 1 34           (one statement)
```

The second statement in each of the first three leads with `rm`, which is not on the whitelist, so
the guard refuses. Driven end to end through `branch-guard.sh` on a scratch repo, every one of the
shapes in the finding is already **BLOCKED**:

```
BLOCK  echo "$(echo x; rm .husky/pre-push)"
BLOCK  echo "$(cat .husky/pre-push; rm .husky/pre-push)"
BLOCK  echo "$(echo x && rm -rf .husky)"
BLOCK  echo "$(echo x | rm .husky/pre-push)"
BLOCK  echo "$(ls; mv .husky /tmp)"
```

### What was right about it

The finding cites a comment in the diff, and that comment is **mine and it is wrong**: it says "a
statement range does not split at a substitution". True only when the substitution contains no
separator — which is exactly the `chmod` case it was written about, and not the general property it
states. A reviewer reading it would conclude precisely what this finding concluded.

Two changes, pushed:

- The comment now states the property exactly, with both measurements beside it, and says what the
  substitution-excluding reading actually cannot see: a command that shares a statement with the one
  it is nested in.
- The property is **pinned** in `hook-reading-matches-bash.test.mjs`, so an assumption that is false
  today cannot quietly become true. Four cases, including the one-statement `chmod` counterpart.

No behaviour change — the guard's verdicts on all five shapes above are unchanged, because they were
already correct.

PR #1593, opened 2026-08-01
fix(agent-core): CORE-027 an abort is a signal, not a substring

woojubb:
First step of **CORE-027**, ranked #3 in the architecture audit (#1591) and observed independently by
three of the six auditors (foundation, runtime, DAG).

## The defect

Three call sites decided whether a failure was an abort by looking at its **prose**:

```ts
error.name === 'AbortError' || message.includes('aborted') || message.includes('abort')
```

and `execution-service.ts:233-247` returned `success: true, interrupted: true` when it said yes.

So a real provider failure whose text happened to contain those letters came back to the caller as a
**successfully interrupted run**:

- `the upstream aborted the stream after a policy check`
- `AbortController is not supported by this runtime`
- `HTTP 502: <html>request aborted by proxy</html>`
- `model refused: the user asked how to abort a deployment`

Nothing downstream — including the print-mode exit code — had any way to tell a cancelled turn from
a failed one.

## The fix

The authoritative signal is the one the caller already holds. An abort happens because **something
aborted**: the `AbortSignal` that was passed in says so, and the platform's abort errors carry
`name === 'AbortError'` — which is also what `AbortSignal.throwIfAborted()` raises. Neither is a
guess about wording.

One owner, `isAbortFailure(error, signal)`, replaces all three copies — including the one in
`agent-framework`'s interactive path, which the first version of this PR missed. That third copy is
the one users meet: `interactive-session-prompt.ts:121` calls it and builds an INTERRUPTED result
when it says yes, so a real failure was presented as the user's own cancellation. The Task warned
about it and the first attempt shipped without it; review caught the gap and it is closed here.

It also reads the **other** direction, which the substring test could not: an aborted signal now
classifies correctly even when the error text says something unrelated (`socket hang up`) — the
common shape when a provider notices a cancellation as a transport failure. The old test would have
called that a genuine failure and reported it as one, which is the same defect pointing the other
way.

## Verification

- **Red-proved:** with the substring classification restored, **6 of the 10** core assertions fail, and with the framework copy restored **3 of the 6** interactive assertions fail.
  The four failing-direction cases are real provider-failure wordings; each was previously reported
  as a successful interrupted run.
- 916 `agent-core`, 155 `agent-session` and 1311 `agent-framework` tests pass.
- `isAbortFailure` is a new public export of `agent-core` and is **documented in its SPEC** — the
  public-surface gate caught it, and a new export that nothing documents is how a surface grows
  without a decision.

## Deliberately not in this change

CORE-027 also covers the other half of the same contract: the failure is rendered into an assistant
message (`Request failed: ${errMsg}`) and later **reconstructed** as `new Error(<display string>)` in
`execution-service-helpers.ts`, so the original error object — its type, its `cause`, its status — is
lost by the time a caller sees it.

That is a result-shape change with a wider blast radius, and mixing it in here would make both halves
harder to review. It stays open in the Task, and the commit says so rather than leaving the
impression that CORE-027 is finished.

### DIFF
--- a/packages/agent-core/docs/SPEC.md
+++ b/packages/agent-core/docs/SPEC.md
@@ -259,6 +259,23 @@ sandbox (`agent-tools`), the CLI monitor asset server (`agent-cli`) and the stud
 | `isPathInside`     | function | Whether `candidate` is `root` itself or lies beneath it, decided on the CANONICAL form of both     |
 | `canonicalizePath` | function | Realpath-resolve a path, tolerating a not-yet-created tail so `Write`/`Edit` targets still resolve |
 
+### Abort Classification Public API (CORE-027)
+
+The SSOT for "was this failure an ABORT?" whenever the answer changes what the caller reports.
+Three call sites decided it by looking at the error's PROSE
+(`message.includes('aborted') || message.includes('abort')`), and the execution service returned
+`success: true, interrupted: true` when it said yes — so a real provider failure whose text happened
+to contain those letters was reported as a successfully interrupted run, with nothing downstream,
+including the print-mode exit code, able to tell. The authoritative signals are the `AbortSignal` the
+caller already holds and the error's own `name`, neither of which is a guess about wording.
+
+Exported because `agent-framework`'s interactive path needs the same decision: a fourth private copy
+is how the third one survived a fix to the first two.
+
+| Export           | Kind     | Description                                                                                     |
+| ---------------- | -------- | ----------------------------------------------------------------------------------------------- |
+| `isAbortFailure` | function | Whether a failure is an abort, decided from the caller's `AbortSignal` and the error's own name |
+
 ### Schema (CORE-015)
 
 | Export                                                                                                    | Kind     | Description                                                                                                               |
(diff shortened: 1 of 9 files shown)


woojubb:
On the MUST — the premise does not hold on this runtime, but the fix is worth making anyway, so it is made.

**Measured:**

```
node v22.14.0
DOMException instanceof Error : true
Error.prototype in chain      : true
name                          : AbortError
```

WebIDL has put `Error.prototype` in `DOMException`'s chain since 2021, so this is the modern-browser behaviour too. The interactive test case the review predicted would fail — *"still recognises a real abort — the DOMException the old code tested for"* — passed before this change and still passes; CI ran that file green.

**Fixed regardless**, for two reasons:

1. Depending on that inheritance buys nothing when the structural test costs one line, and `agent-core` ships a browser build where I would rather not rely on it.
2. There is a second reason nobody raised, and it is the one no `instanceof` spelling can cover — including the `instanceof DOMException` branch the old framework copy had. **`instanceof` fails cross-realm.** An abort thrown in a worker or a `vm` context is an ordinary error that this realm's `Error` does not recognise.

`isErrorShaped` is now structural: a string `name` and a string `message`. A bare `{ name: 'AbortError' }` is still not an abort, because an error carries a message.

The new cross-realm test builds a foreign error through `node:vm` and **asserts its own premise** (`foreign instanceof Error === false`) before asserting the behaviour, so it cannot pass for the wrong reason. Red-proved: with the `instanceof` check restored, that assertion fails.

PR #1698, opened 2026-08-12

woojubb:
Promotes the verified develop tree to main while preserving main ancestry. The repository promotion tool completed the main-only release verification locally, and the promotion tree is identical to origin/develop.


woojubb:
Closing this promotion because the latest automated review found one actionable execution-ownership race. The fix will land through develop first, then the promotion will be regenerated.

Issue #1743, opened 2026-08-15
CORE-039: the universal JSON-schema subset treats an object as a leaf (root item for CORE-037)


woojubb:
Root item filed under `.agents/rules/finding-depth.md` for the `DEPTH: FOUNDATIONAL` verdict on **CORE-037** / #1737 (2026-08-16).

Task file: `.agents/tasks/CORE-039-universal-schema-subset-treats-object-as-a-leaf.md`

## Cause

`IParameterSchema` (`packages/agent-core/src/interfaces/provider.ts:78-90`) — the universal JSON-schema subset every tool schema and structured-output schema is expressed in — can name a nested object's fields but cannot state which are **required**, and has no **union** form. Four independent walks over that subset each re-decide what an `object` node means:

| Walk | Nested object behaviour |
| --- | --- |
| `schema/zod-to-json-schema.ts:96-97` (produce) | returns bare `{ type: 'object' }` — the reported symptom (CORE-037) |
| `schema/structured-output.ts:110` (validate) | `'required' in schema` branch is statically dead at every nested level |
| `tool-registry/parameter-validator.ts:50-54` (validate) | `case 'object'` stops at `typeof` |
| `agent-provider-gemini/src/gemini/tool-schema-converter.ts:61-95` (emit) | forwards `properties`, has no `required` to forward |

The first two are called twelve lines apart on the same schema (`tool-registry/function-tool.ts:65` input, `:102` output) and disagree about how deep `object` goes.

## Why the in-place fix is not enough

Two shipped built-ins are already broken by this (measured on `develop`): `Computer`'s `act` tool emits `action: { type: 'object' }` with all 13 action fields absent, and `AskUserQuestion` emits `questions.items` with every question field absent.

Recursing at `zod-to-json-schema.ts:96` **without** adding union support makes `packages/agent-tools/src/builtins/ask-user-question-tool.ts:169` — a module-level `export const askUserQuestionTool = createAskUserQuestionTool()` whose nested `QuestionSchema.options` is `z.array(z.union([...]))` — hit the `default:` throw at `:143` **at import time**, taking `@robota-sdk/agent-tools` down on load. A silent field loss becomes a package that will not import.

[...]

Issue #1737, opened 2026-08-15
bug(agent-core): zodToJsonSchema drops nested object properties — tools with nested schemas are never invoked


woojubb:
## Summary

`zodToJsonSchema` converts a nested `z.object(...)` to a bare `{ "type": "object" }`, discarding its `properties` and `required`. The model therefore receives a tool schema that says "this field is an object" with no indication of what goes in it. Downstream, `createZodFunctionTool` tools with any nested object are **never successfully invoked**.

Verified on `3.0.0-beta.78` (published) and confirmed still present in `develop` source at `3.0.0-beta.79`.

## Reproduction

```ts
zodToJsonSchema(z.object({ inner: z.object({ s: z.string() }) }))
```

Actual:

```json
{
  "type": "object",
  "properties": { "inner": { "type": "object" } },
  "required": ["inner"]
}
```

Expected: `inner` carries its own `properties` / `required`.

## Root cause

`packages/agent-core/src/schema/zod-to-json-schema.ts:95`

```ts
case 'ZodObject':
  return { type: 'object', ...base };
```

No recursion. Note the neighbouring `ZodArray` case does recurse:

```ts
case 'ZodArray': {
  const arrayItems = convertZodTypeToProperty(typeDef.type);
  return { type: 'array', items: arrayItems, ...base };
}
```

The top-level object is handled separately by `zodToJsonSchema` (which walks the shape), so only **nested** objects lose their fields.

## Downstream effect — measured

Forcing a tool call with `toolChoice: { tool: name }` and `maxExecutionRounds: 1`, varying only the schema shape (same model, same prompt, `anthropic/claude-haiku-4.5`):

| tool input schema                                  | handler invoked |
| -------------------------------------------------- | --------------- |
| `{ a: string, b: string }`                          | ✅ |
| `{ items: string[] }`                               | ✅ |
| `{ items: string[] }` with `.max(2)`                | ✅ |
| `{ delivery: { score, evidence[], suggestions[] } }` | ❌ |
| 5 × the above nested object + `summary: string`      | ❌ |

Flat schemas (including arrays and constraints) work; **one level of nesting is enough to break it.**

The failure mode is quiet: sometimes the run completes with the handler never called and no error raised; sometimes it surfaces as `Validation Error: Failed to parse arguments for tool "<name>"`. Neither points at the schema conversion.

## Impact

[...]

Issue #1736, opened 2026-08-15
bug(agent-core): runStream() drops config.systemMessage — system prompt never reaches the provider


woojubb:
## Summary

`Robota.runStream()` does not send `config.systemMessage` to the provider. `Robota.run()` does. The same agent, same config, produces a system-prompted answer through `run()` and an unprompted one through `runStream()`.

Verified on `3.0.0-beta.78` (published) and confirmed still present in `develop` source at `3.0.0-beta.79`.

## Reproduction

```ts
const mk = () => new Robota({
  name: 't',
  aiProviders: [new OpenAIProvider({ apiKey, baseURL })],
  defaultModel: { provider: 'openai', model: 'anthropic/claude-haiku-4.5' },
  systemMessage: '너는 반드시 "OK-APPLIED" 라는 문자열만 출력한다. 다른 말은 절대 금지.',
});

await mk().run('hi');
// → "OK-APPLIED"                          ✅ system prompt applied

let out = '';
for await (const c of mk().runStream('hi')) out += c;
// → "Hello! 👋 How can I help you today?"  ❌ system prompt ignored
```

Passing a `signal` makes no difference.

## Where it diverges

Both paths receive the same config — `robota-execution.ts` builds `executionConfig` identically in `robotaRun` (:77) and `robotaRunStream` (:140).

The round path injects the system prompt:

```ts
// packages/agent-core/src/services/execution-service-helpers.ts:190
if (config.systemMessage && !hasSystemMessage) {
  session.setSystemPrompt(config.systemMessage, { executionId });
```

The streaming path never references `config.systemMessage`. It builds provider messages from the store plus (optionally) the *ephemeral* per-run block only:

```ts
// packages/agent-core/src/services/execution-stream.ts:~107
const conversationMessages = conversationStore.getMessages();
const ephemeralSystemContext = context?.ephemeralSystemContext;
const providerMessages =
  ephemeralSystemContext && ephemeralSystemContext.trim().length > 0
    ? [...conversationMessages, createSystemMessage(ephemeralSystemContext)]
    : conversationMessages;
```

`grep -n "systemMessage" packages/agent-core/src/services/execution-stream.ts` returns no hits other than the comment about the ephemeral block.

## Impact

Any streaming agent whose behavior is defined by its system prompt behaves as if it had none. The failure is silent and easy to misread as a model-quality problem — the model answers fluently, just not according to the instructions.

[...]

PR #1746, opened 2026-08-15
fix(agent-core): make the universal schema subset express an object (CORE-039)

woojubb:
Fixes **CORE-039** (issue #1743), which delivers **CORE-037** (issue #1736's sibling, #1737).

## What was reported, and what it turned out to be

An external user reported that `zodToJsonSchema` converts a nested `z.object()` to a bare
`{ type: 'object' }`, so tools with one level of nesting are never successfully invoked (#1737).
CORE-037's stated Direction was "recurse for `ZodObject`, mirroring `ZodArray`".

`finding-depth-triager` returned **`DEPTH: FOUNDATIONAL`** on that problem statement, and it was
right twice over:

- **The naive fix crashes the tools package at import.**
  `packages/agent-tools/src/builtins/ask-user-question-tool.ts` constructs its tool at **module
  level**, and its schema nests `z.array(z.union([...]))`. Because the `ZodObject` early return means
  nothing inside a nested object is ever visited, the converter's `default: throw` is unreachable one
  level down today. Recursing without adding union support makes
  `import '@robota-sdk/agent-tools'` throw.
- **Two shipped built-ins were already broken, not one.** Measured with real Zod before any change:
  `Computer`'s `act` tool emitted `action: { type: 'object' }` with all 13 action fields absent, and
  `AskUserQuestion` emitted `questions.items: { type: 'object' }` with every question field absent.

So the root item was filed at `.agents/tasks/CORE-039-*.md` and registered as issue #1743, per
`.agents/rules/finding-depth.md`; CORE-037 carries a re-plan note naming it. **Disposition: re-plan**
— the cause is fixed here rather than contained.

## Accepted recommendation

`REVIEW VERDICT: ENDORSE` — `proposal-reviewer`, 2026-08-16, at revision 2 of 2 (`REVISE` on v1 and
v2), with three binding pre-merge conditions. A further `REVISE` was issued and applied during
implementation. The full endorsed plan is in the item's `## Direction`.

**The cause.** `IParameterSchema` — the universal JSON-schema subset every tool schema and every
structured-output schema is expressed in — was an incomplete type: a nested object could not say
which of its fields were mandatory, its `additionalProperties` was narrower than the root's, and it
had no union form. **Seven** independent walks over that subset each re-decided what a node means.
The two core validators are called 37 lines apart on the same schema by the same method
(`tool-registry/function-tool.ts`, input at one and output at the other) and disagreed about how deep
`object` goes.

**Rejected alternative:** adopting the `zod-to-json-schema` npm package, already a dependency of
`dag-node`/`dag-core`/`agent-cli`. It would permanently dissolve the unsupported-construct tail —
stated because omitting a rejected alternative's real advantage is how a rejection becomes a
rationalisation — but it emits full JSON Schema with `$ref`/`definitions`/`allOf`, and every
downstream mapper here is field-enumerated, so an unrecognised node arrives as `{}`. That is a worse,
quieter failure than the one being fixed. Adopting it would mean widening the subset to all of JSON
Schema — a published-contract redesign the rules put with the user, not the agent.

## Implementation

- **The subset type is completed.** `required` and `anyOf` added; `additionalProperties` widened to
  `boolean | IParameterSchema`; `type` made **optional**, because a union node carries `anyOf`
  *instead* of a type and emitting both is invalid JSON Schema. `IObjectParameterSchema` replaces the
  root object literal that was written out twice purely to regain `required`.
- **One object walk in the converter**, used by the root and every nested object.
  `ZodEffects`/`ZodOptional`/`ZodDefault` unwrap through the same helper at both levels, and a
  non-object root now throws instead of silently emitting an empty schema — `z.object({}).refine(...)`
  used to produce the reported symptom at the root of the very function being fixed.
- **Zod's three unknown-key modes stop collapsing into two emissions.** `strip` (the default) means
  "accept then drop", so it emits `additionalProperties: true`; `strict` emits `false`. Emitting
  nothing read as *closed* to every consumer, i.e. the SSOT producer was mistranslating the author's
  own schema.
- **`ZodUnion`/`ZodDiscriminatedUnion` → `anyOf`, `ZodLiteral` → single-value `enum`.** Absorbed
  rather than filed because a defect sits behind each: unions are forced by the import-time crash,
  and `computer-tool.ts` documents a shipped tool whose schema was *distorted* to dodge discriminated
  unions.
- **`anyOf` is specified into every walk that must see it** — the deep validator, the tool-input
  validator, registration-time schema validation, the Gemini mapper, and Anthropic's
  `closeObjectSchemas`. Adding it to the producer alone would have traded the import-time crash for
  runtime uncallability of the same tool.
- **The two validation walks stop disagreeing.** The tool-input walk keeps its caller-facing leaf
  messages and delegates all depth to `validateAgainstJsonSchema`, the one complete traversal.
- **Closure is defined relative to a declared `properties` set** (see "A claim that did not survive"
  below).
- **Absorbed while editing them:** registration refused `integer`/`null` parameters that the subset
  itself defines, and the input walk's switch had no `default`, so a type-less node would have been
  accepted silently once `type` became optional.

### The three binding conditions

1. **A seventh walk.** `ToolRegistry.validateToolSchema` runs on every `register()` and demanded a
   `type` on each top-level property — so a union-typed argument would convert correctly and then be
   refused at registration: the crash relocated, not removed. Fixed, with the `integer`/`null`
   omission absorbed alongside.
2. **No silent accept.** The input walk's switch gained a `default` that errors.
3. **Message coherence.** The delegated call is rooted at `Parameter "<key>"`, not a bare `$`, so one
   `ValidationError` carries one dialect. Observable in the evidence:
   `Parameter "report".score: required property missing`.

## A claim that did not survive implementation

The reviewer's ruling on closure rested on its own verified finding that no in-repo hand-written
schema used a bare nested `{ type: 'object' }`. That was false — its repository sweep had excluded
test files, and `packages/agent-tools/src/__tests__/function-tool.test.ts` declares exactly that and
asserts a populated payload is accepted. The failure surfaced when the change landed. Re-ruled with
the evidence:

> `additionalProperties` declares closure **relative to a declared `properties` set**. A node that
> declares `properties` — including an empty one — is closed unless `additionalProperties` says
> otherwise. A node that declares **none** permits any properties, as JSON Schema says and as every
> provider reads the document we forward them.

This completes the convention at a position it was never written for rather than carving an exception
out of it: the rule was authored for a tool's `parameters` root, where `properties` is structurally
always present. **The test was left untouched** — it encodes JSON Schema semantics correctly, and
editing the one in-repo detector to preserve the claim that there was nothing to detect is the wrong
direction.

## Process record

`DONE-GATE-STAGE-1` returned **NON-COMPLIANCE** on its first run: implementation had begun before the
gate produced a verdict, though all four criteria were met. The finding was accepted, the
implementation reverted, and the gate re-run — it verified the revert independently, confirmed the
item's diff since the scenario commit was strictly append-only (so no expected result was fitted to
observed output), and reconstructed a scenario script from the document's assertions alone to prove
the restored tree was unfixed in behaviour and not merely in `git status`. **Both runs are kept in
the item.**

## Verification

- `pnpm harness:verify-like-ci` — green. (`dependency audit`, `windows-shell` and `review-gate` are
  not mirrored locally; the last two are relevant to this diff and must be read from the PR checks.)
- Tests: agent-core 999, agent-tools 253, gemini 154, anthropic 95, openai 162, openai-compatible 101,
  agent-tool-mcp 22 — all green.
- New tests are written **against real Zod**, not the `_def` mocks that let this defect pass twenty
  green tests, and pin both built-ins' emitted schemas so the regression cannot recur unnoticed.

### User execution test scenario gate — VERIFIED

Three scenarios, all `agent-executable` and **provider-free** (credential probe recorded: no provider
key in this environment). Each was proven to FAIL on the unfixed tree before being written down, and
all three were re-run against the completed implementation. Scripts and full evidence are inlined in
the item.

| Scenario                                                     | Pre-fix                                             | Post-fix              |
| ------------------------------------------------------------ | ---------------------------------------------------- | --------------------- |
| Built-ins advertise their nested fields                      | `SCENARIO 1 FAIL (8)`, `EXIT:1`                     | `SCENARIO 1 PASS`, `EXIT:0` |
| A tool enforces the nested schema it advertises              | `SCENARIO 2 FAIL (4)`, `EXIT:1` — walks disagree **both** ways | `SCENARIO 2 PASS`, `EXIT:0` |
| A union-typed field is expressible and stays usable          | `Unsupported Zod type: ZodUnion`, `EXIT:1`          | `SCENARIO 3 PASS`, `EXIT:0` |

An earlier draft of scenario 2 was **rejected on substance**: it built the tool with
`createZodFunctionTool`, whose wrapper re-validates with Zod inside the executor, so it would have
passed on unfixed code — observing Zod rather than the walk under change.

## Residual risks and what is deliberately left

- **Gemini has no provider-free user observable.** `convertToolsToGeminiFormat` is not re-exported and
  `GeminiProvider` has no `baseUrl` passthrough, so its `required`/`anyOf` forwarding is covered by
  the engineering test plan only — which is not user-execution evidence. Recorded in the item rather
  than papered over; widening that package's public surface for observability alone was out of scope.
- **`strictTools` on OpenAI still rejects nested objects.** Not a regression, but this change is what
  makes nested schemas reach providers intact, so the exception is now stated where a user meets it
  (that package's SPEC and the option's JSDoc) and filed as **PROV-007**.
- **Filed, not absorbed:** **CORE-040** (the two `agent-tool-mcp` validators bypass the subset
  validator), **CORE-041** (the converter's remaining unsupported Zod constructs), **PROV-007**. The
  boundary is "is there a defect behind it", not "how much work is it".
- **Lint warnings.** Three functions remain over the line/complexity limits — all three were already
  over before this change, and two improved (`validateAgainstJsonSchema` complexity 50→45,
  `convertZodTypeToProperty` 88→69 lines). No new violation is introduced. Two files that grew past
  their frozen size baselines were **split by responsibility** rather than extended, and the baseline
  was re-tightened in the same change.


woojubb:
## Round A — local diff review, resolved before this PR existed

`pr-review-reviewer` reviewed `git diff origin/develop...HEAD` locally, before the PR was opened.
**ACTIONABLE FINDINGS: 6** (2 MUST, 4 SHOULD). All six are **fixed**; none refuted, none deferred.

| # | Sev | Finding | Resolution |
| --- | --- | --- | --- |
| 1 | MUST | `ZodNullable` was unwrapped to its inner type and the **null branch dropped**, so an advertised `{type:'string'}` rejected a `null` that Zod itself accepts. Harmless while nested nodes were opaque; a real false rejection once depth is enforced. | Emits `anyOf: [<inner>, {type:'null'}]`, using the member this change added. The mock test that pinned the loss was updated — it encoded a defect. `b33811569` |
| 2 | MUST | The changeset omitted `@robota-sdk/agent-provider-openai-compatible`, whose shipped source this change edits, and `@robota-sdk/agent-tools`. | Both added. `b33811569` |
| 3 | SHOULD | The tool-input walk's leaf/deep split was by **kind**, so constraint enforcement depended on position: a `minimum` was checked inside a nested object and ignored at the top level, and two message dialects appeared at the same position depending on the declared type. | Split by **concern** instead: the leaf type mismatch and its caller-facing wording stay, everything else goes to the one complete walk, for every node. `b33811569` |
| 4 | SHOULD | `isRequiredField` peeled one `ZodEffects` level while the shared unwrap looped, so `.optional().refine(f).refine(f)` was reported **required**. | Loops now, capped like the shared unwrap. `b33811569` |
| 5 | SHOULD | Three documents written **earlier in this PR** stated the subset "leaves `additionalProperties` unset (closed by convention)" — which step 3 of the same PR made false — and understated the strict-mode blast radius. | Corrected in all three; the strict-mode consequence now says **every** Zod-derived tool is refused, not only nested ones. The moved `closeObjectSchemas` citation was repointed. `b33811569` |
| 6 | SHOULD | Changeset was `patch` across the board. | `agent-core` is now `minor` — a new export plus `type` becoming optional, which needed two call-site edits in this repo alone. `b33811569` |

**CONSIDER/NIT also absorbed** rather than left: registration now validates nested nodes (newly constructible once `type` became optional), `z.literal(null)` converts instead of throwing, the wrapper unwrap is depth-capped because `IZodSchema` is a structural stand-in at an exported boundary, the conversion option reaches nested objects, and the leaf message no longer reports `got object` for a `null`.

### Independent red-proof

The reviewer ran this PR's new tests against the **pre-fix merge-base** (`3d1e63320`) in an isolated worktree: **32 genuine assertion failures**. It also caught that the `agent-tools` case had first passed only because the symlink resolved to the *fixed* `agent-core`, re-pointed it, and got 6/6 failures. No accidental-green.

### After the fixes

- `pnpm harness:verify-like-ci` — **12/12 stages pass**.
- All affected suites green: agent-core 1007, agent-tools 253, gemini 154, anthropic 95, openai 162, openai-compatible 101, agent-tool-mcp 22.
- The three user-execution scenarios were re-run against the reviewed tree and all pass; the item's evidence is repointed at it.
- The converter crossed the 300-line ceiling as these fixes landed and was **split by responsibility** (conversion → inspection, one-directional) rather than baselined.

`dependency audit`, `windows-shell` and `review-gate` have no local mirror; the last two are relevant to this diff and must be read from the checks on this PR.

woojubb:
## Two CI-only failures — both real, both fixed

The first push was red on `commitlint` and `regression-red-proof`. Neither reproduced in `pnpm harness:verify-like-ci` locally, and both turned out to be genuine.

### `commitlint` — a citation the next reader could not follow

One commit cited `afd397c10`. That hash had been rewritten away by an *earlier* history correction on this branch (the one that made every commit message name resolvable repository paths). The repo's `claims-resolve` rule checks that a cited hash names an object — and it does, in my local clone, because the object was still in the object database. On CI's fresh clone it does not exist. Local green, CI red, and CI was right.

Fixed by rewriting that message, and then by removing the whole class: **no branch-internal commit SHA is cited in the item's own evidence any more**. A hash written into a document on the same branch breaks every time that branch's history is corrected. The guardians' recorded verdicts are kept verbatim, SHAs included — editing a recorded verdict to tidy it is worse than a stale citation — with one appended note explaining that the history was corrected twice and that what they verified is unchanged.

### `regression-red-proof` — the gate fired on correct work

`packages/agent-core/src/interfaces/tool-schema.ts` was reported `accidental-green-fail`. It is a **type-only module**, created in this PR because the file-size floor required splitting `interfaces/provider.ts`. Types are erased at transpile, so reversing that file cannot fail any test — it is not a weak regression test, it is a file for which a runtime red proof does not exist. The gate already reaches exactly that verdict for its two *already-existing* siblings (`interfaces/provider.ts`, `schema/zod-schema-types.ts`): `inconclusive (type/comment-only change; runtime red proof is not applicable)`.

The cause: an ADDED file has no base text, and `reversedText === null` short-circuited straight to "this is a runtime mutation". An empty file also has neither import nor export, so it is emitted as a *script* with a `"use strict";` prologue the module form lacks — a difference of framing, not behaviour.

**Fixed in the gate, not worked around.** The documented escape (`allow-green-at-base: <reason>`) skips the gate for the **entire range** — it would have disabled the six genuine red proofs in this PR to accommodate the one file the gate cannot judge. `git-branch.md` is explicit that a check firing on correct work is the thing that changes. The base for an added file is now an empty *module*, with two self-tests: an added type-only module is not a runtime mutation, an added module that emits code still is.

After the fix, the same run reports six `red-proof-ok`, and `tool-schema.ts` joins its siblings at `inconclusive (type/comment-only change)`.

> **Flagging this deliberately:** modifying an enforcement floor is outside the scope this PR's recommendation gate endorsed. It is stated here rather than buried. If you would rather it were reverted, the alternative is the range-wide opt-out, which switches off red-proof checking for this PR entirely.

### State now

- `pnpm harness:verify-like-ci` — **12/12**, and `harness-self-test` covers the two new cases (2309 harness tests green).
- Branch force-pushed with the corrected history; the diff content is unchanged apart from the gate fix and the de-cited evidence text.

Issue #1748, opened 2026-08-15
CORE-042: agent-core implements the execution turn twice (root item for CORE-036)


woojubb:
Root item filed under `.agents/rules/finding-depth.md` for the `DEPTH: FOUNDATIONAL` verdict on **CORE-036** / #1736 (2026-08-16).

Task file: `.agents/tasks/CORE-042-the-execution-turn-is-implemented-twice.md`

## Cause

`agent-core` documents ONE execution engine with two entry points, and states its turn guarantees unconditionally (SPEC § System Prompt names `initializeConversationStore` as the mechanism). There is no shared turn. The round path composes named steps — `buildFullExecutionContext` → `resolveProviderAndTools` → `initializeConversationStore` → `runExecutionLoop`/`executeRound` → `callProviderWithCache` → `finalizeExecution`. `executeStream` re-derives every one of them inline and owns none, so **no seam exists that a new turn capability must pass through**. Parity is a convention held by reviewer memory, and its failures are silent: the model still answers — just without the prompt, the token cap, the plugin, or the tool.

## This is the seventh instance

Every behavioural commit to `execution-stream.ts` is the identical patch — "streaming dropped X, copy X in":

| Commit | Item | Dropped |
| --- | --- | --- |
| `d2015a40a` | CORE-016 | `maxTokens`/`temperature` — "silently dropped **ALL** model options". External report, beta.76 |
| `03a83f3d8` | CORE-017 | `toolChoice` |
| `bda1d4cfa` | CORE-018 | `signal` — "made the public streaming API uncancellable" |
| `8866de037` | CORE-020 | response validation |
| `6f308d102` | BEHAVIOR-005 | token usage on the committed assistant message |
| `4fc3ec266` | SELFHOST-008 P3 | `ephemeralSystemContext` — landed as a *review SHOULD* |
| — | **CORE-036** | `config.systemMessage`. **External report, beta.78** |

`SPEC.md:474` institutionalises it with a heading — "**runStream path (parity)**". A parity heading in a specification is the contract admitting it has two implementations.

## Still live and unfiled

[...]

PR #1749, opened 2026-08-15
fix(agent-core): apply config.systemMessage on the streaming path (CORE-036)

woojubb:
Fixes **CORE-036** (issue #1736). Files **CORE-042** (issue #1748) as its root cause and contains this fix under it.

## What was reported

An external user found that `Robota.run()` and `Robota.runStream()` receive an identical
`executionConfig`, but only `run()` applies `config.systemMessage`. An agent whose entire behaviour is
defined by its system prompt obeys it through one entry point and ignores it through the other —
silently, on the default interactive/TUI surface, in a published beta (3.0.0-beta.78).

The user-visible shape, measured on the unfixed tree:

```
provider call 1 (runStream): [{"role":"user","content":"turn one"}]
provider call 2 (runStream): [{"role":"user",…},{"role":"assistant",…},{"role":"user","content":"turn two"}]
provider call 3 (run):       [{"role":"system","content":"You are PERSONA-ALPHA."}, …]
```

An agent used through both entry points acquires its persona only from the first non-streaming turn
onward, and the prompt then appears **mid-conversation**.

## The depth verdict changed what this PR is

`finding-depth-triager` returned **`DEPTH: FOUNDATIONAL`**, and the evidence is not arguable: **every
behavioural commit ever made to `execution-stream.ts` is the identical patch** — "the streaming path
dropped X that the round path has; copy X in".

| Commit | Item | Dropped |
| --- | --- | --- |
| `d2015a40a` | CORE-016 | `maxTokens`/`temperature` — it had "silently dropped **ALL** model options". External report, beta.76 |
| `03a83f3d8` | CORE-017 | `toolChoice` |
| `bda1d4cfa` | CORE-018 | `signal` — "made the public streaming API uncancellable" |
| `8866de037` | CORE-020 | response validation |
| `6f308d102` | BEHAVIOR-005 | token usage on the committed assistant message |
| `4fc3ec266` | SELFHOST-008 P3 | `ephemeralSystemContext` — landed as a *review SHOULD*, i.e. caught by reviewer memory |
| — | **CORE-036** | `config.systemMessage`. **External report, beta.78** |

The cause is that agent-core declares one execution-turn contract and implements it twice: the round
path composes named steps, and `executeStream` re-derives every one of them inline, so **no seam
exists that a new turn capability must pass through**. `SPEC.md:474` institutionalises this with a
"**runStream path (parity)**" heading — a parity heading in a specification is the contract admitting
it has two implementations.

Filed as **CORE-042** / issue #1748, with five further divergences that had no task
(`beforeProviderCall`/`afterProviderCall` never fire on streaming; a different tool-list predicate
producing a different tool list; skipped tool-description validation; no cache lookup; no capacity
guard), and with the reason nothing catches them: the sanctioned shared test double
(`createScriptedProvider`, exported from `@robota-sdk/agent-core/testing`) has no `chatStream`, so the
blessed way to exercise a turn end-to-end covers exactly one of the two implementations.

## Why this lands as containment rather than the root fix

`finding-depth.md` permits containment under three conditions; all three hold, and the disposition was
endorsed by `proposal-reviewer`:

1. **It must land first** — a declared SPEC invariant is false in a published beta, and the failure is
   mis-attributable (the model answers fluently, just not under its instructions, which reads as a
   model-quality problem).
2. **It is the smallest thing that keeps the tree honest**, and **introduces no new abstraction** — it
   calls the helper that is already § System Prompt's named mechanism.
3. **It names the root item** in a code comment and in the commit body.

The argument that actually distinguishes this hold from the six patches before it: **a contained fix
should be judged on whether it makes the root fix harder, and this one makes it easier.**
`executeStream` goes from owning *zero* of the round path's named steps to entering one of them — a
step along CORE-042's direction. Each of the six prior patches did the opposite: it added a duplicated
clause and raised the seam's eventual cost.

## The change

One call site. `executeStream` enters the same `initializeConversationStore` the round path enters,
passing the `messages` parameter it was already given and had been ignoring — so the prompt, the
inject-once rule (CORE-009/CORE-010) and the conversation restore are owned in one place. A
`Contained — CORE-042.` comment at the site records that this makes the streaming path *call* one
shared helper while remaining a second implementation of the turn.

**Tests are table-driven over both entry points**, placed beside the CORE-016/017/018 parity pairs and
using the `TrackingProvider` that already backs them. Asserting one entry point at a time is what let
the repeat happen; a table fails when the two diverge, instead of when someone remembers to look.
Three of the streaming cases are red on the unfixed path — verified by reverting the source and
re-running, not assumed.

## Verification

- `pnpm harness:verify-like-ci` — **12/12 stages pass**.
- agent-core: **1014 tests green**.
- No SPEC change, deliberately: § System Prompt already states the guarantee unconditionally and
  already names `initializeConversationStore`. The code moved to the spec, not the spec to the code —
  and per-path wording would institutionalise exactly the parity framing CORE-042 was filed against.

### User execution test scenario gate — VERIFIED

Four scenarios, all `agent-executable` and **provider-free**, each inlined verbatim in the item because
`scratch/src/` is gitignored. They drive `AbstractAIProvider` — a public extension point — exactly as
an integrator would, and observe the provider request the SDK actually built.

The live-provider draft was **replaced, not refined**. It was unrunnable here (probe recorded: no
provider key, no populated `.env`) and it was the weaker observable even with a key: "the model replied
`OK-APPLIED`" is probabilistic evidence about model compliance, while "the provider received a `system`
message whose content is `config.systemMessage`, at index 0" observes the thing this change alters.

| Scenario | Pre-fix | Post-fix |
| --- | --- | --- |
| 1 — the prompt reaches the provider on `runStream()`, identically to `run()` | `FAIL (4)`, `EXIT:1` — the streaming request carried no system message | `PASS`, `EXIT:0` — both requests byte-identical |
| 2 — exactly one system head, ever (two streaming turns + a `run()`) | `FAIL (6)`, `EXIT:1` — the `run()` injected the head mid-conversation | `PASS`, `EXIT:0` |
| 3 — `ephemeralSystemContext` reaches the request, never the store | `FAIL (3)`, `EXIT:1` — the ephemeral block was the *only* system message | `PASS`, `EXIT:0` |
| 4 — `updateSystemPrompt()` still wins (regression guard) | `PASS` — labelled honestly as passing pre-fix | `PASS`, reproducing it |

Scenario 4 is **not** cited as evidence the change landed; the stage-1 gate attached that constraint
explicitly. It guards the regression this fix risks — routing through `initializeConversationStore`
runs `setSystemPrompt`, so a mis-ordered implementation could revert a live-updated head or append a
second one — which scenarios 1–3 structurally cannot see.

## Residual risks and stated decisions

- **Two observable effects, not one.** Besides the prompt reaching the provider, `getHistory()` now
  carries a system head on a streaming-only agent, which reaches `agent-session` persistence. It is
  correct (parity with `run()`; SPEC § System Prompt's resume semantics do not restore persisted
  `system` messages) but it is public, and it is recorded rather than left to be discovered.
- **Deliberately not covered:** the persistence/resume leg of that effect, which needs an
  `agent-session` store fixture. Named as a decision, not an omission.
- **Deliberately not absorbed:** the five other divergences. They belong to CORE-042 — absorbing them
  would be re-plan wearing containment's clothes, and each is a behaviour change to a published beta
  deserving its own gate. They are enumerated in CORE-042 under a binding sentence: none may be dropped
  from its scope without being filed as its own task.
- **Debug logging collapsed.** The file-size floor refused the fix (the call plus its label pushed the
  file past its frozen baseline) and the floor says split rather than extend — but containment forbids
  a new abstraction, so the added lines were paid for out of noise instead: five debug calls reported
  one fact, the tool count, in five phrasings. The file is now below its old baseline and the ratchet
  is re-tightened in the same change.
- **Two acknowledgments** were added to `scripts/harness/progress-report-acknowledgments.json`. The
  quantified-progress rule requires a mid-work ratio to carry its percentage; this session's status
  update reported one without, and the message acknowledging that quoted the ratio and did it again.
  A transcript is append-only, so both are recorded rather than edited away.

### DIFF
--- a/.agents/tasks/CORE-036-runstream-never-applies-config-systemmessage.md
+++ b/.agents/tasks/CORE-036-runstream-never-applies-config-systemmessage.md
@@ -113,19 +113,391 @@ whether the two are better delivered together against one shared session-prepara
 
 ## User Execution Test Scenarios
 
-Applies — this changes observable SDK behavior.
+Applies — this changes observable behavior of the published `@robota-sdk/agent-core` SDK surface.
 
-**Scenario 1 — the same system prompt survives streaming**
+**Surface: provider-free, public extension point.** The scenarios drive the SDK exactly as a
+third-party integrator does — a user-written provider implementing `AbstractAIProvider` (exported
+from `@robota-sdk/agent-core`) whose `chat`/`chatStream` record the `TUniversalMessage[]` the SDK
+hands them. **No API key, no network.**
 
-- Prerequisites: a provider API key (or an OpenAI-compatible gateway `baseURL`) exported in the
-  environment; a workspace build (`pnpm build`).
-- Environment: uses the existing examples surface — no new fixture required. Confirm at implementation
-  time which example under `apps/`/`examples/` is the shortest streaming path; if none exists, the
-  work adds a minimal script under the examples surface.
-- Steps: construct one `Robota` with
-  `systemMessage: 'Reply with exactly the string OK-APPLIED and nothing else.'`; call `run('hi')`,
-  print the result; construct an identical agent, iterate `runStream('hi')`, concatenate the chunks,
-  print the result.
-- Expected observable result: both prints are `OK-APPLIED`. (Before the fix, only the first is.)
-- Cleanup: none — no state is persisted.
-- Evidence: _to be filled after implementation_ (paste both printed outputs).
+The live-provider draft this section replaces was rejected on two independent grounds. It is
+unrunnable here — probe recorded 2026-08-16: no `OPENAI_API_KEY`/`ANTHROPIC_API_KEY`/`GOOGLE_*` is
+set, and `find . -maxdepth 3 -name ".env*"` returns only `.env.example` files, none populated. More
+importantly it was the _weaker_ observable even with a key: "the model replied `OK-APPLIED`" is
+probabilistic evidence about model compliance, while "the provider received a `system` message whose
+content is `config.systemMessage`, at index 0" is a deterministic observation of the thing this
+change actually alters.
+
+**Invocation.** Scripts live in `scratch/src/` (the repo's sanctioned home for disposable
+live-verification scripts) and are reproduced in full below, because that directory is gitignored and
+the item is therefore their only durable home. `pnpm run run` is broken in this environment —
+`scratch/node_modules/.bin` was never linked, because `pnpm install` aborted on `better-sqlite3`'s
+native build (`make`/`g++` absent) and the workspace was installed with `--ignore-scripts`. Every
+command below uses `node ../node_modules/tsx/dist/cli.mjs --conditions=source src/core-036-<n>.ts`, run from `scratch/`.
+
+**Stated limitation.** `--conditions=source` exercises the TypeScript sources, not `dist/`. That is a
+real gap versus what a package consumer installs, and it is forced here by the same missing native
+toolchain. Where a working toolchain exists, dropping the flag after `pnpm build` strengthens every
+scenario below at no cost to its design.
+
+**Shared harness**
+
+```ts
+// scratch/src/core-036-lib.ts
+/**
+ * CORE-036 shared harness for the user-execution scenarios.
+ *
+ * RecordingProvider is written against the PUBLIC extension point (`AbstractAIProvider`,
+ * exported from `@robota-sdk/agent-core`) exactly as a third-party integrator would write a
+ * provider. It records the message array the SDK hands it, so the scenarios observe the real
+ * provider request with no API key and no network.
+ */
+import { AbstractAIProvider } from '@robota-sdk/agent-core';
+
+import type { IChatOptions, TUniversalMessage } from '@robota-sdk/agent-core';
+
+export class RecordingProvider extends AbstractAIProvider {
+  readonly name = 'recording-provider';
+  readonly version = '1.0.0';
+  /** One entry per provider call: the message array the SDK sent. */
+  readonly received: TUniversalMessage[][] = [];
+
+  async chat(messages: TUniversalMessage[], _options?: IChatOptions): Promise<TUniversalMessage> {
+    this.received.push([...messages]);
+    return {
+      id: `chat-${this.received.length}`,
+      role: 'assistant',
+      content: 'ack',
+      state: 'complete' as const,
+      timestamp: new Date(),
+    };
+  }
+
+  override async *chatStream(
+    messages: TUniversalMessage[],
+    _options?: IChatOptions,
+  ): AsyncIterable<TUniversalMessage> {
+    this.received.push([...messages]);
+    yield {
+      id: `stream-${this.received.length}`,
+      role: 'assistant',
+      content: 'ack',
+      state: 'complete' as const,
+      timestamp: new Date(),
+    };
+  }
+}
+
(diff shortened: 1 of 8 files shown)


woojubb:
## Round A — local diff review, resolved before this PR existed

`pr-review-reviewer` reviewed `git diff origin/develop...HEAD` locally, before the PR was opened.
**ACTIONABLE FINDINGS: 1** (SHOULD), plus 3 CONSIDER and 1 NIT. All resolved; none refuted, none deferred.

### SHOULD — one of my own tests was accidental-green

`a run() after a runStream() does not add a second system head` **passed on the pre-fix code**. The reviewer proved it in an isolated worktree at the merge-base, and the reason is exact: I asserted the later `run()`'s request and the final `getHistory()`, but `setSystemPrompt` normalizes to *exactly one system message at the head* unconditionally — so both counts read `1` on the buggy code too. It asserted a late invariant both versions satisfy.

That mattered more than a spare case. The mixed entry-point scenario — the shape the defect was actually reported from — had its only other red proof in a `scratch/` script, which is gitignored. **CI had no executable guard for it.**

Fixed by asserting the **streaming turn's own request**, which is what is red pre-fix. Red count on the unfixed source goes from three to four:

```
FAIL runStream(): config.systemMessage reaches the provider as the system head
FAIL runStream(): a second turn does not add a second system head
FAIL a run() after a runStream() does not add a second system head   ← newly discriminating
FAIL runStream(): an ephemeral block joins the system head without being stored
```

### CONSIDER ×3, all taken

- **My debug-collapse claims were wrong in three directions.** The code comment said "four calls, two variables"; the commit body said "five calls, three variables, two remaining". The diff removes **six** calls and **four** variables and leaves **one**. Corrected at the site. In a repo whose rules make quantified claims load-bearing, a record wrong in three directions is worth fixing before it becomes the history.
- **A containment label was invisible to the floor that checks it.** The secondary label in the test file was spelled `Contained -- CORE-042`, and the containment-label pattern (`depth-verdict-reachable.test.mjs`) does not match a double hyphen — so it had silently opted out of the check that verifies a label's ID resolves. The *required* label at the call site was correctly spelled; this one now is too.
- **The restore branch's inertness is a property of the CALLER, not of this call.** Pinned at the site: a future caller passing a foreign array would resurrect a cleared history, with `parts` stripped by `robota-history.ts`'s projection.

### NIT

Use the already-narrowed `streamingConversationId` rather than re-reading `context.conversationId`.

### What the reviewer explicitly did *not* raise

That the turn is still implemented twice — `execution-stream.ts` still re-derives provider resolution, chat options, validation and commit. That is the CORE-042 hold the label at the call site answers, and the reviewer confirmed the label says enough for a next reader not to re-raise it.

### One more file-size round

Adding the review's comments pushed `execution-stream.ts` past its (freshly tightened) baseline again. The floor says split, containment forbids a new abstraction, so the lines were paid for out of noise once more — this time the **per-chunk tool-fragment debug log**, which fired once per stream chunk and printed the raw tool-argument fragment verbatim. Worth removing on its own terms: a per-chunk line is noise at debug level, and one that echoes raw arguments is a leak vector.

### State now

- `pnpm harness:verify-like-ci` — **12/12**.
- agent-core: **1014 tests green**; all four user-execution scenarios re-run and passing.

Issue #1750, opened 2026-08-16
CORE-043: structured-output capability has no runtime representation (root item for CORE-038)


woojubb:
Root item filed under `.agents/rules/finding-depth.md` for the `DEPTH: FOUNDATIONAL` verdict on **CORE-038** / #1738 (2026-08-16).

Task file: `.agents/tasks/CORE-043-structured-output-capability-has-no-runtime-representation.md`

**This is not a proposal to implement.** Several load-bearing decisions are reserved for the owner — listed at the end.

## Cause

`agent-core` emits `IChatOptions.responseFormat` on every structured run. Each provider package either maps it onto a native surface or silently discards it, and **the seam has no way to know which happened**. The one place the fact could live — per-model `TProviderModelCapability`, which already contains `'json_schema'` — is read by nothing (PROV-006) and is declared **falsely** for deepseek (all three catalog entries claim it against an adapter that implements none).

Because capability is a property of the **package the caller instantiated** rather than of the **endpoint and model actually called**, the configuration this project documents as its gateway story reports enforcement it does not have: `agent-provider-openai` passes `baseURL` to the SDK and switches to `chat-completions` — the one surface that maps `response_format` — while `llms.txt:22` advertises exactly that path for "any gateway (Vercel AI Gateway, LiteLLM, OpenRouter), Azure, vLLM, Ollama, LM Studio". The endpoint accepts the parameter, the model may ignore it, and the enforcement loop believes it enforced early.

The SPEC already records the cause as intent — § Structured Output Contract: *"providers without one ignore it — the core-side enforcement loop below is the universal contract either way."*

## Two corrections to what #1738 reported

Both verified against `develop` after CORE-039 landed:

1. **There is no first-attempt prose fallback — there is no first-attempt transport at all.** `spec.jsonSchema` has exactly two consumers in the run path (a config override and the retry-feedback prose). Nothing injects the schema into attempt 1, and the compat adapters never read `responseFormat`, so **attempt 1 carries no schema signal whatsoever**. The reporter's 0/4 is not a property of their gateway; in robota it is structurally guaranteed on that family, and `outputRetries: 0` can only succeed there by luck.
2. **`agent-provider-bytedance` is not a text provider.** It implements `IVideoGenerationProvider` and has no `chat()`. Structured output never routes through it; the grep found nothing because there was nothing to find.

## The same defect, filed three times in three days

| Filed | Item | Channel |
| --- | --- | --- |
[...]

Issue #1738, opened 2026-08-15
proposal(agent-core): use a forced tool call as the structured-output fallback (CORE-015)


woojubb:
## Summary

`CORE-015`'s enforcement loop is a good design — the validate-and-retry guarantee is stronger than what most SDKs offer. This proposal targets one piece of it: **the transport used when a provider has no native structured-output surface.**

Today that fallback re-prompts the model in prose:

```ts
// packages/agent-core/src/core/robota-execution.ts:179
function buildRetryFeedbackInput(spec: IStructuredOutputSpec, issues: string[]): string {
  return [
    'Your previous response did not match the required JSON schema.',
    ...
    'Respond with ONLY a JSON object (no prose, no code fences) matching this JSON schema:',
    JSON.stringify(spec.jsonSchema),
  ].join('\n');
}
```

Proposal: **make a forced tool call (`toolChoice: required`) the fallback transport**, and keep the retry loop exactly as it is. The loop is what guarantees correctness; the transport determines how often the loop has to run.

## Why this matters — the fallback is not a rare path

Native `responseFormat` support is present in 3 of 7 provider packages:

| provider package                   | native structured-output surface |
| ---------------------------------- | -------------------------------- |
| `agent-provider-anthropic`         | ✅ `output_config.format`        |
| `agent-provider-gemini`            | ✅ `responseSchema`              |
| `agent-provider-openai`            | ✅ `response_format`             |
| `agent-provider-openai-compatible` | ❌ prompt fallback               |
| `agent-provider-bytedance`         | ❌ prompt fallback               |

`openai-compatible` is the path for DeepSeek, Groq, Together, OpenRouter, vLLM/Ollama, and any gateway — a large share of real deployments.

## A sharper issue: "native surface" is per-endpoint, not per-model

`agent-provider-openai` accepts a `baseURL`. When that URL points at a gateway serving a **non-OpenAI model**, the request still carries `response_format: json_schema`, the endpoint accepts it, and the underlying model may ignore it. The provider cannot tell the difference, so the enforcement loop believes it is on the "enforce early" path while it is effectively on the prompt path.

We hit exactly this shape outside robota. Same schema, same prompt, through an OpenAI-compatible gateway:

| transport                   | DeepSeek v4-flash | Claude haiku-4.5 |
| --------------------------- | ----------------- | ---------------- |
| JSON-schema response format | **0/4**           | 4/4              |
| forced tool call            | **4/4**           | 4/4              |

[...]

PR #1751, opened 2026-08-16
docs(tasks): file CORE-043 and close CORE-038 as superseded

woojubb:
Closes **CORE-038** as `superseded` (issue #1738). Files **CORE-043** (issue #1750) as its cause.

**Documentation and backlog only — no code changes.** Nothing CORE-038 proposed was implemented, and that is the point of this PR.

## What was proposed, and what the depth verdict found

An external user proposed that when a provider has no native structured-output surface, the fallback transport should be a forced tool call rather than a prose re-prompt, with a measurement outside robota: same schema, same prompt, one OpenAI-compatible gateway — JSON-schema response format gave DeepSeek **0/4**, forced tool call gave **4/4**.

`finding-depth-triager` returned **`DEPTH: FOUNDATIONAL`**. The transport may well be the right one; that is not what was rejected. The finding is that the item states the problem one layer above its cause — **a transport chosen without a capability gate replaces one blind emission with another**, and the item's own open question ("does the schema tool collide with the agent's own tools under `tool_choice: required`?") exists only because nothing in the runtime decides *when* to inject.

## Two factual errors in the report, one of which changes the conclusion

Both verified against `develop` after CORE-039 landed. **Recorded rather than edited away**, so the original report stays readable.

1. **There is no first-attempt prose fallback — there is no first-attempt transport at all.** `spec.jsonSchema` has exactly two consumers in the run path: a config override and the retry-feedback prose. Nothing injects the schema into attempt 1, and the compat adapters never read `responseFormat`. So the reporter's 0/4 is **not** a property of their gateway — in robota it is structurally guaranteed on that family, and `outputRetries: 0` can only succeed there by luck. The proposal was reasoning about a transport that does not exist.
2. **`agent-provider-bytedance` is not a text provider.** It implements `IVideoGenerationProvider` and has no `chat()`. Structured output never routes through it; the item's grep found nothing because there was nothing to find, and its `area:` frontmatter named the package wrongly.

## The same defect, filed three times in three days

| Filed | Item | Channel |
| --- | --- | --- |
| 2026-08-13 | PROV-004 | internal critic — carries this exact symptom, this exact evidence, **and a direction** |
| 2026-08-13 | PROV-006 | internal critic — carries the missing mechanism: per-model capability flags read by nothing |
| 2026-08-16 | **CORE-038** | external issue #1738 — the same symptom again, as a transport proposal, reconciled with neither |

The fact has no owner, so each observer filed it as whatever it looked like from where they stood. That is what CORE-043 exists to stop.

## The cause

Structured-output capability has no representation the runtime reads. `agent-core` emits `responseFormat` on every structured run; each provider maps it or silently discards it; the seam cannot tell which. The one place the fact could live — per-model `TProviderModelCapability`, which already contains `'json_schema'` — is read by nothing and is declared **falsely** for deepseek (all three catalog entries claim it against an adapter implementing none).

Because capability belongs to the **instantiated package** rather than to the **endpoint and model actually called**, the configuration `llms.txt` advertises as the gateway story reports enforcement it does not have: `agent-provider-openai` + `baseURL` switches to the one surface that maps `response_format`, sends it to a model that may ignore it, and the enforcement loop believes it enforced early.

The SPEC already records the cause as intent — § Structured Output Contract: *"providers without one ignore it — the core-side enforcement loop below is the universal contract either way."*

## Decisions reserved for you, not for me

`backlog-execution.md` § Agent Decision Authority reserves each of these, so CORE-043 states them rather than planning around them:

1. **Whether `agent-provider-openai` changes behaviour when `baseURL` is set** — a provider default on a documented, published surface.
2. **Where capability and transport selection live** (provider declaration / agent config / `IRunOptions`) — unanswerable before PROV-006's own reserved decision.
3. **Whether the SPEC's "providers without one ignore it" clause is rewritten** — published contract.
4. **Whether a synthetic schema tool may be injected into a user's tool set** under `tool_choice: required` — observable behaviour change for every tool-using agent.

## What proceeded without them

The one piece the triage identified as non-reserved: **stating the compat family's structured-output no-op in its own SPEC.** A user currently has no way to learn it except by measuring — which is exactly how it reached an external report. The guarantee still holds (the bounded loop returns a conforming object or throws); what was undocumented is the **cost**: at least one extra turn by construction, and `outputRetries: 0` succeeding only by luck. It also names the contradiction next door — the deepseek catalog declaring a capability its adapter does not implement.

## Verification

`pnpm harness:verify-like-ci` — **12/12 stages pass**.

## Why `superseded` and not `done`

The reported concern is real and is carried forward in full by CORE-043. But nothing this item proposed was implemented, and four of the decisions it leaves open are yours. Marking it `done` would claim work that did not happen.

### DIFF
--- a/.agents/tasks/CORE-043-structured-output-capability-has-no-runtime-representation.md
+++ b/.agents/tasks/CORE-043-structured-output-capability-has-no-runtime-representation.md
@@ -1,0 +1,141 @@
+---
+title: 'CORE-043: structured-output capability has no representation the runtime reads — agent-core emits `responseFormat` unconditionally, each provider silently maps or discards it, the seam cannot tell which, and the fact is a property of the instantiated PACKAGE rather than the endpoint and model actually called, so the documented gateway configuration reports enforcement it does not have'
+status: todo
+created: 2026-08-16
+priority: critical
+urgency: now
+area: packages/agent-core, packages/agent-provider-openai, packages/agent-provider-openai-compatible
+depends_on: []
+---
+
+# CORE-043: structured-output capability is not represented at runtime
+
+Root item filed under [finding-depth.md](../rules/finding-depth.md) for the `DEPTH: FOUNDATIONAL`
+verdict on [CORE-038](completed/CORE-038-forced-tool-call-as-structured-output-fallback-transport.md)
+(2026-08-16). Registered as [issue #1750](https://github.com/woojubb/robota/issues/1750); the
+proposal it was raised from is [issue #1738](https://github.com/woojubb/robota/issues/1738).
+CORE-038 proposed a transport (forced tool call instead of a prose re-prompt); the
+transport may well be the right one, but it is stated one layer above its cause.
+
+**This item is not a proposal to implement. Several of its load-bearing decisions are reserved for
+the owner** — see § Decisions reserved for the owner. What can proceed without them is listed there
+too.
+
+## Problem
+
+`agent-core` emits `IChatOptions.responseFormat` on every structured run. Each provider package
+either maps it onto a native surface or silently discards it. **The seam has no way to know which
+happened, and no way to react.**
+
+The one place the fact could live is per-model capability: `TProviderModelCapability`
+(`packages/agent-core/src/interfaces/provider-definition.ts:66-67`) already contains `'json_schema'`.
+It is read by nothing (PROV-006), and it is declared **falsely** — all three entries in
+`packages/agent-provider-openai-compatible/src/deepseek/model-catalog.ts` claim `'json_schema'`
+against an adapter that implements none of it.
+
+Because capability is a property of the **package the caller instantiated** rather than of the
+**endpoint and model actually being called**, the configuration the project documents as its gateway
+story reports enforcement it does not have:
+
+- `packages/agent-provider-openai/src/openai/provider.ts:66` passes `baseURL` to the SDK, and `:216`
+  returns `options.baseURL ? 'chat-completions' : 'responses'` — setting a gateway URL switches to
+  the one surface that maps `response_format`.
+- `llms.txt:22` advertises exactly that: "OpenAI-compatible gateway via `baseURL` — any gateway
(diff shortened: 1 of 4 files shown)


woojubb:
**Finding depth — disposition: `re-plan`**

A foundational finding was judged on this PR. Per `.agents/rules/finding-depth.md` a re-plan disposition WITHDRAWS this change rather than patching it, so it is not to be merged: close it and work the root item.

Root item(s): CORE-043
Recorded from `docs/core-038-capability-representation` at `12e8024a2`.

woojubb:
## Note on the `disposition-re-plan` label — published, then removed by hand

Recording the re-plan disposition published `disposition-re-plan` to this PR, and `merge-gate.sh` then refused it: *"A foundational finding withdrew this change rather than patching it … close it and work the root item instead."*

That instruction is correct when a re-plan withdraws **a change the PR carries**. This PR carries none. Its entire content is CORE-043's problem statement and CORE-038's closure — so closing it would **discard the root item**, which exists only inside it.

I removed the label rather than override the gate. The disposition is still recorded in three places that are not a label:

- CORE-038's `## Outcome` section,
- CORE-043's header (`Disposition: re-plan`),
- this PR's description.

**Filed as PROC-010**, because the failure is in the mechanism, not in this PR. `record-local-review --disposition re-plan` labels the current PR unconditionally, and `record-local-review` also refuses to record a disposition when no PR resolves (PROC-007's rule) — so an author is pushed to open the PR first and then record, which is exactly the sequence that produces the mislabelling. The gate's refusal is fail-closed with a `MERGE_GATE_ACK=1` override, so the first person who hits this learns that the disposition flag is something to avoid. `git-branch.md` names that shape: *a gate that trains people to route around it has already failed.*

PROC-010 states the distinction the mechanism needs — what the PR **contains**, not what the verdict **was** — and deliberately does not choose between the three ways to implement it, because that is a judgement about the enforcement design rather than a defect with one right answer.

PR #1756, opened 2026-08-16
docs(tasks): file CORE-044 and drop CORE-042's invalid reservation

woojubb:
Documentation and backlog only — no code changes.

## What this corrects

I reserved four decisions on **CORE-043** for the owner because each "changes a published or externally visible contract". **That ground does not exist in this repository, and I was misapplying the rule.** No stable version has been released, the current beta is not distributed, and the rules forbid keeping legacy or compatibility code — `code-quality.md:50` says it in the paragraph I had been quoting: *"(unreleased — no backward-compat constraint)"*.

`backlog-execution.md` § Agent Decision Authority reserves a contract change because of the **coordination cost with a party who cannot be updated**. Where every counterparty is in this repository, that cost is zero. The test that replaces "is this exported?" is **"is there a party who cannot be updated?"** — and inside this repo the answer is always no, so the question collapses to size, coherence, and whether the change is the correct design. Those are judgements to defend, not decisions to reserve.

**Four instances this session**, now all withdrawn or already corrected by other routes:

| Where | The invalid framing |
| --- | --- |
| CORE-043 | four decisions "reserved for the owner" as published-contract changes |
| CORE-042 Direction | the test-double work reserved because it "widens a published `./testing` export" |
| CORE-042 plan | "preserve `IStreamChunk` — the shape is public" (already fell to a false premise) |
| CORE-042 plan | "keep the throw — preserves today's error exactly" (already fell to evidence) |

The test-double case was not even a widening: `createScriptedProvider.chat()` ignores `options.onTextDelta`, which is a clause of the `IAIProvider` contract it claims to implement. Bringing a non-conformant double into conformance needs no permission.

## Two facts I had wrong, and what checking them found

I wrote that `/chat` was "served by an external `config.serverUrl`" with no in-repo counterparty. **False** — `apps/agent-server/src/app.ts:120` serves `/api/v1/remote/chat`, BYOK at `:147`. Checking it surfaced two defects worse than the one being filed:

### The server silently discards the agent's tools

`app.ts:121` destructures `{ provider, messages, model }`; `:138` calls `provider.chat(messages, { model })`. The client **does** send `tools` (`chat-http-methods.ts:99-104`) — the server never reads them.

**So a remote-executor agent has no tools at all, and nothing says so.** The model simply never calls one. An agent configured with tools appears to work and quietly cannot act. Same silent-drop failure class as CORE-042, different seam, and user-facing.

### There is no streaming route in this repository

The server's whole route table is `remote/health`, `remote/chat`, `byok/chat`, `/`, `remote/ws/status`, `/health`, and the playground router. Meanwhile the client posts to `${baseUrl}/stream` (`chat-http-methods.ts:174`), a second site names `/chat/stream` (`request-handler-simple.ts:47-48`), and both `app.ts:111` and `apps/agent-server/docs/SPEC.md:31` claim the chat/stream routes were inlined — only the chat route was.

A client calling an endpoint nobody serves, under two spellings, with the documentation asserting otherwise.

## Filed as CORE-044 (`critical`)

The remote provider seam implements **none** of the `IChatOptions` contract at either end: the client's `HttpClient.chat` has no options parameter at all, and the server drops what does arrive. Costs: tools gone (silent), the path uncancellable (`signal` — the class CORE-018 and RUNTIME-004 fixed elsewhere), `toolChoice` silently degraded to `'auto'` (CORE-017's directive), and `maxTokens`/`temperature`/`effort` ignored (CORE-016's options). Nothing tests any of it at the boundary — which is the mechanism, not an aside.

**Not owner-reserved either.** It is sequenced after CORE-042 for correctness and size: routing `executeChat` to the streaming surface would point it at a route that does not exist, would move *all* remote traffic there (the round path passes `onTextDelta` unconditionally), and needs the tool-call fragment assembler CORE-042 deletes.

That last finding also **strengthens** CORE-042's scope decision, which previously rested on "I cannot verify the far end": the far end demonstrably is not there.

## Verification

`pnpm harness:verify-like-ci` — **12/12 stages pass**.

### DIFF
--- a/.agents/tasks/CORE-042-the-execution-turn-is-implemented-twice.md
+++ b/.agents/tasks/CORE-042-the-execution-turn-is-implemented-twice.md
@@ -125,10 +125,21 @@ construction, so it is planned with them rather than beside them; doing this one
 CORE-032 would re-derive the same seam twice. Whether the three are executed as one work unit or as an
 ordered initiative is the first decision the spec-doc must make.
 
-**A prerequisite, not an afterthought:** give the shared test double a `chatStream`, so a streaming
-turn can be exercised from the sanctioned surface rather than from a double re-written per test file
-— the state that let this happen. This is **this item's** work, not a contained fix's: it widens a
-published `./testing` export, which `backlog-execution.md` § Agent Decision Authority reserves.
+**A prerequisite, not an afterthought:** the shared test double must be able to exercise a streaming
+turn, rather than each test file re-writing its own — the state that let this happen.
+
+> **Correction (2026-08-16).** This paragraph previously said the work was reserved because "it widens
+> a published `./testing` export, which `backlog-execution.md` § Agent Decision Authority reserves."
+> **That ground does not exist here and is withdrawn.** No stable version has been released, the beta
+> is not distributed, and the rules forbid keeping legacy or compatibility code (`code-quality.md:50`
+> — _"unreleased — no backward-compat constraint"_). That clause reserves a change for the coordination
+> cost with a party who cannot be updated; inside this repository there is none. The test to apply is
+> **"is there a party who cannot be updated?"**, not "is this exported?".
+>
+> It is also not a widening: `createScriptedProvider.chat()` currently ignores `options.onTextDelta`,
+> which is a clause of the `IAIProvider` contract it claims to implement. Bringing it into conformance
+> needs no permission. Recording the `IChatOptions` it was called with IS additive, and is ordinary
+> agent-authority work.
 
 ## Test Plan
 
--- a/.agents/tasks/CORE-044-remote-executor-drops-every-per-call-option.md
+++ b/.agents/tasks/CORE-044-remote-executor-drops-every-per-call-option.md
@@ -1,0 +1,143 @@
+---
+title: 'CORE-044: the remote provider seam implements none of the IChatOptions contract at either end — the client sends no options, the server drops the `tools` the client does send so a remote agent silently has NO TOOLS, and the streaming client posts to a route the server does not serve, spelled two different ways'
+status: todo
+created: 2026-08-16
+priority: critical
+urgency: soon
+area: packages/agent-remote-client, apps/agent-server
+depends_on: [CORE-042]
+---
+
+# CORE-044: the remote seam implements none of the contract it declares
+
+Found while planning [CORE-042](CORE-042-the-execution-turn-is-implemented-twice.md), which made the
+gap load-bearing: once the streaming entry becomes an adapter over the round path, a remote provider
+that ignores `signal` is a turn that cannot be cancelled.
+
+## Problem
+
+`IChatOptions` carries `signal`, `onTextDelta`, `toolChoice`, `maxTokens`, `temperature` and `effort`.
+The remote provider seam implements **none** of it, at **both** ends — and the damage is worse than
+options.
+
+### The server silently discards the agent's tools
+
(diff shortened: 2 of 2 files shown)


woojubb:
## Rebased — the CORE-043 half landed independently, so this PR is now narrower

While this was open, **#1755** landed the same correction for CORE-043 and PROV-006, with better evidence than mine: `@robota-sdk/agent-core` has **71 versions on npm and zero non-prerelease**, and the `latest` dist-tag itself resolves to `3.0.0-beta.79`. Two sessions reached the same conclusion independently, which is worth noting on its own — the invalid "published contract" reservation was not a one-off slip.

I rebased and took **#1755's version of CORE-043 verbatim**. This branch now touches two files, neither of which #1755 touched:

- **`CORE-044` (new, `critical`)** — the remote provider seam.
- **`CORE-042`** — dropping the same invalid reservation from its Direction (the test-double work, reserved as "widens a published `./testing` export"). Third instance of the ground.

The findings that justify CORE-044's severity are unchanged and stand on their own:

- **The server silently discards the agent's tools.** `apps/agent-server/src/app.ts:138` calls `provider.chat(messages, { model })` while the client *does* send `tools`. A remote-executor agent has no tools at all, and nothing says so.
- **There is no streaming route in this repository.** The client posts to `${baseUrl}/stream`, a second site names `/chat/stream`, and both `app.ts:111` and the server's SPEC claim the routes were inlined — only the chat route was.

## Verification after the rebase

`pnpm harness:verify-like-ci` — **12/12**.

One note for anyone reading the intermediate state: the first post-rebase run failed `typecheck` on `IScheduleEditPatch` not being exported from `@robota-sdk/agent-executor`. That was **stale `dist/` from the newly landed ARCH-025 (#1754)**, not this diff — this branch touches only `.agents/tasks/`. Rebuilding `agent-executor` cleared it. Recording it because a docs-only PR going red on a type error is exactly the kind of thing that gets waved through as noise, and it was worth the two minutes to confirm which of the two it was.

woojubb:
## Merging with `MERGE_GATE_ACK=1`, and why — plus a gate gap worth filing

`merge-gate.sh` refuses this PR:

```
[merge-gate] Blocked: no comment on #1756 is from the reviewer this gate looks for.
[merge-gate]   looked for: ^github-actions(\[bot\])?$   comments are from: woojubb
```

The reviewer login has not changed. **The automation deliberately produced no review for this PR**: `classify review applicability` passed and `Claude review` resolved to `skipping`, because this diff touches only `.agents/tasks/*.md`. That matches `git-branch.md`'s own Pre-Merge Code-Review Gate scope — *"A documentation/spec/backlog-only PR (markdown/JSON config only, no code diff) is exempt — running `/code-review` on it yields no code findings."*

So the gate is demanding an artifact its own applicability classifier withheld by design. That is not the reviewer being late and it is not a login change; the two mechanisms disagree about the same PR.

Using the documented inline exception rather than routing around it, and recording the reason here so the override is a visible choice rather than a silent one. A local review **was** run and recorded (`harness:review:record`, 0 findings), and `pnpm harness:verify-like-ci` is 12/12.

**Filing this as a harness item**, because it is the second instance this session of a gate blocking on a state its own machinery created — PROC-010 was the first (recording a re-plan disposition labels the PR that *files* the root item, then refuses to merge it and instructs you to close it). Both teach the same lesson to whoever hits them next: reach for `MERGE_GATE_ACK=1`. `git-branch.md` names that shape exactly — *"a gate that trains people to route around it has already failed"* — and the fix belongs in the gate, not in the habit.

PR #1759, opened 2026-08-16
docs(tasks): file PROC-011 — the merge gate blocks a docs-only PR

woojubb:
Backlog only — no code changes.

## What happened

PR #1756 was a documentation-only change. `merge-gate.sh` refused it:

```
[merge-gate] Blocked: no comment on #1756 is from the reviewer this gate looks for.
[merge-gate]   looked for: ^github-actions(\[bot\])?$   comments are from: woojubb
[merge-gate] If the reviewer's login changed, fix REVIEWER_RE — do not route around it.
```

The login had not changed. **The reviewer never ran** — `classify review applicability` passed and `Claude review` resolved to `skipping`, because the diff carried no code. That is exactly what `git-branch.md`'s own Pre-Merge Code-Review Gate scope says should happen:

> **Scope:** required for any PR that changes code … A documentation/spec/backlog-only PR (markdown/JSON config only, no code diff) is exempt.

So the rule exempts the PR, the CI classifier exempts the PR, and the hook refuses it anyway. The diagnostic then proposes a cause it has not established, which is what sent the investigation down the wrong path first.

## Why this is worth a task rather than a shrug

The only route through is `MERGE_GATE_ACK=1`. Every documentation-only PR therefore ends in the override, and whoever learns that carries it to a PR where the gate **is** load-bearing. `git-branch.md` names the shape:

> a gate that trains people to route around it has already failed

and, in the same section:

> If a check is wrong, unrunnable, or fires on correct work, **the check is what changes**.

**Second instance today**, which is what makes it a pattern rather than an inconvenience. PROC-010 is the first: recording a re-plan disposition publishes a label to the PR that *files* the root item, after which this same hook refuses it and instructs the author to close it — which would discard the root item. Both end in `MERGE_GATE_ACK=1`; both are the gate blocking on a state the harness itself produced. The two items are cross-linked, since a fix that teaches the hook *why* an artifact is absent would likely address both.

## Direction — deliberately not chosen

Two shapes, and picking between them is a judgement about the enforcement design rather than a defect with one right answer: have the hook read the applicability signal CI already computes, or have the review workflow always post a verdict (`ACTIONABLE FINDINGS: 0`, body "not applicable — docs-only") so the artifact always exists and the hook stays simple.

One thing is not optional either way: the diagnostic must stop asserting a cause it has not established. "No verdict found" and "the reviewer's login changed" are different claims.

## Verification

`pnpm harness:verify-like-ci` — 12/12. A local review was run and recorded (0 findings).

**This PR will hit the same gate it describes.** I will merge it with the documented inline override and say so in a comment, rather than routing around it silently — which is the behaviour the item is about.

### DIFF
--- a/.agents/tasks/PROC-010-re-plan-disposition-blocks-the-filing-that-carries-it.md
+++ b/.agents/tasks/PROC-010-re-plan-disposition-blocks-the-filing-that-carries-it.md
@@ -50,6 +50,13 @@ Note also that `record-local-review` already refuses to record a disposition whe
 (PROC-007's rule), so the author is pushed to open a PR _first_ and then record — which is exactly the
 sequence that produces the mislabelling.
 
+**A second instance landed the same day.**
+[PROC-011](PROC-011-merge-gate-demands-a-review-its-own-classifier-withheld.md): the merge gate
+demands a reviewer verdict on a documentation-only PR that the review automation deliberately skips,
+so that PR also ends in `MERGE_GATE_ACK=1`. Two different mechanisms, one shape — the gate refusing a
+PR because of a state the harness itself produced. Whatever fixes one should be checked against the
+other.
+
 ## Direction
 
 The distinction the label needs to carry is **what the PR contains**, not what the verdict was. A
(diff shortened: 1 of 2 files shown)


woojubb:
Merging with `MERGE_GATE_ACK=1`, on the same gate this item describes — recorded here rather than done silently, because doing it silently is the habit the item is about.

`classify review applicability` passed and `Claude review` resolved to `skipping`: no code diff, so no verdict exists for the gate to read. A local review was run and recorded (0 findings) and `pnpm harness:verify-like-ci` is 12/12.

This is the third override in the session under the same hook — PR #1751 (PROC-010's label), PR #1756 (PROC-011's missing verdict), and now this one. Three is enough to say the pattern is in the mechanism rather than in the PRs.

Issue #1810, opened 2026-08-16
Add same-environment identity authentication for local agent-cli peers


woojubb:
Parent: #1807

## Purpose

Define and implement the local peer-admission security boundary that allows two `agent-cli` sessions on the same computer to communicate when they share the configured authentication certificate/identity and trusted environment.

This issue owns authentication and admission only. The session messaging behavior is tracked separately under the other child issue for #1807.

## Scope

- Define the identity and environment-binding contract for a local agent-cli peer.
- Integrate the existing WebRTC pairing/data-channel gate with certificate/identity verification before exposing any session or message frames.
- Bind admission to the intended local session/connection and reject unknown, mismatched, expired, replayed, or malformed authentication material.
- Fail closed before session content is exposed; define timeout, cancellation, teardown, and re-authentication behavior.
- Expose a narrow authenticated-peer result/port consumed by the messaging feature without coupling security logic to TUI components.
- Document the local trust assumption and its limits, including what happens when the certificate is shared beyond the intended environment.

## Out of scope

- Message ordering, conversation history, agent response behavior, or TUI conversation UX; see the functional child issue.
- Cross-computer same-user authentication and authorization; see the security child issue for #1808.

## Acceptance criteria

- A local peer with the expected certificate/identity and environment binding is admitted and receives an authenticated-peer result.
- A peer with mismatched, missing, expired, replayed, or malformed identity material is rejected before any session content is exposed.
- Authentication failures and connection teardown leave no usable session/message handler or leaked pending state.
- Tests cover valid admission, mismatch, replay/tamper, timeout, concurrent attempts, and cleanup using deterministic WebRTC peers.
- The security contract and threat assumptions are documented in the owning pairing/transport specification.

## Contract ownership and boundaries

- **Cryptographic protocol SSOT:** `@robota-sdk/agent-remote-pairing` should own the proof and result contract, or a new zero-dependency isomorphic security leaf should be introduced if “same local environment” is broader than device pairing.
- **Admission consumer:** `@robota-sdk/agent-transport-webrtc` should own the fail-closed channel gate and consume the authentication result. WebRTC must not implement cryptographic policy itself.
- **Feature consumer:** #1809 should receive a narrow typed authenticated-peer/admission result, not a generic boolean. The result should carry the verified peer binding and an explicit trust level.
- **CLI role:** `agent-cli` may collect local consent/configuration and compose dependencies, but must not own the crypto types or proof algorithm.

[...]

PR #1841, opened 2026-08-17
feat(agent-transport-webrtc): bind the rendezvous nonce to the admitted channel (#1810)

woojubb:
SEC-010 **TC-08** — the last piece of #1810's proof chain. The guarded-directory check and the grant ledger both landed already; this joins them to the channel the session actually runs on.

## Why neither half is enough alone

| Mechanism | Binds | Silent about |
| --- | --- | --- |
| Pairing handshake | the **channel** — it provably terminates at the peer that knew the secret, DTLS fingerprints covered | where that peer runs |
| Guarded rendezvous | the **environment** — only this user on this machine could have reached a 0700 directory they own | which channel that peer later opens |

Without a binding, a peer could pass the kernel's check at the socket and hand the channel to somebody else — and the admission would still read `same-user-same-host`. Presenting the nonce **over the already-bound channel** joins them: the channel is authenticated by then so nobody else can inject the frame, and the nonce is single-use so nobody can re-present one they observed.

## Both edges of the step are load-bearing

- **After the handshake**, because before it the channel is not yet authenticated — the nonce would be handed to an unproven counterpart, weakening the very thing it exists to strengthen.
- **Before the session is exposed**, because "fail closed before content" is SEC-010's first failure rule. A peer that never presents a nonce would otherwise already be talking to the session, and refusing it afterwards is not refusing it.

## Every non-admitted path closes the channel

A refused nonce, a frame that is not a proof frame, and a ledger that *throws*. "Not reached" is not "allowed" — and a gate that merely **ignored** a bad frame would park with the channel open, which is a hang: fail-open wearing a stall's clothes.

**Red-proofed**: disabling the local-proof step fails six of the thirteen new tests and no others.

## This package decides nothing

#1810 is explicit that WebRTC must not implement cryptographic policy. Single-use, expiry and revocation stay in the grant ledger and arrive as an injected `redeem`, which also keeps `node:fs` out of this package and keeps the ledger testable without a data channel. The consumer is notified on **refusals as well as admissions** — "no local peer connected" and "a local peer was refused" call for different operator responses — and the result carries an explicit trust level, not a boolean.

Absent `localPeer`, behaviour is exactly as before: a remote peer has no rendezvous to have reached, and demanding one unconditionally would refuse every legitimate remote session.

## Three splits, because the files were at the wall

`pairing-gate.ts` was 293 lines against a 300 limit; `webrtc-transport.ts` was 299. The anti-monolith rule says split rather than extend, so the controller construction, the session attachment, and the options interface each moved to a file that owns their subject.

**All three are behaviour-preserving, and that is verified rather than asserted**: the 43 pre-existing webrtc tests pass unchanged after each step.

## Also caught in self-review

The barrel initially exported the judge and the frame predicate. `orphan-exports` flagged the predicate as referenced nowhere — and on inspection, exporting either would offer a caller a way to check the frame shape and act on it *without going through the judge*, which is how an admission decision ends up made in two places. The predicate is module-private now.

I also confirmed the feature is **not inert**: `localPeer` is threaded through `IWebRtcTransportOptions` to the gate, so a composition root can actually configure it. Without that the whole step would be unreachable from outside the package.

## Verification

- 56 tests in the package (43 pre-existing + 13 new), all green
- `pnpm harness:scan` — 123 passed, 2 skipped
- `pnpm typecheck` clean; `verify-like-ci.mjs --only format-check` green
- SPEC `## Boundaries` documents what the gate owns, what it does not, and why both edges matter

https://claude.ai/code/session_01NQQwC79KAtQwUjD4QoTNPD

### DIFF
--- a/packages/agent-transport-webrtc/docs/SPEC.md
+++ b/packages/agent-transport-webrtc/docs/SPEC.md
@@ -36,6 +36,30 @@ protocol is shared, not duplicated.
   written-reason requirement lives in `resolveAdmission` in `@robota-sdk/agent-transport-protocol`, so the sibling
   transports cannot drift apart on what counts as an answer.
 
+- **Owns the local-peer channel GATE, not the local-peer policy (SEC-010, #1810).** When `localPeer` is
+  configured, an accepted handshake is not enough: the peer must also present the nonce it was issued at the
+  guarded rendezvous, and only then is the session exposed. This package holds the state machine and the
+  frame; single-use, expiry and revocation belong to the grant ledger in
+  `@robota-sdk/agent-remote-pairing/local`, which is injected as a `redeem` port. That keeps the
+  cryptographic/OS policy out of the transport (the issue is explicit about this) and keeps `node:fs` out of
+  this package.
+
+  **Why both edges of the step are load-bearing.** The handshake binds the CHANNEL — after it, the channel
+  provably terminates at the peer that knew the secret — and says nothing about where that peer runs. The
+  guarded rendezvous binds the ENVIRONMENT and says nothing about which channel that peer later opens.
+  Presenting the nonce OVER the already-bound channel joins them. Earlier than that and the nonce would be
+  handed to an unproven counterpart; later than that and a peer would already be talking to the session, where
+  refusing it is not refusing it.
+
+  Every non-admitted path closes the channel: a refused nonce, a frame that is not a proof frame, and a ledger
+  that throws. "Not reached" is not "allowed", and a gate that merely ignored a bad frame would park with the
+  channel open — a hang, which is fail-open wearing a stall's clothes. The consumer is notified on refusals as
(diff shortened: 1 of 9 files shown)


woojubb:
Pushed `34c84f6fb` — a self-review finding on this PR, disclosed rather than folded in quietly.

The three file splits left the SPEC's **Type Ownership** and **Cross-Package Port Consumers** tables naming locations that no longer hold what they name. A spec that points at the wrong file is worse than one that says nothing: a reader who follows it and finds nothing concludes the spec is stale in general and stops trusting the rows that *are* right.

Corrected to where each type actually lives, with the new modules documented rather than left out. The port-consumer rows are split too — `createWsHandler` now has consumers in two files, and the two pairing entry points are no longer consumed from the same place, so one row claiming both would have been accurate only by being vague.

No code changed; scans and format-check re-run green.

PR #1888, opened 2026-08-18
feat(harness): the same tree is not scanned twice (HARNESS-109)

woojubb:
Closes HARNESS-109. #1873 landed the measurable half (an untracked file the agent harness writes made
`isCleanTree()` false, so no receipt was ever written and every push re-ran the full gate). This is
the mechanism.

## The same tree is not scanned twice

`scripts/harness/run-all-scans.mjs` is the most repeated command in a session — `pre-push` runs it
through `CI_SCANS_JOB_MIRROR` on every push, CI runs it again, and an agent runs it by hand for a
signal. Nothing connected those runs, so scanning one tree three times was indistinguishable, to the
harness, from scanning three trees.

`scripts/harness/scan-receipt.mjs` gives the suite what `scripts/harness/verification-receipt.mjs`
already gave `verify-like-ci`. Identity is `headTree` plus the toolchain: the tree hash is the content
of every tracked file, so any change to a scan, a baseline, or a source file it reads invalidates it.

Measured end to end on the CI mirror set (`--skip dist --skip build-contracts` — the set `pre-push`
repeats):

| Run | Result |
| --- | --- |
| first, clean tree | 123 passed / 3 skipped, receipt written, **14s** |
| second, unchanged tree | `126 scans not re-run: identical tree scanned at …`, **0s** |
| third, one tracked file touched | `scan receipt not reused: working tree is not clean:  M AGENTS.md` |

## What a tree hash cannot speak for

`dist` and `build-contracts` compare `dist/` against `src/`, and `dist/` is ignored — two runs under
one tree hash can legitimately disagree. Both are named in `TREE_EXTERNAL_SCANS`, and a set containing
either is ineligible in **both** directions: it neither reuses a receipt nor writes one. Asserted even
in the case where identity, receipt and cleanliness all agree, because that is the case where a
mechanism built for convenience would say yes.

## The direction it fails in

A missing, malformed, failed, or wrong-identity receipt re-runs the suite. There is no path where an
unreadable receipt becomes a pass. The decision is one pure function so every refusing branch is a
test rather than a claim — 14 tests, and the refusing branches outnumber the reusing ones.

A miss also reports **why**: a reuse mechanism that silently never fires cannot be told from one that
is not wired at all, which is the defect that made #1873 necessary in the first place.

## The ignore rule is now a class

#1873 named one filename — exactly how the lessons-churn instance was closed, and naming files one at
a time is what let a second instance happen. The rule is now the tool's own convention for a per-clone
file (`*.local.json` under its config directory), with the shared settings file still tracked. A test
plants both in a temporary repository and asserts the exemption stays narrow.

## Not in scope, deliberately

A diff-scoped **subset** of the suite. `scripts/harness/pre-push.mjs` records the decision that the
whole suite runs because it is cheap relative to CI. This removes the duplicate run; it does not
reopen that trade-off.

## Verification

`pnpm harness:scan -- --skip dist --skip build-contracts`: **124 passed / 2 skipped**. One finding was
found and fixed before the push — `unearned-done-claims` caught acceptance boxes that said "proved by
a test" without citing it; each now cites its test or its measurement.

Process note: this was implemented on the owner's instruction to continue, without a gate-pipeline
spec document. Stated as a fact, not a verdict.


woojubb:
## 지적 반영 완료 (`23155f068`)

맞는 지적입니다. 리셉트 히트 시 조기 반환했기 때문에, `--write-adoption-baseline`이 요청한 **재동결이라는 부수효과가 조용히 사라지고** 출력에는 "N scans not re-run"만 남았습니다. 호출자가 부수효과를 요청했는데 캐시가 그걸 삼키면 그건 절약이 아니라 결함입니다.

**수정**: `--write-adoption-baseline`은 다른 어떤 검사보다 **먼저** 재사용을 거부하고, 어떤 플래그 때문인지 밝힙니다.

유효한 리셉트가 존재하는 상태에서 실측:

```
▶ scan receipt not reused: --write-adoption-baseline re-freezes from an observed pass
✎ re-froze examined-size adoption: 78 scan(s) in scripts/harness/examined-adoption-baseline.json.
124 scans passed, 2 skipped (77 declared what they examined)
```

플래그를 호출부에서 검사하지 않고 **순수 판정 함수 안으로** 넣었습니다. 그래야 이 거부가 다른 모든 거부 분기와 같이 테스트가 됩니다 — 신원·리셉트·클린 상태가 완벽히 일치하는 경우에도 이 플래그 아래에서는 재사용하지 않는다는 케이스 포함. 16/16 통과.

Issue #1863, opened 2026-08-18
agent-cli: local peer discovery and the command that addresses another live session (follow-up to #1809)


woojubb:
Split out of #1809 so that issue can close on the work it specified. **The transport-neutral contract, the wire semantics, and the session ingress are all built and merged, with two-session functional coverage against real sessions.** What remains is the shell — which #1809 itself scopes as `agent-cli` owning "discovery, commands, lifecycle, and TUI projection only".

## What already exists, and where

| Piece | Package | Export |
| --- | --- | --- |
| Peer-message contract, trust levels, type predicates | `agent-interface-transport` | `IPeerMessage`, `IPeerAdmission`, `TPeerTrust`, `isSameEnvironmentPeer` |
| Ordering, duplicates, acks, retries, disconnect | `agent-transport-protocol` | `admitPeerMessage`, `acknowledgePeerMessage`, `forgetPeerOrigin` |
| Session ingress — admission fail-close, origin preserved, ack from turn settlement | `agent-framework` | `PeerMessageIngress` |
| A peer turn the runtime can branch on | `agent-interface-transport` | `TTurnSource` gained `'peer'` |

Verified at the time of writing: **no file under `packages/agent-cli/src` references `PeerMessageIngress`.**

## What this issue is

1. **Local peer discovery** — how one session learns another is live and addressable. The guarded rendezvous directory from #1862 is the natural registry; a live session's entry in it is what makes it discoverable, and removing the entry is what makes it not.
2. **The command** that addresses another session and sends text.
3. **TUI projection** — render the peer's origin so an operator can see a message came from another session, not from themselves.
4. **CLI documentation** of the message flow and its concurrency behaviour.

## Constraints inherited from #1809 — do not re-decide these

- **`TDriverId` is display attribution and never an authorization input.** The driver id for a peer turn is derived from the peer's session id rather than taken from the peer's own `IPeerOrigin.driverId`: that field is peer-supplied, and a name the transcript's reader trusts must not be chosen by the party being named.
- **A peer turn must carry the peer's driver id.** Submitting one without it is an error, not a default — falling through to the owner would put another session's message in the transcript under the operator's name.
- **Do not add a second queue.** A message arriving mid-turn is handled by the session's existing pending queue. PEER-002's first draft reimplemented that and immediately grew two bugs the original never had.

## Related, already filed

[...]

Issue #1809, opened 2026-08-16
Implement local agent-cli session-to-session messaging


woojubb:
Parent: #1807

## Purpose

Implement the functional session-to-session conversation capability for two `agent-cli` instances running in the same local environment.

This issue owns message exchange and session behavior. Authentication and peer admission are provided by the separate child issue for #1807 and are not reimplemented here.

## Scope

- Define the transport-neutral peer-message contract and sender/session metadata.
- Add the local peer discovery/rendezvous and `agent-cli` command/action needed to address another live session.
- Establish the authenticated peer channel through an injected admission/authentication port.
- Support ordered bidirectional text messages with explicit delivery, acknowledgement, duplicate, disconnect, retry, and shutdown behavior.
- Deliver incoming peer messages into the receiving session runtime as typed external input that the agent can read and answer, while rendering the origin clearly in the TUI.
- Keep the feature reusable below the TUI and independent of a specific provider.
- Document concurrency semantics when either session is already executing a turn.

## Out of scope

- Designing or implementing certificate validation, same-environment identity proof, or cross-computer user authentication; those belong to the security child issues.
- Transferring session ownership or work state to another computer; see #1808.

## Acceptance criteria

- Two live local `agent-cli` sessions can exchange ordered messages in both directions through the authenticated-peer port.
- A received peer message is added to the receiving runtime context with explicit peer origin and can trigger an agent response.
- Delivery acknowledgements, duplicates, disconnects, retries, and shutdown produce deterministic, documented outcomes.
- The transport/session contract is covered independently from the TUI, and framework-level functional coverage exercises a two-session conversation with deterministic providers.
- The owning package specs and CLI documentation describe the new message flow and its concurrency behavior.

## Contract ownership and boundaries

- **Semantic SSOT:** `@robota-sdk/agent-interface-transport) should own the new peer-message/session-ingress contract (for example, a typed `IPeerMessage` plus a session peer-messaging port). The contract should carry stable message identity, peer/session origin, ordering metadata, and delivery state.
- **Wire SSOT:** `@robota-sdk/agent-transport-protocol) should own the serialized peer-message frame, acknowledgement, sequence, retry, and duplicate semantics.
- **Runtime implementor:** `@robota-sdk/agent-framework) should implement the session ingress and history/runtime projection. It must define how a peer message becomes agent-readable input without losing peer origin.
[...]

PR #1897, opened 2026-08-19
feat(agent-cli): announce this session, and show the operator the others (#1863)

woojubb:
## What this changes

The local peer stack landed in issue #1809 (rendezvous directory, announce/withdraw, discovery) had **zero consumers**: nothing announced this session, and nothing read the directory. This is stage 2 of issue #1863 — the wiring, plus the first command that shows an operator what it found.

- `agent-framework` gains a `localPeers` host-adapter port (`ICommandLocalPeersAdapter`, `ILocalPeerSummary`) alongside the existing `remoteControl` port.
- `agent-cli` announces this session into the guarded rendezvous directory at startup and withdraws on `exit`; the same module projects discovery into the summary shape the port asks for.
- `agent-command` gains `/peers`, which lists the live sessions, marks this session, and says plainly when discovery is unavailable rather than failing silently.

## Notes on shape

- `cli.ts` is at its frozen `file-size` baseline, so `attachCommandHostAdapters` absorbs the remote-control assignment too — the call site is a **net zero** line change. The baseline was not raised.
- `buildRemoteControlHostAdapter`, `buildLocalPeersHostAdapter` and `attachLocalPeerDiscovery` are module-private; `orphan-exports` and `spec-public-surface` both refuse them as exports, and the correct answer was to stop exporting rather than to document them.
- `robota-assembly-equivalence` asserts a command-module **set**; `agent-command-peers` is recorded in the baseline with the reason, because that test exists to make an unexplained gain or loss visible.

## Verification

`pnpm harness:pre-push` exits 0 on this branch.

Closes nothing on its own — issue #1863 stays open for stage 3 (`/peers send` over the local channel).


woojubb:
## Review — feat(agent-cli): announce this session, and show the operator the others (#1863)

Stage 2 of issue #1863: wiring a stack that was fully built and had **zero consumers**, plus the first command that shows an operator what it found.

### What I checked

- **The premise is real, not assumed.** Before this branch, nothing announced this session and nothing read the rendezvous directory — the local peer stack from issue #1809/#1862 was declared and never wired. That is the class of defect where build, typecheck, lint and tests all stay green for weeks, which is exactly what `orphan-exports` exists to catch and did not, because the symbols were referenced by their own tests.
- **The port mirrors the one beside it.** `localPeers` sits alongside the existing `remoteControl` adapter on `ICommandHostAdapters` with the same optionality, so a host that cannot offer discovery says so by absence rather than by a stub that lies.
- **Failure is visible, not silent.** `/peers` distinguishes three states an operator can act on: discovery unavailable, no other live session, and a list. Collapsing the first two would tell someone their peer is gone when the feature simply is not present.
- **Withdrawal is wired to process exit**, so a session that ends stops being addressable — which is the half of discovery that makes the other half trustworthy.
- **Nothing was exported that nothing uses.** `buildRemoteControlHostAdapter`, `buildLocalPeersHostAdapter` and `attachLocalPeerDiscovery` are module-private. `orphan-exports` and `spec-public-surface` both refused them as exports, and the right answer was to stop exporting rather than to add SPEC rows documenting a surface nobody consumes.
- **`cli.ts` is a net ZERO line change.** It sits at its frozen `file-size` baseline, and `attachCommandHostAdapters` absorbs the remote-control assignment so the call site neither grows nor needs the baseline raised. I did not use `--write-baseline`.
- **The reporter is passed as an object, not an unbound method.** `terminal.writeError` detached from its receiver would fail at the one moment it is needed — when reporting that discovery could not start.
- **The assembly baseline gained an entry with its reason.** `robota-assembly-equivalence` asserts a command-module *set*; `agent-command-peers` is recorded with a comment saying the baseline exists to make an unexplained gain or loss visible. That test caught this branch during pre-push after I had run only the three touched packages' suites.

### The `quality` failure, and why it was not this branch

CI reported `quality` red on this branch's head. It was **develop's** defect, not this change: ARCH-035 made `createSession` async and left a caller in `agent-command`'s scenario verifier unawaited. CI tests a PR *merged into its base*, so a branch cut before that change shows the base's failure on its own head sha. Reproducing on a detached clean `origin/develop` is what established this; the fix went in as PR #1901, and rebasing here turned `quality` green.

### Scope

`/peers send` is not wired — that is issue #1863's remaining stage, now that PR #1906 has landed the carrier. This does not close the issue and does not claim to.

### Verification

`pnpm harness:pre-push` exits 0. Every CI check passes, including `quality`, `scans`, `windows-shell` and `tui-e2e`.

REVIEWED BASE: fb819383ba3bd1c2a1f7202d03d338ae7237cae1
REVIEWED HEAD: 0aa42cb63973e365afcba61d6ba65cd59988af2d
ACTIONABLE FINDINGS: 0

PR #1892, opened 2026-08-19
fix(docs): qualify three bare issue references that turned develop red

woojubb:
`develop` is red. `scan-reference-kind-qualified` (PR #1890) went red the moment two records landed
carrying bare `#N` references it had never seen:

```
- [unfrozen] .agents/tasks/completed/ARCH-035-…md: 2 unqualified reference(s)
- [unfrozen] .agents/tasks/completed/HARNESS-109-…md: 1 unqualified reference(s)
exit=1
```

Three edits, one word each — `#1854` → `issue #1854` (twice), `#1873` → `PR #1873`. Each is exactly
the issue/pull-request distinction a reader cannot make from the number alone, which is what the scan
exists to prevent, so the corrections are the ones the rule asks for rather than a way to quiet it.

## Why one of them escaped

The HARNESS-109 reference had already been qualified on its own branch. PR #1888 merged the
**pre-rebase** head, so the fix was not in what landed. Recorded because the same thing will happen to
the next branch that is rebased and then merged from an older head.

## Verification

- `node scripts/harness/scan-reference-kind-qualified.mjs` — exit **1** before, exit **0** after
- `pnpm harness:scan -- --skip dist --skip build-contracts` — 124 passed / 3 skipped

PR #1900, opened 2026-08-19
docs(memory): mirror the reference-kind directive into the in-repo corpus

woojubb:
## What this changes

`memory-mirroring.md` makes `.agents/memory/` the shared corpus — anything written to session memory has to be readable from a fresh clone. The reference-kind directive (rule landed in PR #1890, tree scan in PR #1892) was mirrored to host memory but not here.

Documentation only: one memory file plus its index line. No rule text is restated — the rule stays owned by `.agents/rules/naming-style.md` § "Reference Kind Is Named".

## What the file records that the rule cannot

- **The count that decided the design.** 2,500 `#N` occurrences across 443 tracked markdown files, 552 already qualified. A flat gate would have been red on arrival, and a check that is red on arrival gets suppressed rather than obeyed — hence a per-file ratchet. The file also reconciles that raw count against the scan's smaller frozen baseline (1,523 across 294 files), because the two numbers read as a contradiction otherwise.
- **The generalisation behind the two enforcement shapes.** A check that judges the artefact being **written** is green on arrival by construction and needs no ratchet; a check that judges the whole tree does. When a rule has both surfaces, they need different shapes.
- **The `Closes #N` exemption**, and why it is load-bearing rather than a convenience — the INFRA-104 promotion machinery parses that exact shape.
- **Three measured traps**: `$` under the `m` flag closes a fence on its own first line; a ratchet count that fell to zero reads as a deleted file; two custom commitlint rules must be ONE plugin object or commitlint refuses the whole config.

## Verification

`pnpm harness:pre-push` exits 0 on this branch.


woojubb:
## Review — docs(memory): mirror the reference-kind directive into the in-repo corpus

Documentation only: one new memory file (62 lines) and one index line in `.agents/memory/MEMORY.md`. No rule text, no code, no harness wiring.

### What I checked

- **It does not restate the rule.** `memory-mirroring.md` makes the in-repo corpus the shared one; the risk with a mirror is that it becomes a second owner of the rule and then drifts. This file explicitly points at `naming-style.md` § "Reference Kind Is Named" as the owner and records only what the rule cannot carry — the measurement, the design decision it forced, and the traps.
- **Both relative links resolve.** `../rules/naming-style.md` and `../tasks/INFRA-104-promotion-carries-closing-keywords-to-main.md` are present in the tree. `resolving-claims` is green on this branch, which is the mechanical form of the same check.
- **The two counts reconcile.** The file cites 2,500 references across 443 files (the raw pre-implementation count) while the scan's frozen baseline is 1,523 across 294 files. Those read as a contradiction on their face; the branch adds a paragraph naming the exclusions (fenced blocks, link targets, `CHANGELOG.md`, already-qualified references) that account for the gap, so a future reader does not conclude the ratchet drifted.
- **The claims are the ones that were actually measured**, not a retelling: the `$`-under-`/m` fence bug, the ratchet count that fell to zero reading as a deleted file, and the two-commitlint-plugin-entries failure were each hit and fixed in the work this mirrors.

### What I deliberately did not flag

The raw count is a snapshot and will drift as the tree grows — it is 2,968 across 505 files today. That is not a defect: the file states it as the count taken *before anything was written*, which is what makes the design decision legible. Freezing a live number would be the mistake.

### Verification

`pnpm harness:pre-push` exits 0 on this branch. CI is green with no failing or pending checks.

REVIEWED HEAD: aafeb4f2879ecdb7d3803d9f1cb8a8fb4c18ac2e
ACTIONABLE FINDINGS: 0

PR #1901, opened 2026-08-19
fix(agent-command): await the createSession the scenario verifier calls (INFRA-108)

woojubb:
## develop is red, and this is why

`pnpm scenario:verify` in `packages/agent-command` fails on a clean detached `origin/develop`. Every branch rebased onto develop therefore inherits a red `quality` check for a defect it did not introduce.

PR #1885 (ARCH-035) changed `createSession` from synchronous to `async`. One caller was not updated: `readSystemMessage` in the semantic-command-role scenario helpers read `.session` off the unawaited promise, so every session it tracked was `undefined`.

## The symptom named the wrong place

| where it was reported | what it said |
| --- | --- |
| `verify-semantic-command-roles.ts:198` | `Cannot read properties of undefined (reading 'shutdown')` |
| the actual failure, two frames earlier | `undefined` has no `getSystemMessage` |

The reported frame is the `finally` cleanup loop. It iterates the array the `try` block pushed the `undefined` into, and its TypeError **replaces** the original one — so the message points at a cleanup step that was never the problem. The masking had to be removed before the cause was visible; both frames are kept in the task record so the next reader does not repeat that step.

## Attribution

Reproduced on a detached clean `origin/develop` **before** anything was changed. That step is what corrected a wrong first reading: CI tests a pull request *merged into its base*, so a branch cut before ARCH-035 shows the base's defect on its own head sha and reads as the culprit. I initially blamed PEER-004 for exactly this reason.

## Red-proofed, one at a time

| await removed | what the scenario reported |
| --- | --- |
| `await createSession(...)` in the helper | `Cannot read properties of undefined (reading 'shutdown')` — the original symptom, at the original frame |
| one `await readSystemMessage(...)` call site | `coincidentalPrompt.includes is not a function` |

Both were reverted before committing.

## The gap underneath — filed separately, not fixed here

`packages/*/examples/` sits outside every package's tsconfig `include`, and the root `examples:typecheck` script filters `./examples/**` — the workspace root only, despite the CI job's name. **Nine packages** carry such a directory, so a breaking signature change to a published function can land green.

Closing that is a nine-package change whose blast radius has to be measured against a full declaration build first, and develop is red now. Filed as its own issue.

I also deliberately did **not** null-guard the cleanup loop that masked the error. That is a second defect, and it belongs with the gap it hides rather than smuggled into a one-line fix.

## Verification

`pnpm harness:pre-push` exits 0. `pnpm scenario:verify` passes here and fails on develop — both halves were run.

### DIFF
--- a/.agents/tasks/INFRA-108-scenario-verifier-not-typechecked.md
+++ b/.agents/tasks/INFRA-108-scenario-verifier-not-typechecked.md
@@ -52,9 +52,9 @@ Two separable things, and this item does only the first:
 
 1. **The break** — `readSystemMessage` becomes `async` and awaits, and its four call sites await it.
    That is the whole fix; nothing else about the scenario changes.
-2. **The gap** — bringing `packages/*/examples/` under a typecheck. Filed separately, because it is a
-   nine-package change whose blast radius has to be measured against a full declaration build before
-   anyone can say what it costs, and develop is red NOW.
+2. **The gap** — bringing `packages/*/examples/` under a typecheck. Filed as issue #1902, because it
+   is a nine-package change whose blast radius has to be measured against a full declaration build
+   before anyone can say what it costs, and develop is red NOW.
 
 ## Plan
 
@@ -63,7 +63,7 @@ Two separable things, and this item does only the first:
       attributed to the right change.
 - [x] TC-03: the real frame was established by unmasking it, not by inference.
 - [x] TC-04: `pnpm harness:pre-push` is green.
-- [ ] TC-05: the typecheck gap is filed as its own issue.
+- [x] TC-05: opened issue #1902 for the typecheck gap, with the measured scope it needs.
 
 ## Test Plan
 
--- a/.agents/tasks/INFRA-108-scenario-verifier-not-typechecked.md
+++ b/.agents/tasks/INFRA-108-scenario-verifier-not-typechecked.md
@@ -1,0 +1,91 @@
+---
+title: 'INFRA-108: a breaking signature change reached develop because packages/*/examples is typechecked by nobody'
+status: in-progress
+created: 2026-08-20
+priority: high
+urgency: now
+area: packages/agent-command
+depends_on: []
+---
+
+# INFRA-108: the scenario verifier awaits `createSession`, and the gap that let it break
+
+## Objective
+
+`packages/agent-command`'s `scenario:verify` fails on a clean `origin/develop`, so every branch
+rebased onto develop inherits a red `quality` check for a defect it did not introduce.
+
+## What broke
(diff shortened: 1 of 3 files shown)


woojubb:
## Review — fix(agent-command): await the createSession the scenario verifier calls (INFRA-108)

Six lines of real change (one signature, one `await`, four call sites) plus a task record. It unblocks `develop`, which is currently red for every branch rebased onto it.

### What I checked

- **The diagnosis is established by reproduction, not inference.** The failure was reproduced on a detached clean `origin/develop` before anything changed. That step is load-bearing here: CI tests a PR *merged into its base*, so a branch cut before ARCH-035 shows the base's defect on its own head sha. I had in fact mis-attributed this to PR #1897 on the first reading, and this is what corrected it.
- **The real frame was unmasked, not guessed.** The reported error (`… reading 'shutdown'`, in the `finally` cleanup loop) is a *replacement* for the original one thrown inside the `try`. The masking was removed to see the true frame, then restored.
- **Every call site is covered.** `readSystemMessage` has exactly four callers — two in `verify-semantic-command-roles.ts`, two inside `verifyOmissionBehaviors` — and all four now await. Both enclosing functions were already `async`, so no signature above this had to change.
- **No behavioral change beyond sequencing.** The three awaits inside the `prompts` object literal serialize calls that were synchronous before, so the ordering they impose is the ordering that already existed.
- **The scope is honest.** The cleanup loop that masked the error is a second defect and is deliberately left alone, filed with issue #1902 rather than smuggled into a one-line fix. Widening this PR to "fix everything I noticed" is exactly what would make it hard to revert if it were wrong.

### Red-proofing

Both directions were run, one at a time, and reverted before committing:

| await removed | reported |
| --- | --- |
| `await createSession(...)` in the helper | `Cannot read properties of undefined (reading 'shutdown')` — the original symptom, at the original frame |
| one call-site await | `coincidentalPrompt.includes is not a function` |

The first reproducing the exact original symptom is what establishes the diagnosis rather than merely being consistent with it.

### The gap it does not close

`packages/*/examples/` is outside every package's tsconfig `include`, and the root `examples:typecheck` filters `./examples/**` — the workspace root only, despite the CI job's name. Nine packages carry such a directory. A typechecker reading this file would have refused the unawaited call outright. Filed as issue #1902 with the measurement work it needs, and correctly not attempted here while develop is red.

### Verification

`pnpm harness:pre-push` exits 0. `pnpm scenario:verify` passes on this branch and fails on develop — both halves were run.

REVIEWED HEAD: ec8b76e4b6bf725f4109cc503fc3be71fd87cba5
ACTIONABLE FINDINGS: 0

Issue #1862, opened 2026-08-18
agent-cli: compose the local-peer rendezvous — guarded directory, grants, and the proof frame (follow-up to #1810)


woojubb:
Split out of #1810 so that issue can close on the work it actually specified. **Everything #1810 asked for on the security side is built and merged**; what remains is composition, and it is a different job with a different owner.

## What already exists, and where

| Piece | Package | Export |
| --- | --- | --- |
| Directory guarantee (owner + 0700, validated before binding) | `agent-remote-pairing/local` | `admitLocalPeerDirectory`, `admitLocalPeerSocket` |
| Creating that directory and verifying it afterwards | `agent-remote-pairing/local` | `ensureGuardedDirectory`, `GUARDED_MODE` |
| Single-use, time-bounded, revocable grants | `agent-remote-pairing/local` | `RendezvousGrantLedger`, `DEFAULT_GRANT_TTL_MS` |
| Fail-closed channel gate consuming a redeem port | `agent-transport-webrtc` | `IPairingGateOptions.localPeer`, `ILocalProofFrame` |

Verified at the time of writing: **no file under `packages/agent-cli/src` references any of these.**

## What this issue is

1. **Decide where the guarded directory goes** and create it at startup. The existing precedent in this package is `~/.robota/…` (`host-identity.ts`). A runtime directory (`XDG_RUNTIME_DIR`) is the other candidate and is already 0700 per-user; whichever is chosen, the decision belongs here and not in the security leaf — that leaf deliberately knows only what must be true of whatever path it is handed.
2. **Issue a grant** when a peer reaches the rendezvous, and wire `localPeer: { redeem, onAdmission }` into the WebRTC transport options.
3. **The peer side**: present `{ t: 'local-proof', nonce }` on the channel after the handshake accepts.
4. **Revoke on shutdown** — the runtime entry IS the grant, so a departing session must call `revokeRendezvous`.

## Constraints inherited from #1810 — do not re-decide these

- **The transport implements no cryptographic policy.** Single-use, expiry and revocation live in the ledger; the gate takes an injected `redeem` and reports what it was told. That is also what keeps `node:fs` out of the WebRTC package.
- **`ensureGuardedDirectory` must not be bypassed** by a `mkdir` here. `mkdir` with `recursive: true` leaves an *existing* directory's mode untouched, so one left at 0777 by an earlier process would be adopted silently — the chmod-then-verify sequence is the whole guarantee.
[...]

PR #1906, opened 2026-08-19
feat(agent-cli): carry a peer message between two local sessions (#1863)

woojubb:
## What this changes

Stage 3 of issue #1863: the carrier that moves one peer message between two local sessions. Everything above it was already built and waiting — the message and ack contracts (`agent-interface-transport`), ordering/duplicates/ack issuance (`agent-transport-protocol`'s ledger), and the session-side ingress (`agent-framework`). This adds bytes on a wire and nothing else.

`listenForPeerMessages` and `sendPeerMessage`, over a unix socket inside the guarded rendezvous directory from issue #1862.

## The decisions, and what they rest on

**Why a unix socket in the guarded directory.** The `same-user-same-host` claim rests on the *directory's* ownership and mode, exactly as it does for the rendezvous itself: a socket inside a 0700 directory owned by this uid is connectable only by that uid. Admission is established once, at bind time, by `admitLocalPeerSocket` — never re-derived per connection from anything a peer chooses.

Node's `net` exposes no `SO_PEERCRED`, so there is no per-connection credential to read. The module says that plainly rather than carrying a comment that would assert a property the code does not have.

**The sender admits the TARGET path before connecting.** That is why `admitLocalPeerSocket` takes a path rather than a directory: a path resolving outside the guarded directory carries none of its guarantees, and connecting first would already have handed the text over.

**One message per connection.** Connect, write one JSON line, read one line, close. A persistent multiplexed stream would need framing, backpressure and a reconnect policy — three places to be wrong — to carry a control message a person types. Ordering across messages is the ledger's `sequence`, never the socket's.

**A closed connection with no line is a failure, not an empty message.** Resolving `''` there would hand the caller a parse error one layer away from the fact that nothing arrived.

**A message that cannot be read is answered with a `refused` ack, not dropped.** The sender is waiting, and silence is indistinguishable from a peer that died mid-send — which is the distinction the delivery states exist to make.

**Duplicates are not re-decided here.** The ledger answers a repeated id with the ORIGINAL verdict; this carries that back rather than forming its own opinion.

## What this does not do

It is not wired to `/peers send` or to `PeerMessageIngress` yet — that is the next stage and what issue #1863's definition of done needs. The task record tracks it as an open item rather than implying this closes the issue.

## Verification

9 tests, and `pnpm harness:pre-push` exits 0.

One test correction worth naming: the first crash-recovery test proved nothing — an in-process server cannot leak its socket path, because closing it unlinks the file. It now writes the stale socket file directly.

### DIFF
--- a/.agents/tasks/PEER-005-a-local-channel-between-two-sessions.md
+++ b/.agents/tasks/PEER-005-a-local-channel-between-two-sessions.md
@@ -1,0 +1,100 @@
+---
+title: 'PEER-005: nothing carried a message between two local sessions'
+status: in-progress
+created: 2026-08-19
+priority: high
+urgency: now
+area: packages/agent-cli
+depends_on: []
+---
+
+# PEER-005: the carrier, and only the carrier
+
+## Objective
+
+Stage 2 of issue #1863's remaining scope. Everything above a carrier was already built and waiting
+for one:
+
+| Piece                              | Package                                  | State before this            |
+| ---------------------------------- | ---------------------------------------- | ---------------------------- |
+| Message + ack contracts            | `agent-interface-transport`              | built                        |
+| Ordering, duplicates, ack issuance | `agent-transport-protocol` (the ledger)  | built                        |
+| Session-side ingress               | `agent-framework` (`PeerMessageIngress`) | built, and called by nothing |
+| Bytes between two processes        | —                                        | **absent**                   |
+
+So the message could be shaped, judged and delivered into a session, and there was no way for one to
+reach another. That is what this adds, and it adds nothing else.
+
+## The decisions
+
+**A unix socket inside the guarded directory.** The claim `same-user-same-host` rests on that
+directory's ownership and mode, exactly as it does for the rendezvous. A socket inside a 0700
+directory owned by this uid can be connected to only by that uid — or by root, which already controls
+the process. Admission is established once, at bind time, by `admitLocalPeerSocket`, and never
+re-derived per connection from something a peer chooses.
+
+**The absent guarantee is stated, not implied.** Node's `net` exposes no `SO_PEERCRED`, so there is
+no per-connection credential to read. The guarantee is the directory's. A comment claiming
+per-connection verification would assert a property the code does not have — which is the defect
+class this repository has spent the most review rounds on.
+
+**One message per connection.** Connect, write one JSON line, read one line, close. A persistent
+multiplexed stream needs framing, backpressure and a reconnect policy — three places to be wrong —
+to carry a control message that is sent when a person types. Ordering across messages is the
+ledger's `sequence`, never the socket's.
+
+**The carrier forms no verdict.** `duplicate`, `refused` and `delivered` all come from the ledger and
+are carried back untouched. A carrier that re-decided delivery states would give the sender a second
+opinion about rules that must have exactly one owner.
+
+**An unreadable message is answered with a `refused` ack, not dropped.** The sender is waiting, and
+silence is indistinguishable from a peer that died mid-send — the exact distinction the delivery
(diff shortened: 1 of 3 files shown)


woojubb:
## Review — feat(agent-cli): carry a peer message between two local sessions (#1863)

Three files, all new: the carrier module, its tests, and the task record. Nothing existing changes, so the blast radius is exactly what the new exports reach — which today is nothing, by design.

### What I checked

- **Admission is established once, at bind time, and never re-derived from anything a peer supplies.** `admitOrThrow` runs `admitLocalPeerSocket` before the server binds and before the sender connects. The sender admitting the TARGET path *before* connecting is the load-bearing ordering: connecting first would already have handed the text over, and a path resolving outside the guarded directory carries none of that directory's guarantees.
- **`admitLocalPeerSocket` tolerates a socket that does not exist yet.** The listener depends on this at bind time, so I read the implementation rather than assuming: it resolves through the parent directory and containment-checks the computed path, with the fallback documented as `allow-fallback` and reasoned. Had it required the file to exist, the listener could never have bound.
- **The security claim is stated at its true strength.** The comment says plainly that Node's `net` exposes no `SO_PEERCRED` and that the guarantee is the directory's, not per-connection. A comment claiming per-connection verification would have asserted a property the code does not have — the failure mode this repo has been bitten by before.
- **A refusal carries its reason, and a failed read still answers.** A message that cannot be parsed produces a `refused` ack rather than silence; silence would be indistinguishable from a peer that died mid-send, which is precisely the distinction the delivery states exist to make.
- **It does not re-decide what the ledger owns.** Duplicates, ordering and ack issuance stay in `agent-transport-protocol`; this carries the ledger's verdict back rather than forming its own. The header comment names duplicates specifically as the one easiest to re-decide by accident.
- **Removing a stale socket before bind is safe for a stated reason**, not by convention: the directory is 0700 and ours, so nothing else could have put a file there.

### Test quality

The crash-recovery test earns a note because its first version proved nothing: an in-process server cannot leak its socket path, since closing it unlinks the file. It now writes the stale socket file directly, which is the condition the code actually has to survive. A test that cannot fail on the condition it names is worse than no test, and this one was caught rather than shipped.

### Scope

Not wired to `/peers send` or `PeerMessageIngress` — that is the next stage and what issue #1863's definition of done requires. The task record tracks it as an open item rather than implying this closes the issue, which is the right call: `orphan-exports` will start asking about these two functions the moment they have no consumer, and the answer will be the wiring rather than a suppression.

### Verification

9 tests pass. `pnpm harness:pre-push` exits 0. CI is green with no failing or pending checks.

REVIEWED HEAD: cba21aeb00a60e7b9364cb3817357a8c343289ca
ACTIONABLE FINDINGS: 0

woojubb:
## Correction to the review above

I asserted that `orphan-exports` "will start asking about these two functions the moment they have no consumer." That is wrong, and I checked it rather than leaving it standing.

`check-orphan-exports.mjs` defines an orphan as a symbol whose **name appears in no other scanned file**. `local-peer-channel.test.ts` references both `listenForPeerMessages` and `sendPeerMessage`, so the scan is satisfied today and would remain satisfied even if the wiring never landed. That is the scan's documented v1 scope — a tripwire for refactor fallout, not reachability analysis — and it is why the pre-push on this branch was green rather than in spite of it.

This changes nothing about the verdict; the scope note it was attached to still holds, and the reason the wiring must land is issue #1863's definition of done, not a scan that would otherwise complain. But a review that names a mechanism as the guard is telling the next reader where to look, and pointing them somewhere that will never fire is worse than saying nothing.

REVIEWED HEAD: cba21aeb00a60e7b9364cb3817357a8c343289ca
ACTIONABLE FINDINGS: 0

PR #1907, opened 2026-08-19
docs(tasks): complete four records whose work is merged on develop

woojubb:
## What this changes

Four task records still read `status: in-progress` while their work is merged on develop. `status` is the single source of truth for what is still being worked on, so a stale one misdirects the next reader.

| record | what it delivered | verified present on develop |
| --- | --- | --- |
| INFRA-106 | the reference-kind rule, its tree ratchet and the commitlint rule | `scripts/harness/reference-kind.mjs`, `scan-reference-kind-qualified.mjs` |
| INFRA-107 | the authenticated action-manifest probe | `pathExistsAtCommit` in `scan-action-references.mjs` |
| INFRA-108 | the `await` the scenario verifier was missing | `await createSession` in the scenario helpers |
| PEER-005 | the local peer channel | `packages/agent-cli/src/remote-control/local-peer-channel.ts` |

Each was confirmed by reading the file on develop, not by trusting the PR state — a merged PR and a present deliverable are not the same claim.

## One judgement call, stated rather than buried

PEER-005 carried an unchecked **TC-10: "wired to `/peers send` and to `PeerMessageIngress`"**, whose own line says *"the next stage"*. That was never this item's work, and leaving it as a plan item made the record unclosable for a reason that is not real — the kind of thing that either blocks a correct close forever or gets quietly ticked.

It moves to a `## Follow-on` section with the reasoning written down. A plan item is what this unit of work delivers; work nobody intends to do here belongs where a reader will find it, which is issue #1863's remaining stage.

I did **not** tick it.

## Verification

`pnpm harness:scan` — 126 passed, 3 skipped. `pnpm harness:pre-push` green.

### DIFF
--- a/.agents/tasks/completed/INFRA-106-a-bare-hash-number-does-not-say-issue-or-pull-request.md
+++ b/.agents/tasks/completed/INFRA-106-a-bare-hash-number-does-not-say-issue-or-pull-request.md
@@ -1,7 +1,8 @@
 ---
 title: 'INFRA-106: a bare #N does not say whether it is an issue or a pull request'
-status: in-progress
+status: done
 created: 2026-08-19
+completed: 2026-08-20
 priority: medium
 urgency: next
 area: .agents/rules, scripts/harness, commitlint.config.js
--- a/.agents/tasks/completed/INFRA-107-action-references-is-rate-limited-into-a-red-gate.md
+++ b/.agents/tasks/completed/INFRA-107-action-references-is-rate-limited-into-a-red-gate.md
@@ -1,7 +1,8 @@
 ---
 title: 'INFRA-107: action-references probes an anonymous endpoint, so a shared CI address turns it red'
-status: in-progress
+status: done
 created: 2026-08-19
+completed: 2026-08-20
 priority: high
 urgency: now
 area: scripts/harness
--- a/.agents/tasks/completed/INFRA-108-scenario-verifier-not-typechecked.md
+++ b/.agents/tasks/completed/INFRA-108-scenario-verifier-not-typechecked.md
@@ -1,7 +1,8 @@
 ---
 title: 'INFRA-108: a breaking signature change reached develop because packages/*/examples is typechecked by nobody'
-status: in-progress
+status: done
 created: 2026-08-20
+completed: 2026-08-20
 priority: high
 urgency: now
 area: packages/agent-command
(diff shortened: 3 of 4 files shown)


woojubb:
## Review — docs(tasks): complete four records whose work is merged on develop

Lifecycle only: four frontmatter edits and four `git mv` into `completed/`. No code, no rules, no harness wiring.

### What I checked

- **Each deliverable was confirmed present on develop by reading the file**, not inferred from a merged PR. Those are different claims — a PR can merge and leave the thing it promised elsewhere. `reference-kind.mjs` and `scan-reference-kind-qualified.mjs` (INFRA-106), `pathExistsAtCommit` (INFRA-107), `await createSession` in the scenario helpers (INFRA-108), and `local-peer-channel.ts` (PEER-005) are all there.
- **The lifecycle rule is followed exactly as written**: `status` set to `done`, a real `completed:` date added, and the move to `completed/` in the SAME commit. `.agents/tasks/README.md` states that splitting them is what must not happen, and it is not split here.
- **No open plan item was ticked to make a record closable.** Three of the four had none. The fourth is the judgement call below.
- **No status text was added to any body.** The README makes frontmatter the single source of truth and bans `## Status` sections; the `## Follow-on` section added to PEER-005 states remaining WORK, not status.

### The one judgement call

PEER-005 carried an unchecked **TC-10: "wired to `/peers send` and to `PeerMessageIngress`"**, and that line says on its own face *"the next stage"*. It was never this item's work. Leaving it made the record unclosable for a reason that is not real, which ends one of two ways: the record stays open forever, or someone eventually ticks it and the plan becomes a lie.

It moves to `## Follow-on` with the reasoning recorded in the commit message. **It was not ticked.** A plan item is what this unit of work delivers; work nobody intends to do here belongs where a reader will find it — issue #1863's remaining stage, which is now the only thing between that issue and closure.

Worth being explicit that this is the kind of edit that can be abused: rewriting a plan to make a task closable is exactly how a record stops meaning anything. The defence is that it is stated rather than quiet, and that the fact it carried survives in a section a reader will actually reach.

### Verification

`pnpm harness:scan` — 126 passed, 3 skipped. `pnpm harness:pre-push` green. CI clean with no failing or pending checks.

REVIEWED BASE: fb819383ba3bd1c2a1f7202d03d338ae7237cae1
REVIEWED HEAD: 5df69965e9216c59b733481d5916f597f9e62a05
ACTIONABLE FINDINGS: 0

PR #1914, opened 2026-08-19
fix(ci): exempt the three sharp wrappers that carry libvips themselves (#1889)

### LINKED ISSUE #1889: Dependency review blocks every lockfile-touching PR: the sharp license allow-list omits win32
## Symptom

`Dependency review` fails on any PR that touches `pnpm-lock.yaml`:

```
The following dependencies have incompatible licenses:
  pnpm-lock.yaml » @img/sharp-win32-arm64@0.35.2 – License: Apache-2.0 AND LGPL-3.0-or-later
  pnpm-lock.yaml » @img/sharp-win32-ia32@0.35.2  – License: Apache-2.0 AND LGPL-3.0-or-later
  pnpm-lock.yaml » @img/sharp-win32-x64@0.35.2   – License: Apache-2.0 AND LGPL-3.0-or-later
```

No vulnerability is involved — the same run reports *"did not detect any vulnerable packages with severity level high or higher"*. It is purely the license allow-list.

## The gap

`allow-dependencies-licenses` in the workflow enumerates sharp's platform binaries by name and covers **darwin, linux, linuxmusl and wasm32 — but not win32**:

```
@img/sharp-libvips-darwin-arm64, @img/sharp-libvips-darwin-x64,
@img/sharp-libvips-linux-arm, @img/sharp-libvips-linux-arm64,
@img/sharp-libvips-linux-ppc64, @img/sharp-libvips-linux-riscv64,
@img/sharp-libvips-linux-s390x, @img/sharp-libvips-linux-x64,
@img/sharp-libvips-linuxmusl-arm64, @img/sharp-libvips-linuxmusl-x64,
@img/sharp-wasm32
```

Every sharp platform binary carries the same `Apache-2.0 AND LGPL-3.0-or-later`, so the omission is a coverage hole in the list rather than a judgement that win32 is different. Note the listed names are the `sharp-libvips-*` prefabs while the flagged ones are the `sharp-win32-*` wrappers — worth checking whether both families need entries for every platform, or whether only one family is ever surfaced.

## Not branch-local

Established rather than assumed, on `fix/arch-035-single-owner-default-tool-set` (PR #1885):

- `git diff origin/develop -- pnpm-lock.yaml | grep '^+.*sharp-win32'` returns **nothing** — this branch adds none of them.
- `git show origin/develop:pnpm-lock.yaml | grep -c sharp-win32` returns **12** — they are already on the target branch.

So the branch did not introduce the dependency. It touched `pnpm-lock.yaml` (adding a new workspace package), which is what makes `dependency-review` evaluate the file at all. Any PR that adds or removes a dependency hits this.

## Why it is worth fixing rather than re-running

This gate cannot go green by retry — the input does not change. It blocks every lockfile-touching PR on a condition unrelated to that PR's content, which trains people to read a red `Dependency review` as noise. That is the failure mode a license gate can least afford: the one time it flags a genuinely incompatible new dependency, it will look like this.

## Suggested direction

Add the `@img/sharp-win32-*` entries (and check whether the `sharp-libvips-win32-*` prefabs need them too). Alternatively decide the platform-binary enumeration is the wrong shape — a per-package allow entry that has to list every platform triple will keep drifting as sharp adds targets — and express it as a single `@img/*` allowance with a recorded reason instead.

Found while merging ARCH-035 (#1787).

woojubb:
Closes #1889.

## The symptom

`Dependency review` fails on any pull request that touches `pnpm-lock.yaml`, naming three packages the pull request did not introduce:

```
  pnpm-lock.yaml » @img/sharp-win32-arm64@0.35.2 – License: Apache-2.0 AND LGPL-3.0-or-later
  pnpm-lock.yaml » @img/sharp-win32-ia32@0.35.2  – License: Apache-2.0 AND LGPL-3.0-or-later
  pnpm-lock.yaml » @img/sharp-win32-x64@0.35.2   – License: Apache-2.0 AND LGPL-3.0-or-later
```

No vulnerability is involved — the same run reports none. It is the license allow-list alone.

## Why fix rather than re-run

The gate cannot go green by retry, because the input does not change. It blocks every lockfile-touching pull request on a condition unrelated to that pull request's content, which teaches every reader to treat a red `Dependency review` as noise. That is the failure a license gate can least afford: the one time it catches a genuinely incompatible new dependency, it will look exactly like this.

## The question the issue left open, answered by measurement

The issue asked whether both `sharp-*` families need entries for every platform. They do not — and the reason is structural, not an oversight in the list.

Read from the registry for all 27 `@img/*` packages in this lockfile:

| package | license | libvips comes from |
| --- | --- | --- |
| `@img/sharp-darwin-arm64` | `Apache-2.0` | optionalDep → `sharp-libvips-darwin-arm64` |
| `@img/sharp-linux-x64` | `Apache-2.0` | optionalDep → `sharp-libvips-linux-x64` |
| `@img/sharp-freebsd-wasm32` | `Apache-2.0` | — |
| `@img/colour` | `MIT` | — |
| `@img/sharp-libvips-darwin-arm64` | `LGPL-3.0-or-later` | *is* the prefab |
| **`@img/sharp-win32-x64`** | **`Apache-2.0 AND LGPL-3.0-or-later`** | **no optionalDep — bundled in the wrapper** |
| **`@img/sharp-win32-arm64`** | **`Apache-2.0 AND LGPL-3.0-or-later`** | **no optionalDep — bundled** |
| **`@img/sharp-win32-ia32`** | **`Apache-2.0 AND LGPL-3.0-or-later`** | **no optionalDep — bundled** |

**Windows is the one platform where the wrapper and the prefab are the same package.** Every other wrapper delegates libvips to a separate `sharp-libvips-*` package and stays plain `Apache-2.0`, which `allow-licenses` already passes — which is why the wrapper family needed no entries and still does not. The win32 wrappers ship the DLLs inside themselves, so they carry the LGPL leaf directly, and the action's `A AND B` rule passes only when **both** leaves are allowed.

So this is a coverage hole in the existing exemption, not a new judgement about Windows. LGPL stays off the general allow-list; a new LGPL dependency still fails.

## Two things deliberately not done

**`@img/sharp-libvips-win32-*` is not listed.** It exists on the registry at 1.3.2 — checked — but is absent from this lockfile. An exemption for a package the graph does not contain is one nobody can observe being wrong.

**Not replaced with a blanket `@img/*` allowance**, which the issue offered as an alternative. `@img/colour` is MIT and the other wrappers are Apache-2.0, so it would exempt 24 packages that need no exemption and would silently absorb any future `@img/*` package whatever its license. The enumeration drifting is the price of it being observable — and this change *is* that drift being caught.

## Verification

`pnpm harness:scan` — 127 passed, 3 skipped. The workflow still parses as YAML with the job name unchanged.

The task record leaves **TC-06 open**: only a real lockfile-touching pull request can show the gate green, and a workflow edit that looks right is not the same claim as a gate that passed.

### DIFF
--- a/.agents/tasks/INFRA-115-dependency-review-omits-win32.md
+++ b/.agents/tasks/INFRA-115-dependency-review-omits-win32.md
@@ -1,0 +1,111 @@
+---
+title: 'INFRA-115: dependency-review blocks every lockfile-touching pull request on a coverage hole'
+status: in-progress
+created: 2026-08-20
+priority: high
+urgency: now
+area: .github/workflows
+depends_on: []
+---
+
+# INFRA-115: the sharp allow-list covers four platforms and there are five
+
+## Objective
+
+`Dependency review` fails on any pull request that touches `pnpm-lock.yaml`, on three packages the
+pull request did not introduce. Issue #1889.
+
+```
+  pnpm-lock.yaml » @img/sharp-win32-arm64@0.35.2 – License: Apache-2.0 AND LGPL-3.0-or-later
+  pnpm-lock.yaml » @img/sharp-win32-ia32@0.35.2  – License: Apache-2.0 AND LGPL-3.0-or-later
+  pnpm-lock.yaml » @img/sharp-win32-x64@0.35.2   – License: Apache-2.0 AND LGPL-3.0-or-later
+```
+
+No vulnerability is involved; the same run reports none. It is the license allow-list alone.
+
+## Why this is worth fixing rather than re-running
+
+The gate cannot go green by retry — the input does not change. It blocks every lockfile-touching
+pull request on a condition unrelated to that pull request's content, which teaches every reader to
+treat a red `Dependency review` as noise. That is the failure a license gate can least afford: the
+one time it catches a genuinely incompatible new dependency, it will look exactly like this.
+
+## Measured, not assumed
+
+The issue asks whether both `sharp-*` families need entries for every platform. They do not, and the
+measurement says why:
+
+| package                           | license                                | libvips comes from           |
+| --------------------------------- | -------------------------------------- | ---------------------------- |
+| `@img/sharp-darwin-arm64`         | `Apache-2.0`                           | optionalDep on a prefab      |
+| `@img/sharp-linux-x64`            | `Apache-2.0`                           | optionalDep on a prefab      |
+| `@img/sharp-freebsd-wasm32`       | `Apache-2.0`                           | —                            |
+| `@img/sharp-webcontainers-wasm32` | `Apache-2.0`                           | —                            |
+| `@img/colour`                     | `MIT`                                  | —                            |
+| `@img/sharp-libvips-darwin-arm64` | `LGPL-3.0-or-later`                    | is the prefab                |
+| **`@img/sharp-win32-x64`**        | **`Apache-2.0 AND LGPL-3.0-or-later`** | **no optionalDep — bundled** |
+| **`@img/sharp-win32-arm64`**      | **`Apache-2.0 AND LGPL-3.0-or-later`** | **no optionalDep — bundled** |
+| **`@img/sharp-win32-ia32`**       | **`Apache-2.0 AND LGPL-3.0-or-later`** | **no optionalDep — bundled** |
+
+Windows is the one platform where the wrapper and the prefab are the same package. Every other
+wrapper delegates libvips to a `sharp-libvips-*` package and stays plain `Apache-2.0`, which
+`allow-licenses` already passes — which is why the wrapper family needed no entries and still does
+not. The win32 wrappers ship the DLLs inside themselves, so they carry the LGPL leaf directly, and
+the action's `A AND B` rule passes only when BOTH leaves are allowed.
+
+So this is a coverage hole in the existing exemption, not a new judgement about Windows.
+
+`@img/sharp-libvips-win32-*` **does** exist on the registry at 1.3.2, but is absent from this
+lockfile. Listing it would exempt a package the graph does not contain — an entry nobody can observe
+being wrong.
+
+## The alternative, and why it was not taken
+
+The issue offers replacing the enumeration with a single `@img/*` allowance. Rejected: `@img/colour`
+is MIT and the non-win32 wrappers are Apache-2.0, so a blanket entry would exempt 24 packages that do
+not need exempting, and would silently absorb any future `@img/*` package whatever its license. The
+enumeration drifting is the cost of it being observable — and this task is the drift being caught,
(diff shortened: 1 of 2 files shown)


woojubb:
## Review — fix(ci): exempt the three sharp wrappers that carry libvips themselves (#1889)

Three entries appended to an existing allow-list, plus the measurement behind them and a task record. No code.

### What I checked

- **The cause was established, not inferred from the three failing names.** Every `@img/*` package in this lockfile (27 of them) had its license read from the registry. That is what showed the pattern is structural: every wrapper except win32 delegates libvips to a `sharp-libvips-*` package and stays plain `Apache-2.0`; the win32 wrappers have **no** `optionalDependencies` and ship the DLLs themselves, so they carry the LGPL leaf directly. Windows is the one platform where the wrapper and the prefab are the same package.
- **The exemption stays narrow.** LGPL is still off `allow-licenses`, so a genuinely new LGPL dependency still fails. This adds three names to a family exemption that already existed for exactly this reason.
- **The absent prefab is deliberately absent.** `@img/sharp-libvips-win32-*` exists on the registry at 1.3.2 — checked — and is not in this lockfile. Listing it would exempt a package the graph does not contain, which is an entry nobody can observe being wrong.
- **The blanket alternative was considered and rejected with a reason.** `@img/*` would exempt 24 packages needing no exemption (`@img/colour` is MIT, the other wrappers Apache-2.0) and would absorb any future `@img/*` package whatever its license. The enumeration drifting is the price of it being observable, and this change is that drift being caught rather than the mechanism failing.
- **The comment in the workflow carries the measurement**, not just the conclusion — including the three-row table showing why darwin/linux need nothing. The next person to hit a new sharp platform can tell which list to extend without re-deriving it.
- **The YAML still parses** and the job name is unchanged, so branch protection sees the same required check.

### The honest limit

**TC-06 is left open in the task record**: only a real lockfile-touching pull request can show this gate green. A workflow edit that looks right is not the same claim as a gate that passed, and this pull request does not itself touch `pnpm-lock.yaml`, so its own `Dependency review` run does not prove the fix.

That is the right call rather than a gap — but it means the first lockfile-touching pull request after this merges is the evidence, and someone should look at it.

### Why this is worth merging promptly

The gate cannot go green by retry; the input does not change. Every lockfile-touching pull request is currently red on a condition unrelated to its content, which trains readers to treat `Dependency review` as noise — the one failure mode a license gate cannot afford.

### Verification

`pnpm harness:scan` — 127 passed, 3 skipped. CI clean, no failing or pending checks.

REVIEWED BASE: da6c633e2f609150cb13405747a1f13a4ea76be9
REVIEWED HEAD: d85c16e54bde4bf225f1a16490ff1d68e1b4214f
ACTIONABLE FINDINGS: 0

Issue #1916, opened 2026-08-19
A work-item ID has no allocator, and three collisions landed in one session


woojubb:
A work-item ID (`INFRA-nnn`, `PEER-nnn`, `ARCH-nnn`) is allocated by a human or agent reading the current highest number and adding one. Nothing owns the allocation and nothing checks it, so two items can hold one ID and the tree will not say so.

## Three collisions in one session, all the same shape

| mine | collided with | filed |
| --- | --- | --- |
| `INFRA-108` (scenario verifier) | issue #1898, 2026-08-19T14:30:24Z | renumbered in PR #1911 |
| `INFRA-115` (dependency-review win32) | issue #1912, 2026-08-19T16:52:03Z | **live on develop now** |
| `INFRA-116` (the renumber of the first) | issue #1913, 2026-08-19T16:54:51Z | **live on develop now** |

The third is the one that shows this is not carelessness. It happened **while fixing the first**: I surveyed the claimed IDs, chose the next free one, and by the time the rename was written the number had been taken by an issue opened minutes earlier. Review caught that one and I re-derived — and the number I re-derived to had *also* been taken, by an issue I could not see because it was opened between my two reads.

## Why a survey cannot fix it

The repository already has the lesson as `claims-not-rederived-after-facts-moved`: a claim true when written is not one re-derived after the facts moved. An ID survey is exactly that shape — a measurement with a shelf life measured in minutes when more than one session is working.

Re-deriving "at the moment of use" narrows the window; it does not close it. Two sessions allocating concurrently will collide however carefully each one reads first, because there is no allocation, only observation.

## Why it matters more than tidiness

An ID is how a commit message, a pull request body, a review comment and a rule section all point at one record. When two items share one:

- A reader following `INFRA-115` out of a commit message lands on the wrong work.
- `named-artifact-resolves` and `resolving-claims` both pass, because each ID resolves to *a* file — just not the one the citation meant.
- The merged history cannot be rewritten, so the correction is always a forwarding note rather than a fix.

## What would actually close it

The mechanism has to make a collision **impossible to commit**, not merely visible afterwards. Options, in the order I would consider them:

1. **A scan over the union of claimed IDs** — tracked task records plus open and closed issue titles — refusing any ID that names two distinct items. Catches it at pre-push instead of at review. Does not prevent two sessions from choosing the same number, but neither of them can push it.
[...]

PR #1917, opened 2026-08-19
feat(harness): derive a hook's accepted override forms from its own source (#1904)

### LINKED ISSUE #1904: INFRA-112: a hook's accepted override forms have no owner, and the declarations have drifted both ways
Every guard in `.claude/hooks/` declares an escape hatch. What it accepts is decided by the code reading the variable; what it is DECLARED to accept is written by hand in up to five other places — the refusal message, a second refusal in the same file, the rule section, the card in `AGENTS.md`, and the task record's completion criterion. Nothing compares any of them to the code.

Both drift directions are now measured.

**Declared but not accepted** — corrected in pull request #1886, round two: `bulk-edit-guard.sh` advertised `BULK_EDIT_ACK=1` on a `Write` refusal, and a `Write` carrying it exported was refused anyway, because the tool branch exited above the check. That fix updated one of four copies of the declaration and left three.

**Accepted but not declared** — live on the integration branch today:

- `no-foreground-wait.sh` accepts an exported `FOREGROUND_WAIT_ACK=1` (probed: `sleep 300` exits 0), while its refusal and `operational.md` name the inline form only.
- `bulk-edit-guard.sh` accepts an exported `BULK_EDIT_ACK=1`, while `operational.md`, `AGENTS.md` and its own enumerator refusal name the inline form only.

`AGENTS.md` states "Each has a documented **inline** override", which is false for at least these two. An undeclared bypass is worse than a declared one: the reader who needs it does not know it exists, and the reader auditing the guard does not know it is there.

Correcting the two strings this round found leaves the enumerator refusal, the `AGENTS.md` card, and the whole `no-foreground-wait` pair wrong — and the next hook restates the declaration a sixth time.

Root item: `.agents/tasks/INFRA-111-override-forms-have-no-owner.md`. Filed as FOUNDATIONAL from review of pull request #1886.

woojubb:
Closes #1904.

## What was actually wrong

Every guard in `.claude/hooks/` declares an escape hatch. What it **accepts** is decided by the code reading the variable; what it is **declared** to accept was written by hand in up to five other places, and nothing compared any of them to the code.

Measured across all 14 hooks before anything was written — and the shape is worse than the issue described. There are **three** accepted forms, not one:

| form | variables | how the hook reads it |
| --- | --- | --- |
| inline only | `MERGE_GATE_ACK`, `PRE_PUSH_ALLOW_UNREVIEWED`, `WORKTREE_CWD_GUARD_ALLOW_MAIN` | greps the command **string** |
| either | the six `BRANCH_GUARD_ALLOW_*`, `BULK_EDIT_ACK`, `FOREGROUND_WAIT_ACK` | both, via `stmt_override` or a pair of checks |
| environment only | `HOOK_EDIT_ACK`, `LOCKFILE_CHURN_ACK` | `${VAR:-}` from its own environment |

The forms are not interchangeable and have **opposite lifetimes**. Inline excuses only the statement it prefixes; an exported variable stays armed for every later command until unset. A PreToolUse hook runs as a separate process, so an inline assignment on the agent's command never reaches one that reads its environment — the command runs, the guard refuses anyway, and the reader concludes the guard is broken rather than the declaration.

## Four live defects, found by the scan on its first run

| finding | variable |
| --- | --- |
| declared INLINE in its own refusal, accepted only from the environment | `LOCKFILE_CHURN_ACK` |
| accepted by a hook, declared in no rule at all | `BRANCH_GUARD_ALLOW_BADNAME` |
| accepted by a hook, declared in no rule at all | `BRANCH_GUARD_ALLOW_MAIN_MERGE` |
| accepted by a hook, declared in no rule at all | `WORKTREE_CWD_GUARD_ALLOW_MAIN` |

`AGENTS.md`'s "Each has a documented **inline** override" was false for two of the fourteen and understated eight more. All four are fixed here, and `AGENTS.md` now points at the owning section instead of generalising.

## The design decision

**The accepted forms are derived from the hook source, never listed.** A hand-kept list would be a sixth copy of exactly the thing that drifted, and it would drift the same way — one edit to a hook, none to the list, nothing comparing them. `git-branch.md` § "Which Form An Override Takes" says on itself that it is not the authority; the hook source is.

## Two detectors narrowed, one widened — each for a reason worth keeping

Narrowed, and both narrowings are one lesson: **a detector that reads prose loosely reports drift that is not there**, and the cost lands on whoever rewords a sentence to make a scan happy without understanding why.

- `HOOK_EDIT_ACK` was reported as declared-inline because its refusal reads `HOOK_EDIT_ACK=1 (git-branch.md)` — a citation, not a command line. The hook says "in the environment" two lines above, correctly.
- `MERGE_GATE_ACK` was reported as declared-exported because a wrapped paragraph put the variable on the same **source line** as a sentence about the environment form. Where prose happens to wrap is not a claim.

Widened, in the other direction: the first inline detector required the character class immediately beside `NAME=1` and so missed `merge-gate.sh` and `pre-push-check.sh`, whose acceptors put a capture group between them. Checking both directions is what surfaced it — those two came back accepting **nothing**, and a hook that accepts nothing is a finding rather than a quiet zero.

Both false positives are pinned as passing cases, so a future widening cannot silently reintroduce them.

## Red-proofed against the real tree, both directions, each reverted

| defect injected | what went red |
| --- | --- |
| declare `LOCKFILE_CHURN_ACK` as inline in the rule table | `[wrong-form]` naming the variable and the reason |
| read a new `BRAND_NEW_HATCH_ACK` in `no-foreground-wait.sh` | `[undeclared]` naming the hook |

## Harness conformance

The scan brought four floors with it, each satisfied rather than waived: `fixture-floor` (test named after the scan), `harness-script-import-safety` (a test per module), `measurement-provenance` (exported reader **and** finder, exact assertion, re-asserted after a second run so an accumulating counter is told from a growing subject), and `examined-size adoption drift` (registered in the frozen set in this same change).

## Verification

`pnpm harness:scan` — 128 passed, 3 skipped. `pnpm harness:pre-push` exits 0. 10 unit cases across the two modules.

### DIFF
--- a/.agents/rules/git-branch.md
+++ b/.agents/rules/git-branch.md
@@ -499,6 +499,46 @@ malformed, duplicate, stale, or unreadable markers. Deliberate exception: `MERGE
 in the same command**, which prints that the gate did not verify — an override is a visible choice,
 not a silent one.
 
+## Which Form An Override Takes
+
+Every guard here declares an escape hatch, and **the form is not interchangeable**. There are two,
+they have different lifetimes, and using the wrong one fails silently — the command runs, the guard
+refuses anyway, and the reader concludes the guard is broken.
+
+**Inline** — `VAR=1 <command>`, in the same statement. The hook reads the command STRING, so the
+assignment must prefix the guarded command. It excuses only the statement it prefixes: `VAR=1 date;
+git push` does not disarm the push. This is the safer form, and most hatches use it.
+
+**Environment** — `export VAR=1`. The hook reads `${VAR}` from its own environment. A PreToolUse hook
+runs as a separate process and never sees an inline assignment on the agent's command, so only an
+exported variable reaches it. It then stays armed for every later command until unset, which is the
+opposite lifetime from inline and the reason this form is worth naming rather than assuming.
+
+| variable                                                                                          | hook                          | form        |
+| ------------------------------------------------------------------------------------------------- | ----------------------------- | ----------- |
+| `MERGE_GATE_ACK`                                                                                  | `merge-gate.sh`               | inline      |
+| `PRE_PUSH_ALLOW_UNREVIEWED`                                                                       | `pre-push-check.sh`           | inline      |
+| `WORKTREE_CWD_GUARD_ALLOW_MAIN`                                                                   | `worktree-cwd-guard.sh`       | inline      |
+| `BRANCH_GUARD_ALLOW_DELETE`, `_BASE`, `_BRANCH_COPY`, `_OPEN_BRANCHES`, `_BADNAME`, `_MAIN_MERGE` | `branch-guard.sh`             | either      |
+| `BULK_EDIT_ACK`                                                                                   | `bulk-edit-guard.sh`          | either      |
+| `FOREGROUND_WAIT_ACK`                                                                             | `no-foreground-wait.sh`       | either      |
+| `HOOK_EDIT_ACK`                                                                                   | `check-forbidden-patterns.sh` | environment |
+| `LOCKFILE_CHURN_ACK`                                                                              | `pre-push-check.sh`           | environment |
+
+`BRANCH_GUARD_ALLOW_BADNAME` exempts a branch name from the naming convention, and
+`BRANCH_GUARD_ALLOW_MAIN_MERGE` a push or merge into `main` for a release the user approved. Both
+were accepted by the hook and named in no rule until this section; an undeclared bypass is worse than
+a declared one, because the reader who needs it cannot find it and the reader auditing the guard does
+not know it is there.
+
+This table is NOT the authority on what works — the hook source is. It is the declaration the scan
+compares against that source, in both directions.
+
+Enforced by: `hook-override-declarations` (`scripts/harness/scan-hook-override-declarations.mjs`),
+which DERIVES the accepted forms from each hook's own code rather than from a list. A hand-kept list
+of accepted forms would be one more copy of exactly the thing that drifted — the declaration existed
+in up to five places and nothing compared any of them to the code.
+
 The hook deliberately does NOT judge whether a prose finding was addressed; that is the reviewer's
 call, and a hook guessing at it would be a check measuring the wrong thing. It establishes only that
 CI is green and that a current review exists to be read.
--- a/.agents/tasks/INFRA-117-override-forms-have-no-owner.md
+++ b/.agents/tasks/INFRA-117-override-forms-have-no-owner.md
@@ -1,0 +1,129 @@
+---
+title: 'INFRA-117: a hook escape hatch is declared in five places and compared against none'
+status: in-progress
+created: 2026-08-20
+priority: high
+urgency: now
+area: .claude/hooks, scripts/harness
+depends_on: []
+---
+
+# INFRA-117: the accepted form is derived from the hook, not restated beside it
+
+## Objective
+
+Issue #1904. Every guard in `.claude/hooks/` declares an escape hatch. What it ACCEPTS is decided by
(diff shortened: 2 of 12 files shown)


woojubb:
## Review — feat(harness): derive a hook's accepted override forms from its own source (#1904)

A derivation module, a scan, two test files, four live drift fixes, and the owning rule section.

### What I checked

- **The problem was measured before anything was written.** All 14 hooks, all 14 override variables — which is what showed **three** accepted forms rather than the one the issue assumed. That measurement is the whole design: a fix written from the issue's framing would have corrected two strings and left the shape wrong.
- **The forms genuinely differ in kind, not spelling.** Inline excuses only the statement it prefixes; an exported variable stays armed until unset. A PreToolUse hook is a separate process, so an inline assignment never reaches one reading its environment — the command runs, the guard refuses anyway, and the reader blames the guard.
- **Four live defects, each verified in the source before being called one.** `LOCKFILE_CHURN_ACK` said "inline" in its own refusal at `pre-push-check.sh:726` while reading `${LOCKFILE_CHURN_ACK:-}` at 715. The three undeclared hatches are read by their hooks and named in no rule — I read each acceptor rather than trusting the scan that had just been written to find them.
- **The derivation is not a list.** A hand-kept table of accepted forms would be the sixth copy of exactly what drifted. The rule section says on itself that the hook source is the authority and it is the declaration being judged.
- **The indirect acceptor is followed.** `stmt_override` tests `${!token:-0}` *and* a statement regex, so six `BRANCH_GUARD_ALLOW_*` variables accept both while appearing beside neither construct. A scan that looked only for the constructs would have reported six false "declared but not accepted" findings.

### The detector work, which is where the care went

Two narrowings and one widening, each from measuring against the real tree:

| change | why |
| --- | --- |
| inline claim → explicit only | `HOOK_EDIT_ACK=1 (git-branch.md)` is a citation, not a command line; the hook says "in the environment" correctly two lines above |
| exported claim → `export NAME=` only | a wrapped paragraph put `MERGE_GATE_ACK` on the same **source line** as a sentence about the environment form — where prose wraps is not a claim |
| inline detection widened | the first version required the character class beside `NAME=1` and missed two hooks whose acceptors put a capture group between them |

The widening is the one I'd point at: those two hooks came back accepting **nothing**, and a hook that accepts nothing is a finding rather than a quiet zero. Checking both directions is what surfaced it. Both false positives are pinned as passing cases, so a future widening cannot silently reintroduce them.

### Harness conformance, satisfied rather than waived

`fixture-floor` (test named after the scan), `harness-script-import-safety` (a test per module), `measurement-provenance` (exported reader **and** finder, exact assertion, re-asserted after a second run so an accumulating counter is told from a growing subject), `examined-size adoption drift` (registered in the frozen set in this same change).

### Red-proofed, both directions, against the real tree

Declaring an environment-only hatch as inline produces `[wrong-form]`; reading a new variable in a hook produces `[undeclared]`. Both probes reverted.

### One wrinkle, stated

The task record is `INFRA-117` while issue #1904's title carries `INFRA-112`. Those issue-title IDs were themselves renumbered once during this session, which is the instability issue #1916 is about. I did not renumber the record to match, because the pull request is open and the churn would add risk for no reader benefit — the record and the issue are linked in both directions.

### Verification

`pnpm harness:scan` — 128 passed, 3 skipped. `pnpm harness:pre-push` exits 0. 10 unit cases. CI clean.

REVIEWED BASE: c43580783b837c721e042ce7523ca4774c7dba24
REVIEWED HEAD: 96a66ec7a3e190e506b385490fb160783fe3ab60
ACTIONABLE FINDINGS: 0

Issue #1913, opened 2026-08-19
INFRA-116: a reference RANGE is qualified only at its first member


woojubb:
`reference-kind-qualified` requires a `#N` to say whether it is an issue or a pull request. Its `REFERENCE` model in `reference-kind.mjs` is a single qualifying token immediately before a single number, so every member of a RANGE after the first reads as bare — a form this repository writes routinely and a reader resolves without opening anything.

Measured on the current rule via `unqualifiedReferences`:

| line | reported |
| --- | --- |
| `Measured over PRs #1525-#1530: twelve runs` | yes |
| `issues #10-#12 are open` | yes |
| `See PR #1525 and PR #1530` | no |

It surfaces late because the scan is a per-file ratchet and every existing instance of the range form is frozen in its 294-entry baseline — including the completed record for INFRA-071, whose only frozen violation is the second member of that exact sentence. The gap costs nothing until a NEW document writes a range, at which point both site fixes are wrong: alter a verbatim quotation to satisfy a check mistaken about the text, or mark a real claim as a code-span "specimen", an exemption meant for something else.

INFRA-106 owns the rule and is complete; it enumerates its exemptions one at a time and never considered a range.

Root item: `.agents/tasks/INFRA-116-a-reference-range-is-qualified-only-at-its-first-member.md`. Found in review of the follow-up to pull request #1886.

woojubb:
Fixed by PR #1925, merged into develop.

**Wider than described.** Reproducing on the current model before writing anything showed ranges are one of three affected forms — a comma list and an `A and B` pair have the same defect. A fix written from this issue's framing would have left two of the three broken:

| line | before | after |
| --- | --- | --- |
| `Measured over PRs #1525–#1530` | flagged the second member | clean |
| `claimed by issues #1899, #1903, #1904` | flagged the second and third | clean |
| `pull requests #10 and #12` | flagged the second | clean |

**The separators are enumerated, not "any short run of punctuation".** The point is to recognise a phrase, not to forgive proximity: a verb between two references keeps them separate, because the second is a different claim and treating nearness as governance would let a real bare reference hide behind an unrelated qualified one on the same line. Chaining carries a qualifier forward and never invents one, so two bare references joined by a conjunction are still two findings.

Both dash forms are handled — the en dash is what this repository's prose actually uses, and a hyphen-only check would have been right about the form nobody writes.

[...]

Issue #1912, opened 2026-08-19
INFRA-115: "is this committed file a script, and in what language" has no owner


woojubb:
A scan that judges committed scripts answers two halves of one question — which files are scripts, and in what language — with two independent hand-written constants, and nothing checks that the halves describe the same population.

Measured in `scan-symlink-following-enumeration.mjs`: `dash`, `ksh` and `ash` are admitted by shebang while `SCANNED_EXTENSIONS` has no matching entry. Same content, same file:

| written as | verdict |
| --- | --- |
| `scripts/sweep` | reported |
| `scripts/sweep.ksh` | **clean** |

The file's own comment claims the interpreters are "exactly the ones `SCANNED_EXTENSIONS` also admits, and no more" — and the commit that wrote that sentence shipped the counter-example.

It is a cause rather than a typo because the three names were copied verbatim out of `scan-shell-portability.mjs`, which carries the identical asymmetry (`EXTENSIONS = /\.(sh|bash|zsh)$/` against `SHEBANG = /^#!.*\b(sh|bash|zsh|dash|ksh|ash)\b/`), into the file whose comment cites that scan as the lesson to follow. The lesson copied; the defect copied with it.

Aligning three names in one file leaves the first copy self-disagreeing, leaves the invariant in prose in both, and leaves the next scan to hand-write a third pair. A test can pin one out-of-set example — the current one does — but cannot pin an invariant that is a sentence.

Root item: `.agents/tasks/INFRA-115-is-this-file-a-script-and-in-what-language.md`. Filed as FOUNDATIONAL from review of the follow-up to pull request #1886.

woojubb:
Closed by `45e371183` on `develop` (PR #1945).

`scripts/harness/script-language.mjs` owns "is this a script, and in what language"; `scan-symlink-following-enumeration.mjs` and `scan-shell-portability.mjs` — where the asymmetry started — both read it.

Stronger than the issue asked for, because two weaker shapes were built first and each failed its own red proof. There is now no second list: each interpreter carries the extensions that name a file it runs, and a language's set is their union. The halves cannot disagree because there is only one, and the remaining assertion cannot be switched off by the row it checks.

Closing by hand rather than by the `Closes` footer: GitHub ignores closing keywords on a PR that does not target the default branch, which is INFRA-104.

PR #1918, opened 2026-08-19
docs(tasks): renumber two more records whose IDs an issue claimed first

woojubb:
## Two more IDs claimed by an issue first

| record | collided with | filed |
| --- | --- | --- |
| `INFRA-115` (dependency-review win32) | issue #1912, 2026-08-19T16:52:03Z | → `INFRA-118` |
| `INFRA-116` (scenario verifier) | issue #1913, 2026-08-19T16:54:51Z | → `INFRA-119` |

Both issues predate the records under those numbers. The earlier claim wins.

## Why this keeps happening, which matters more than the rename

This is the third and fourth collision of the session, and **the third happened while fixing the second**. In PR #1911 I re-derived the claimed IDs, chose a free one, and by the time the rename was written that number had been taken by an issue opened minutes earlier.

The repository already carries the lesson as `claims-not-rederived-after-facts-moved` — a claim true when written is not one re-derived after the facts moved — and I applied it. It was not enough. Re-deriving at the moment of use narrows the window; it cannot close it, because **there is no allocation, only observation**. Two sessions reading the same list and each taking "the next free number" will collide however carefully both read first.

Filed as issue #1916 with three candidate mechanisms and what each costs. This pull request is the cleanup that issue names as immediate, not the fix.

Each record carries a forwarding paragraph: the commits and pull requests that delivered the work say the old numbers and cannot be rewritten, so a reader following one out of a commit message needs to be told where it went. `INFRA-119` now records all three of its numbers, because it has had three.

## INFRA-118 also closes

Its `TC-06` asked for `Dependency review` **green on a pull request that actually touches `pnpm-lock.yaml`** — deliberately left open in PR #1914, because re-reading a workflow file is not the same claim as a gate that passed.

PR #1909 touched the lockfile and merged 2026-08-19T18:00:18Z, an hour after the fix, with `Dependency review: SUCCESS`. That is the gate itself, so the item is ticked against the run rather than against the diff, and the record moves to `completed/`.

## Verification

`pnpm harness:scan` — 127 passed, 3 skipped. The suite caught a real slip on the way: the first rename left the index pointing at the old path, so `reference-kind-qualified` died on `ENOENT` for a file that no longer existed. Staging the rename fixed it — and a scan that crashes rather than reporting is worth noticing, since a crash is not a pass.

### DIFF
--- a/.agents/tasks/completed/INFRA-118-dependency-review-omits-win32.md
+++ b/.agents/tasks/completed/INFRA-118-dependency-review-omits-win32.md
@@ -1,14 +1,25 @@
 ---
-title: 'INFRA-115: dependency-review blocks every lockfile-touching pull request on a coverage hole'
-status: in-progress
+title: 'INFRA-118: dependency-review blocks every lockfile-touching pull request on a coverage hole'
+status: done
 created: 2026-08-20
+completed: 2026-08-20
 priority: high
 urgency: now
 area: .github/workflows
 depends_on: []
 ---
 
-# INFRA-115: the sharp allow-list covers four platforms and there are five
+# INFRA-118: the sharp allow-list covers four platforms and there are five
+
+## Renumbered from INFRA-115
+
+Issue #1912 claimed `INFRA-115` on 2026-08-19T16:52:03Z, before this record was written. The earlier
+claim wins, so this one moved. PR #1914 and its commits still say INFRA-115 and cannot be rewritten;
+this paragraph is the bridge for a reader who follows the old number out of one.
+
+Re-derived at the moment of writing: issues hold INFRA-103 and INFRA-109 through INFRA-116, the tree
+holds INFRA-100 and INFRA-102 through INFRA-107, and INFRA-117 is taken by an open pull request. 118
+is the first free one.
 
 ## Objective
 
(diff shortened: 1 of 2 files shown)


woojubb:
## Review — docs(tasks): renumber two more records whose IDs an issue claimed first

Two `git mv`s with forwarding notes, plus closing one record on real evidence. No code.

### What I checked

- **Which record moves is decided by timestamps.** Issue #1912 claimed `INFRA-115` at 2026-08-19T16:52:03Z and issue #1913 claimed `INFRA-116` two minutes later; both records were written 2026-08-20. Earlier claim wins, verified against each issue's own `createdAt`.
- **The target numbers are free**, re-derived immediately before writing rather than from an earlier reading — which is the whole lesson of the preceding collisions. Issues hold INFRA-103 and 109–116; the tree holds 100 and 102–107; INFRA-117 is taken by an open pull request. 118 and 119 are the first two free.
- **Both forwarding notes exist and say where.** The commits and pull requests that delivered the work carry the old numbers and cannot be rewritten. `INFRA-119` records all three of its numbers, because it has had three.
- **`INFRA-118` closes against the gate, not the diff.** Its TC-06 asked for `Dependency review` green on a pull request that *actually* touches `pnpm-lock.yaml` — deliberately left open in PR #1914, because re-reading a workflow file is not that claim. PR #1909 touched the lockfile and merged 2026-08-19T18:00:18Z with `Dependency review: SUCCESS`. I checked that result directly rather than inferring it from the merge.

### The part that matters more than the rename

This is the third and fourth collision of the session, and **the third happened while fixing the second** — the survey was re-derived, a free number chosen, and by the time the rename was written that number had been taken by an issue opened minutes earlier.

Re-deriving at the moment of use narrows the window and cannot close it: there is no allocation, only observation. Two sessions each taking "the next free number" collide however carefully both read first. Filed as issue #1916 with three candidate mechanisms and what each costs; this pull request is the cleanup that issue names as immediate, not the fix.

### A slip worth naming

The first rename left the index pointing at the old path — a `git mv` that had failed silently fell through to a plain `mv`. `reference-kind-qualified` then died on `ENOENT` for a file that no longer existed. Staging the rename fixed it, and the useful part is that **a scan which crashes is not a scan that passed**; it was only visible because the suite reports the crash rather than the exit code alone.

### Verification

`pnpm harness:scan` — 127 passed, 3 skipped. CI clean, no failing or pending checks.

REVIEWED BASE: d72acdd412fa8222893b3d48b1bed9e53df616bd
REVIEWED HEAD: 93d85d8f8226b634e759e0b88d57a05318661731
ACTIONABLE FINDINGS: 0

PR #1920, opened 2026-08-19
feat(harness): exempt a pure re-export barrel from the size limit, once earned (ARCH-038)

### LINKED ISSUE #1806: ARCH-038: agent-framework's root barrel is a 689-line export list — ARCH-029 raised its frozen size and nothing splits it
## What

`packages/agent-framework/src/index.ts` is a 689-line barrel that lists every public export by name.
ARCH-029 raised its frozen entry in `scripts/harness/file-size-baseline.json` from **656 to 689** to
publish 27 command-axis role ports.

Two smaller entries moved in the same change: `interactive-session-base.ts` 385 → 396 (one required
member implementation plus its imports) and `interactive-session.ts` 1000 → 1001.

## Why it is filed rather than absorbed

`scan-file-size.mjs:11` is explicit: *"A baselined file may not GROW past its recorded count —
existing debt is frozen, not licensed."* ARCH-029 regenerated the baseline instead, which the scan
supports for "adopting a legitimate refactor" — but adopting it quietly is indistinguishable from
adopting debt, and independent review flagged exactly that: the constraint was accepted and
discussed in the spec document, and then nothing was filed. This is the missing half.

What ARCH-029 did do first, so this item starts from a real floor rather than a licence:

| file | before | after | how |
| --- | --- | --- | --- |
| `command-api/host-context.ts` | 448 | 127 | split into `host-roles.ts`, `session-roles.ts`, `agent-job-roles.ts` |
| `testing/command-host-double.ts` | 302 | 227 | split into `agent-job-host-double.ts` + `double-constants.ts` |
| `agent-command/src/context/context-command.ts` | 536 | 306 | split into `context-breakdown.ts` — **208 lines under its old frozen 514**, re-ratcheted |
| `commands/index.ts` | 306 | 279 | ports re-exported from the root barrel only; removed from the baseline entirely |

## The constraint

No `export *` form is available: `sdk-public-surface` rejects it ("agent-framework public barrels
must use explicit named exports so owner boundaries are auditable"), and that rule is right — it is
what keeps the published surface auditable. So a sub-barrel does not reduce the line count at the
top; every public name costs one line wherever it is listed.

## Direction

Either the root barrel stops being a single list (per-domain entry points the package's `exports`
map points at, so no one file carries all of them), or the size floor grows an explicit,
reason-carrying exemption for a pure re-export list — the way `aggregateNaming` and
`rolePortOptionals` carry per-entry reasons in `.agents/harness.config.json`. The second is smaller
and probably right: a barrel is not a monolith in the sense the anti-monolith rule means, and
treating it as one is what pushes people to regenerate the baseline.

## Found by

Independent review of the ARCH-029 branch (second round).

woojubb:
Closes #1806.

## The argument

The 300-line rule is an **anti-monolith** rule — "a file that handles multiple independent concerns must be split". A barrel handles no concerns: it lists names.

`sdk-public-surface` refuses `export *`, and that rule is right — it is what keeps the published surface auditable. The consequence is that every public name costs one line **wherever it is listed**, so a sub-barrel moves lines rather than removing any. There is no split that reduces the count; only a change to the package's `exports` map and its consumers, which is a real decision and not a size remedy.

Judging the list as debt is what made ARCH-029 raise the frozen count from 656 to 689 instead of arguing it. Adopting debt quietly is indistinguishable from adopting a refactor, and independent review flagged exactly that — this is the missing half.

## The exemption is EARNED, which is the whole design

`.agents/harness.config.json` → `fileSizeReexportBarrels` names each file with its own reason, the way `aggregateNaming` and `rolePortOptionals` already do. `scan-file-size.mjs` grants the exemption **only while every statement in that file is a re-export**:

- a listed file that has grown a function, a constant or an import → `reexport-barrel-exemption-unearned`, **and the limit is enforced on it as usual**
- a listed file the scan never measured → `reexport-barrel-exemption-unused`

A named exemption nothing verifies is the same defect as a guard that cannot fire, one level up — which is the reason this is an item rather than a config line.

## What moved

`packages/agent-framework/src/index.ts` leaves the burn-down list entirely (82 baselined entries → 81). An exempt file is never frozen: a count for a file the limit does not apply to is debt the ratchet can never collect, and it would reappear on every `--write-baseline`.

The rule text in `.agents/rules/code-quality.md` states the exemption and its earned condition, since that file owns the limit.

## Verification

- Red-proofed in **both** halves: granting the exemption without earning it fails exactly the unearned case; accepting every line as a re-export fails four of the purity cases.
- `pnpm harness:scan` — 127 passed, 3 skipped
- harness test tier — 4023 assertions green; the hermetic tier exits 0 with 1113/1113
- `pnpm typecheck` — 0 errors

One defect was caught by the repository's own gates before this was pushed: the config was read at module load, so importing the scan inside the hermetic tier's stripped tree threw before anything ran. It is read lazily now, and the exemptions are passed into the evaluator so the pure function stays a function of its arguments.


woojubb:
## Pre-merge review

**base:** `17e9c493e78822e725e5256e276c0784dfab6f51`
**head:** `6bfa29b7fb18d8f39b9eb3f68025dc10319e4c55`

Reviewed the whole diff at this head. No review threads were opened; CI is green.

### What I checked, and why it is the part that matters

The risk in this change is not the exemption — it is that the exemption becomes a **licence keyed on a filename**. So the review focused on whether the earned condition can actually fail:

- `isPureReexportBarrel` rejects an `import`, a local `const`, an exported `function`, an exported `const`, a local `type` alias, and a re-export block that never closes. Each is a case.
- A listed file that fails it gets **both** `reexport-barrel-exemption-unearned` **and** `file-too-large` — the entry does not suppress the limit it failed to earn.
- A listed file the scan never measured is reported, so a stale entry cannot sit in review looking protective.
- An **unlisted** pure barrel is still `file-too-large`: being a barrel is not itself an exemption, someone has to say so with a reason.

Red-proofed in both halves — granting without earning fails exactly the unearned case; accepting every line as a re-export fails four purity cases.

### One defect the repository's own gates caught first

The exemption config was read at module load, so importing the scan inside the hermetic tier's stripped tree threw before anything ran — reported by the tier as a broken checkout rather than as the config coupling it was. Read lazily inside `main()` now, and the exemptions are passed into `evaluateFileSizes` so the pure function stays a function of its arguments (which also un-broke the six pre-existing cases my first cut had made fail).

### Verification at this head

- `pnpm harness:scan` — 127 passed, 3 skipped
- harness test tier — 4023 assertions green; hermetic tier exits 0 with 1113/1113
- `pnpm typecheck` — 0 errors
- `packages/agent-framework/src/index.ts` leaves the burn-down list: 82 baselined entries → 81

**ACTIONABLE FINDINGS: 0**

PR #1923, opened 2026-08-19
feat(harness): enumerate the tree a scan judges, not the git index (#1908)

### LINKED ISSUE #1908: INFRA-114: a scan's population source is a per-scan choice nobody owns, and none of them disclose what it excluded
A harness scan decides for itself where its population comes from. Five enumerate through `git ls-files`, so their coverage depends on the git index; sixty enumerate the filesystem, so theirs does not. Nothing owns that choice, nothing states it, and no scan reports what its enumeration left out.

The consequence is a false green at exactly the moment a writer checks their work. Measured in review of pull request #1886, round four — a newly written task document carried a citation `reference-kind-qualified` refuses:

- before `git add`: **passed**, printing `::examined:: 2936 tracked document(s)`
- after the file was staged: **failed**, naming that file and that line

The suite ran, declared its size, and reported success on a tree whose newest file it could not see. A human found it a review round later.

Why this is a cause rather than an incident:

- The split is 5 against 60 over one tree, and neither number is stated anywhere.
- `enforcement-architecture.md` says silence is not success, and the same suite enforces a declared `::examined::` size on its own members — yet the size an `ls-files` scan declares is silently conditional on the index.
- `createDistFreeTree` (HARNESS-047, PR #1394) already copies `ls-files --others --exclude-standard` into the CI-mirror worktree so filesystem-reading scans stay faithful — but never stages the copies, so the `ls-files` scans are blind inside the tree built for fidelity.
- Pull request #1886 adds a sixth `ls-files` scan, which is the recurrence answering itself.

Distinct from issue #1905, whose blindness comes from reverse-applying a diff in `check-regression-red-proof.mjs`.

Root item: `.agents/tasks/INFRA-113-a-scans-population-source-is-an-undisclosed-per-scan-choice.md`. Filed as FOUNDATIONAL from review of pull request #1886.

woojubb:
Closes #1908.

## The false green, measured

A harness scan decided for itself where its population came from. Eight enumerated through `git ls-files`, so their coverage depended on the git index; the rest read the filesystem, so theirs did not. Nothing owned the choice and nothing stated it.

From review of PR #1886, round four — a newly written task document carrying a citation `reference-kind-qualified` refuses:

| when | verdict |
| --- | --- |
| before `git add` | **passed**, printing `::examined:: 2936 tracked document(s)` |
| after staging | failed, naming that file and that line |

The suite ran, declared a size, and reported success on a tree whose newest file it could not see — at exactly the moment a writer checks their own work, which is who that scan is for. A human found it a review round later.

**This is worse than a scan with no coverage.** A missing scan is visibly missing; this one printed a number and a pass. `enforcement-architecture.md` says silence is not success, and a size silently conditional on the index is that defect wearing a measurement.

## The recurrence answered itself — twice

Issue #1908 notes PR #1886 adding a *sixth* `ls-files` scan while the issue was open. By the time this was picked up there were **eight**, because two of them are mine, written earlier in this same session, both copying the pattern from the scan beside them.

A per-scan choice nobody owns gets copied by whoever writes the next one, and neither of us knew a choice was being made. That is the argument for an owner rather than a fix to six call sites.

## The change

`enumerate-files.mjs` enumerates tracked files **and** untracked files git does not ignore. Both are part of the tree the author is asking about.

What stays out is only what `.gitignore` already declares. That needs no per-run disclosure, and I measured before deciding: the ignored set for `*.md` alone is **3,021 paths**, almost all under `node_modules`. A four-digit number printed on every run is a disclosure readers learn to skip, which is how a real one would be missed.

Five scans converted: `reference-kind-qualified` (where the incident was measured), `hook-override-declarations`, `aggregate-naming`, `preset-projection`, and the shared shape they use.

## Two scans deliberately NOT converted

Named here rather than left for a reader to discover:

- **`check-done-evidence`** answers about an arbitrary `root` under a git hook, with documented `GIT_DIR`/`GIT_INDEX_FILE` stripping the owner does model — but its root semantics differ.
- **`scan-legacy-typescript`** deliberately counts index entries whose files are **absent from disk**, which is a different question from "what should I judge".

Converting either without modelling those would trade a visible gap for a silent one.

The `createDistFreeTree` half of the issue is also untouched: with the owner reading untracked files directly, staging the mirror's copies is no longer what stands between those scans and the truth. Whether it should still happen is a question about the mirror's fidelity rather than about this blindness.

## Red-proofed against the real tree

Not a fixture, because the incident was a real-tree one. Writing an unstaged document with a bare `#1234` and running the scan:

- with the owner in place → `[unfrozen] .agents/tasks/ZZZ-unstaged-probe.md: 1 unqualified reference(s)`
- with untracked enumeration removed → **passed**

The measured before/after, reproduced in both directions.

## One process note, in the record

The red-proof **revert did not take**. The module is new and therefore untracked, so `git checkout --` silently did nothing and the probe stayed in the file — and the scan afterwards still passed, because the probe only affects a case the passing run does not exercise. Checking that the *revert* was applied, not just the probe, is what caught it. Same lesson as a red-proof that reports green, in the other direction.

## Verification

`pnpm harness:scan` — 129 passed, 2 skipped. `pnpm harness:pre-push` exits 0. 6 unit cases, including the exact-size assertion re-asserted after a second run.

### DIFF
--- a/.agents/tasks/INFRA-121-scan-population-source-has-no-owner.md
+++ b/.agents/tasks/INFRA-121-scan-population-source-has-no-owner.md
@@ -1,0 +1,98 @@
+---
+title: 'INFRA-121: a scan enumerated through the git index, so a file not yet staged was invisible'
+status: in-progress
+created: 2026-08-20
+priority: high
+urgency: now
+area: scripts/harness
+depends_on: []
+---
+
+# INFRA-121: enumerate the tree the author is asking about, not the index
+
+## Objective
+
+Issue #1908. A harness scan decided for itself where its population came from. Eight enumerated
+through `git ls-files`, so their coverage depended on the git index; the rest read the filesystem, so
+theirs did not. Nothing owned the choice and nothing stated it.
+
+## The false green, measured
+
+Review of pull request #1886, round four. A newly written task document carried a citation
+`reference-kind-qualified` refuses:
+
+| when             | verdict                                                      |
+| ---------------- | ------------------------------------------------------------ |
+| before `git add` | **passed**, printing `::examined:: 2936 tracked document(s)` |
+| after staging    | failed, naming that file and that line                       |
+
+The suite ran, declared its size, and reported success on a tree whose newest file it could not see —
+at exactly the moment a writer checks their own work. A human found it a review round later.
+
+That is worse than a scan with no coverage at all. A missing scan is visibly missing; this one printed
+a number and a pass. `enforcement-architecture.md` says silence is not success, and a size silently
+conditional on the index is that same defect wearing a measurement.
+
+## The recurrence answered itself, twice
+
+Issue #1908 notes that pull request #1886 adds a sixth `ls-files` scan while the issue is open. By the
+time this item was picked up there were **eight** — because two of them are mine, written earlier in
+this same session, both copying the pattern from the scan beside them. A per-scan choice nobody owns
+is copied by whoever writes the next one, and neither of us knew there was a choice being made.
+
+## Approach
+
+One owner, `enumerate-files.mjs`, enumerating tracked files AND untracked files git does not ignore.
+Both are part of the tree the author is asking about.
+
+An ignored file stays out, and that exclusion needs no per-run disclosure: `.gitignore` declares it
+in one place for every reader. Measured before deciding: the ignored set for `*.md` alone is 3,021
+paths, almost all under `node_modules`. A four-digit number printed every run is a disclosure readers
+learn to skip, which is how a real one would be missed.
+
+## Plan
+
+- [x] TC-01: an unstaged new file is enumerated.
+- [x] TC-02: the old behaviour is shown to miss exactly that file.
+- [x] TC-03: a path git reports twice is not double-counted.
+- [x] TC-04: the order is stable, so a finding list does not churn between runs.
+- [x] TC-05: the size is asserted exactly, and again after a second run.
+- [x] TC-06: five scans converted — `reference-kind-qualified`, `hook-override-declarations`,
+      `aggregate-naming`, `preset-projection` and their shared shape.
+- [x] TC-07: `pnpm harness:scan` green (129 passed, 2 skipped).
+- [ ] TC-08: `pnpm harness:pre-push` green.
+
+## Not converted, and why
+
+`check-done-evidence` and `scan-legacy-typescript` still enumerate directly. Both carry documented
+subtleties the owner does not yet model — the first answers about an arbitrary `root` under a git
+hook, the second deliberately counts index entries whose files are ABSENT from disk, which is a
+different question from "what should I judge". Converting them without modelling those would trade a
+visible gap for a silent one. Left as they are, named here rather than left for a reader to discover.
+
+The `createDistFreeTree` half of issue #1908 — the CI-mirror worktree copies untracked files but never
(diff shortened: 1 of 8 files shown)


woojubb:
## Review — feat(harness): enumerate the tree a scan judges, not the git index (#1908)

One new owner module, five scans converted, two deliberately not, and a task record.

### What I checked

- **The defect is a false green, not missing coverage**, and that distinction drives everything. A missing scan is visibly missing; this one printed `::examined:: 2936 tracked document(s)` and a pass on a tree whose newest file it could not see. A size silently conditional on the index is "silence is not success" wearing a measurement.
- **The moment it fails is the moment it matters.** `reference-kind-qualified` exists for a writer checking their own work, and before `git add` is exactly when that writer looks. The fix is not incidental to the scan's purpose; it restores it.
- **`--others --exclude-standard` is the right predicate.** Tracked ∪ untracked-not-ignored is the tree the author is asking about. An ignored file stays out because `.gitignore` declared it — a decision made once, in one place, for every reader, rather than one a scan discovers by accident of when someone last staged.
- **The choice not to disclose per-run is measured, not assumed**: 3,021 ignored paths for `*.md` alone, almost all under `node_modules`. A four-digit number on every run is a disclosure readers learn to skip, which is how a real one gets missed. I agree with the call.
- **Dedup and stable ordering** are both covered, so a path git reports twice does not double-count and a finding list does not churn between runs.

### The recurrence, which is the argument for an owner

The issue notes PR #1886 adding a *sixth* `ls-files` scan while it was open. By the time this was picked up there were **eight** — two of them written earlier in this same session, both copying the pattern from the scan beside them, neither author aware a choice was being made.

That is the case for one owner rather than six corrected call sites: the next writer copies whatever is adjacent.

### What is NOT converted, and why that is right

`check-done-evidence` (answers about an arbitrary `root` under a git hook) and `scan-legacy-typescript` (deliberately counts index entries whose files are **absent from disk** — a different question from "what should I judge"). Converting either without modelling those semantics would trade a visible gap for a silent one, which is the failure this whole change is about. Both are named in the record rather than left for a reader to find.

The `createDistFreeTree` half is also correctly left: with the owner reading untracked files directly, staging the mirror's copies is no longer what stands between those scans and the truth.

### Red-proofed against the REAL tree

Right call — the incident was a real-tree one, and a fixture would have proved the fixture. An unstaged document with a bare reference produces `[unfrozen] … 1 unqualified reference(s)` with the owner in place, and a **pass** without it. The measured before/after, reproduced both ways.

### The process note worth keeping

The red-proof **revert did not take**: the module is new and therefore untracked, so `git checkout --` silently did nothing and the probe stayed in the file — and the run afterwards still passed, because the probe only affects a case that run does not exercise. Checking that the *revert* applied, not just the probe, is what caught it. Recorded in the task file, where the next person doing this will read it.

### Harness conformance

`fixture-floor`, `harness-script-import-safety`, `measurement-provenance` (exported reader and finder, exact assertion, re-asserted after a second run) — each satisfied rather than waived. The reader and finder were renamed to the convention rather than the convention argued with.

### Verification

`pnpm harness:scan` — 129 passed, 2 skipped. `pnpm harness:pre-push` exits 0. CI clean.

REVIEWED BASE: 6e2f57e6994d4533b56462edf1b06436a0aa6e99
REVIEWED HEAD: e55e420cfc348e4e643ca7d6cf9d22f22403b0c4
ACTIONABLE FINDINGS: 0

PR #1924, opened 2026-08-19
docs(tasks): complete INFRA-120, whose work merged as PR #1921

woojubb:
Lifecycle only: one frontmatter change and one `git mv` into `completed/`.

## Why a plan item was ADDED, not just ticked

`TC-08` asked for a green pre-push, and PR #1921 is merged with CI clean — that one is a straightforward tick.

`TC-09` is new. Review of that pull request found the first cut skipping the summary `worst` update and omitting the `rank` entry, so a range whose only pair had no earlier state returned the initial `red-proof-ok`. That is **the item's own defect reproduced one level up** — an unverifiable state read as a verdict — and it was fixed in the same pull request.

A record that ticked TC-08 and said nothing would read as if the change had always covered both halves. The plan is the place a reader looks to see what was actually established, so the second half is written down as its own item rather than absorbed into the first.

## Verification

### DIFF
--- a/.agents/tasks/completed/INFRA-120-red-proof-blind-to-an-added-file.md
+++ b/.agents/tasks/completed/INFRA-120-red-proof-blind-to-an-added-file.md
@@ -1,7 +1,8 @@
 ---
 title: 'INFRA-120: the accidental-green gate cannot reach a verdict for a file added in its own range'
-status: in-progress
+status: done
 created: 2026-08-20
+completed: 2026-08-20
 priority: high
 urgency: now
 area: scripts/harness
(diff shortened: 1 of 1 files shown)


woojubb:
## Review — docs(tasks): complete INFRA-120, whose work merged as PR #1921

One frontmatter change and one `git mv`. No code.

### What I checked

- **The work is genuinely on develop.** PR #1921 is merged with CI clean, so `TC-08` is ticked against the run rather than the intent.
- **A plan item was ADDED, not just ticked**, and that is the part worth reviewing. Review of PR #1921 found the first cut skipping the summary `worst` update and omitting the `rank` entry, so a range whose only pair had no earlier state returned the initial `red-proof-ok`. That is the item's own defect — an unverifiable state read as a verdict — reproduced one level up.

  Ticking `TC-08` and saying nothing would leave a record that reads as if the change had always covered both halves. The plan is where a reader looks to see what was actually established, so the second half is written down as `TC-09` rather than absorbed into the first.

- **Frontmatter and location move together**, which `.agents/tasks/README.md` requires in one commit, and the `completed:` date is real rather than a placeholder.
- **No status text leaked into the body** — the README makes frontmatter the single source of truth and bans `## Status` sections.

### Verification

`pnpm harness:scan` — 128 passed, 3 skipped. CI clean, no failing or pending checks.

REVIEWED BASE: ea8a2d932c42eb9bdb1f5251a3d1656b195dcfdb
REVIEWED HEAD: 35cd6c4948855394e116e7ee6caa97c5e053ae68
ACTIONABLE FINDINGS: 0

Issue #1903, opened 2026-08-19
INFRA-111: "what does this redirect write to" has no owner, so each guard hand-rolls it


woojubb:
Two guards ask the same question — does this command redirect output into a path I protect? — and each answers it with its own regex over redirection spellings. Neither consumes `.claude/hooks/lib/command-scan.sh`, which already encodes the redirection grammar and exposes none of it.

`bulk-edit-guard.sh`, over its real payload protocol: `>`, `>>`, `>|`, `&>`, `2>` into `node_modules/` all exit 2, while `>&`, `>&x`, `2>&`, `>&"x"` and `>>&` all exit **0**.

`branch-guard.sh:1149` asks the same question of `.husky/` with a different regex and has a DIFFERENT hole set — it matches `>` and `&>` but misses `>&` and `>|`. That last one is a live walk-through of a gate whose own refusal says "Zero exceptions".

This is the third cut of one regex: the first shipped it, the second added `>|` and its commit claimed that was "the only miss across eight probed spellings", and the third measures five more. A hand-enumeration that certifies its own exhaustiveness and is wrong the following round is the evidence that enumeration is the wrong shape.

The grammar is already written. `command-scan.sh` (INFRA-085) parses redirections, and its own comment records the identical tuition: "`2>&1` and `>&2` put the ampersand AFTER the arrow; `&>` and `&>>` put it BEFORE. Reading only the character in front caught the first pair and missed the second." The library has already paid for the exact asymmetry that cost this hook a round, and kept the lesson private.

Not issue #1898 (flag attribution) — landing that in full leaves `>& node_modules/x` permitted.

Root item: `.agents/tasks/INFRA-110-redirect-target-has-no-owner.md`. Filed as FOUNDATIONAL from review of pull request #1886.

woojubb:
**Renumbered.** This item was originally filed with an ID that collides with `INFRA-108` already used on `develop` ("a breaking signature change reached develop because packages/*/examples is typechecked by nobody", completed 2026-08-20) — the filing session was working from a branch that predated it. The title above carries the corrected ID; the task record under `.agents/tasks/` will use the same one.

The measurements in the body are unaffected.

woojubb:
Closed by `45e371183` on `develop` (PR #1945).

[...]

Issue #1899, opened 2026-08-19
INFRA-110: the bulk-edit guard hand-rolls path canonicalization, and it fails open


woojubb:
`bulk-edit-guard.sh` answers "where does this write actually land?" with `resolved_through_existing_ancestor` — twelve lines of bash that climb with `dirname`, resolve with `cd` + `pwd -P`, and re-attach the segments that do not exist yet. It has no stated domain, so its correctness is defined by whichever input shapes a reviewer happened to try. Three review rounds of pull request #1886 have each found a new one.

Measured in a sandbox where `app/vendored -> ../node_modules/pkg` and `app/filelink.ts -> ../node_modules/pkg/src/index.ts`:

| write target | exit | should be |
| --- | --- | --- |
| `app/vendored/src/new.ts` | 2 | 2 |
| `app/filelink.ts` | 0 | 2 |
| `app/nonexistent/../vendored/src/x.ts` | 0 | 2 |
| `app/../app/vendored/src/x.ts` | 2 | 2 |
| `app/vendored/src/new.ts`, `CDPATH` set | 0 | 2 |

- `cd` honours an exported `CDPATH`, so the resolution silently selects a different directory — and `cd` prints it to stdout, corrupting `RESOLVED`. This is a fail-OPEN, against the refuse direction the file declares in its own header.
- A `..` after a missing segment is re-attached verbatim and never normalised.
- The final component is never resolved, only its ancestors, so a symlinked FILE into the store is permitted while the sibling symlinked directory is refused.

Each hole is a one-line patch, and patching all three produces a hand-rolled realpath with four special cases and still no stated domain. The three rounds already spent are the evidence.

Root item: `.agents/tasks/INFRA-109-hand-rolled-path-canonicalization-in-the-bulk-edit-guard.md`.
Filed as a FOUNDATIONAL finding from review of pull request #1886, which ships the guard with the three holes labelled at the site.

woojubb:
**Renumbered.** This item was originally filed with an ID that collides with `INFRA-108` already used on `develop` ("a breaking signature change reached develop because packages/*/examples is typechecked by nobody", completed 2026-08-20) — the filing session was working from a branch that predated it. The title above carries the corrected ID; the task record under `.agents/tasks/` will use the same one.

The measurements in the body are unaffected.

woojubb:
Closed by `a7a8a2f20` on `develop` (PR #1947).

[...]

PR #1925, opened 2026-08-19
fix(harness): let one qualifier govern a whole reference phrase (#1913)

### LINKED ISSUE #1913: INFRA-116: a reference RANGE is qualified only at its first member

woojubb:
Closes #1913.

## Wider than the issue described

The issue names ranges. Reproduced on the current model first, and **three** forms have the same defect — every one is a phrase where a single qualifier names every member:

| line | before | after |
| --- | --- | --- |
| `Measured over PRs #1525–#1530: twelve runs` | flagged `#1530` | clean |
| `issues #10-#12 are open` | flagged `#12` | clean |
| `claimed by issues #1899, #1903, #1904` | flagged `#1903`, `#1904` | clean |
| `pull requests #10 and #12` | flagged `#12` | clean |
| `See PR #1525 and PR #1530` | clean | clean |

**The list form is one I hit twice this session** while working other items — once quoting a range verbatim out of another record, once writing a list of issue numbers. Both times the fix available at the site was wrong in exactly the way the issue describes: alter a verbatim quotation to satisfy a check mistaken about the text, or mark a real claim as a code-span "specimen", an exemption meant for something else.

## The separators are enumerated, not "any short punctuation"

The point is to recognise a **phrase**, not to forgive **proximity**.

`PR #10 broke #12` stays two references: a verb between them means the second is a different claim, and treating nearness as governance would let a real bare reference hide behind an unrelated qualified one on the same line. A new sentence breaks it too, and so does an intervening word.

Chaining also **carries a qualifier forward and never invents one** — `See #10 and #12` is still two findings.

Both dash forms are listed. The en dash is what this repository's prose actually uses, and a check that only knew the hyphen would be right about the form nobody writes.

## Red-proofed in BOTH directions, because this rule can be wrong two ways

Each way passes the other's cases, so neither probe alone is evidence:

| probe | what went red |
| --- | --- |
| chaining disabled entirely | the four accept-cases |
| chaining widened to any short run | the three refuse-cases |

A rule that only accepts is a rule that accepts everything. Both probes were reverted, and each revert was verified by re-reading the file rather than assumed.

## The baseline drop was read, not counted

Re-frozen in this same change: **1,520 → 1,496** unqualified references across **292 → 281** files, with nothing grown or unfrozen.

I checked a sample rather than trusting the direction. `content/guide/cli.md` fell by one, and the instance is `issues #22732, #3045` — a genuine list under one qualifier, which is the form this change is for rather than a coincidence of the widening.

## Verification

`pnpm harness:scan` — 128 passed, 3 skipped. `pnpm harness:pre-push` exits 0. 40 cases in the reference-kind suite, 8 of them new.


woojubb:
## Review — fix(harness): let one qualifier govern a whole reference phrase (#1913)

One chaining rule in the reference model, the rule section that declares it, eight new cases, and a re-frozen baseline.

### What I checked

- **The scope was widened by reproduction, not by guessing.** The issue names ranges; running the current model first showed comma lists and `A and B` pairs have the same defect. A fix written from the issue's framing would have left two of three forms broken.
- **The separators are enumerated, and that is the load-bearing decision.** The point is to recognise a *phrase*, not to forgive *proximity*. `PR #10 broke #12` staying two references is what keeps a real bare reference from hiding behind an unrelated qualified one on the same line — and the refuse-cases pin exactly that.
- **Chaining carries a qualifier forward and never invents one.** `See #10 and #12` is still two findings, which is the property a naive "everything after a `#N` is fine" rule would lose.
- **Both dash forms are handled.** The en dash is what this repository's prose actually uses; a hyphen-only check would have been right about the form nobody writes.
- **The `previous` tracking does not leak across a specimen.** A reference inside a code span records `governed: false`, so a qualified specimen cannot govern a real bare reference that follows it.

### Red-proofing — both directions, and here that is required rather than thorough

A chaining rule can be wrong two ways, and **each way passes the other's cases**:

| probe | what went red |
| --- | --- |
| chaining disabled entirely | the four accept-cases |
| chaining widened to any short run | the three refuse-cases |

Neither probe alone is evidence. A rule that only accepts is a rule that accepts everything, and a rule that only refuses is the bug being fixed. Both probes were reverted and each revert verified by re-reading the file — this session has already had a probe survive a `git checkout --` that silently did nothing.

### The baseline drop was read, not counted

1,520 → 1,496 across 292 → 281 files, nothing grown or unfrozen. A sample was opened rather than the direction trusted: `content/guide/cli.md` fell by one and the instance is `issues #22732, #3045` — a genuine list under one qualifier, which is the form this change is for rather than a coincidence of the widening. That check matters because a widening always reduces the count; the question is whether it reduced it for the right reason.

### Provenance worth noting

The list form is one the author hit twice in this same session, and both times the only site-level fixes available were wrong — alter a verbatim quotation, or mark a real claim as a code-span specimen. That is the issue's own argument for fixing the rule instead, demonstrated rather than asserted.

### Verification

`pnpm harness:scan` — 128 passed, 3 skipped. `pnpm harness:pre-push` exits 0. 40 cases, 8 new. CI clean.

REVIEWED BASE: 086854484df4b43641cbb420e7bfd1e8e4e6509b
REVIEWED HEAD: af212f4319cd741f82197808e1fe749bef4d3556
ACTIONABLE FINDINGS: 0

PR #1926, opened 2026-08-19
docs(tasks): complete INFRA-121, whose work merged as PR #1923

woojubb:
Lifecycle only: one frontmatter change and one `git mv` into `completed/`.

`TC-08` asked for a green pre-push; PR #1923 is merged with CI clean, so the item is ticked against the run rather than the intent.

## What stays in the record

The two scans INFRA-121 deliberately did **not** convert — `check-done-evidence` and `scan-legacy-typescript`, each with its reason — remain in the completed file. That is the part a future reader needs: a completed item that looks total invites the next person to assume the whole class is handled, and here it is not.

The `createDistFreeTree` half of issue #1908 is recorded the same way, for the same reason.

## Verification

### DIFF
--- a/.agents/tasks/completed/INFRA-121-scan-population-source-has-no-owner.md
+++ b/.agents/tasks/completed/INFRA-121-scan-population-source-has-no-owner.md
@@ -1,7 +1,8 @@
 ---
 title: 'INFRA-121: a scan enumerated through the git index, so a file not yet staged was invisible'
-status: in-progress
+status: done
 created: 2026-08-20
+completed: 2026-08-20
 priority: high
 urgency: now
 area: scripts/harness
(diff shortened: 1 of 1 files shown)


woojubb:
## Review — docs(tasks): complete INFRA-121, whose work merged as PR #1923

One frontmatter change and one `git mv`. No code.

### What I checked

- **The work is on develop.** PR #1923 is merged with CI clean, so `TC-08` is ticked against the run rather than the intent.
- **The record keeps what it did NOT do.** `check-done-evidence` and `scan-legacy-typescript` stayed unconverted, each with its reason, and both survive into `completed/`. That is the part a future reader needs: a completed item that reads as total invites the next person to assume the whole class is handled, and here two scans still enumerate the old way for reasons that were deliberate rather than overlooked.
- **The `createDistFreeTree` half of issue #1908 is recorded the same way** — untouched, with the reasoning, so the silence is not read as an oversight.
- **Lifecycle mechanics are right**: frontmatter and location move in one commit as `.agents/tasks/README.md` requires, `completed:` is a real date, and no status text leaked into the body.

### Verification

`pnpm harness:scan` — 128 passed, 3 skipped. CI clean, no failing or pending checks.

REVIEWED BASE: 086854484df4b43641cbb420e7bfd1e8e4e6509b
REVIEWED HEAD: f2c5f9ff7b3f1773ae0d82ca6233b5096240b146
ACTIONABLE FINDINGS: 0

Issue #1929, opened 2026-08-20
Three provider-settings tests read the developer's real ~/.robota/settings.json, so "no settings anywhere" is false on a configured machine


woojubb:
## What

Three tests fail on any machine that has a real robota profile, and pass on CI:

- `packages/agent-framework/src/command-api/provider/__tests__/env-default-provider.test.ts` — `TC-01: no settings anywhere + env key set → env-default config`
- the same file — `TC-04: no profile + no recognized env key → ProviderConfigError unchanged`
- `packages/agent-framework/src/command-api/provider/__tests__/provider-factory.test.ts` — `TC-04: throws ProviderConfigError when no provider configuration exists`

Reproduced on `develop` at 6e2f57e69 with an unrelated branch stashed, so it is not caused by any work in flight:

```
AssertionError: expected 'gemma' to be 'anthropic'
AssertionError: expected function to throw an error, but it didn't
```

## Cause

Each case isolates the **project** directory — `mkdtempSync(join(tmpdir(), 'robota-env-default-'))` — and injects `env`. But `readProviderSettings` also reads the **user-level** settings at `~/.robota/settings.json`, and nothing in the test isolates that.

So the premise "no settings anywhere" is only true when the developer has never configured the CLI. On this machine `~/.robota/settings.json` holds a `gemma` profile, and the user profile wins over the env key — which is correct behaviour (`TC-03: settings profile wins over the env key` asserts exactly that) reported as a failure.

## Why it is worth filing rather than fixing in passing

The failure direction is the harmless one. The other direction is not: a case whose premise is "no configuration exists" goes GREEN for the wrong reason on any machine whose real profile happens to match what the case expects. It is the same shape as a guard whose input cannot vary — the result stops depending on the code under test and starts depending on the host.

It also makes the pre-push gate un-passable on a configured developer machine while CI stays green, which pushes people toward the override.

## Fix direction

Isolate the user-settings root the same way the project root is already isolated — the reader should take it from an injected path (as it already takes `env`), and the cases should point it at a temp directory. Fixing it by deleting the developer's profile, or by making the assertion tolerant of whatever is there, would both preserve the defect.

## Found by

The pre-push full-suite run on an unrelated ARCH branch, on a machine that had been configured earlier the same day.

woojubb:
Hit this independently from another session and filed it as #1932 before finding this one. Closing mine as the duplicate; carrying over the three things it had that this one does not.

### The trigger is the product's own onboarding

[...]

Issue #1820, opened 2026-08-17
ARCH-013 stage 2 remainder: ten resolved-preset fields need a projection decision, not just a seam


woojubb:
Filed by ARCH-013 stage 2, which landed the measurement. `scripts/harness/scan-preset-projection.mjs`
now checks that every `IResolvedPresetOptions` field appears in a **declared** projection — a surface
interface member or a `Pick` of the source — and that the two projection surfaces do not diverge.

Run `node scripts/harness/scan-preset-projection.mjs` with
`presetProjection.pendingProjection` emptied to reproduce all ten findings.

## Why these are not in the same change as the scan

Each of the ten needs a decision with user-visible consequences that the scan cannot make and the
task file does not answer. They are recorded as **named, expiring exemptions** in
`.agents/harness.config.json` → `presetProjection.pendingProjection`, each with its own reason, so
the list is visible in review and a stale entry is reported as `preset-exemption-unused`.

## The decisions, grouped

**1. Model group — `temperature`, `maxOutputTokens`.** Applied by the live `/preset` path through
`applyModelOptions`, applied by nothing at startup. There is no `ICreateSessionOptions` seam for
either (all 54 keys enumerated), so these need a route to the provider config — a new seam, not a
re-declaration. Same class as `effort` before stage 1 fixed it.

**2. `model` — hand-mapped, not dropped.** `cli.ts:262` computes
`resolvedPreset.model ?? providerSettings.model` and threads it to all three construction sites, so
the value does reach the session. What is missing is the declared projection. Moving it onto
`IPresetSurfaceOptions` requires deciding where the `?? providerSettings.model` fallback lives,
because the surface is built at `cli.ts:330` and the model id is first needed at `:262`.

**3. `agentName` — the reverse divergence, previously unnamed.** Startup declares it; the live
`/preset` path does not. Starting with a preset sets the agent name; switching to the *same* preset
mid-session leaves the old one. **Decision: should `/preset` rename the agent mid-session?**

**4. Prompt group — `systemPrompt`, `appendSystemPrompt`.** `appendSystemPrompt` has a session seam
already forwarded by `create-session-projection.ts`, so only the preset→surface hop is missing — but
wiring it needs a merge order against the CLI-sourced append text. `systemPrompt` has no seam at all;
the framework composes prompts via `systemPromptBuilder`, so a preset-supplied full prompt must
either replace or seed that. **Both are user-visible semantics.**

[...]

PR #1931, opened 2026-08-20
refactor(preset): ARCH-040 groups A and E, one projection declaration (ARCH-041), and issue #1929

woojubb:
Progress on #1820 (ARCH-040), plus the cause found underneath it and one unrelated blocker fixed on the way.

## ARCH-041 — the projection was written THREE times, and had already drifted

`IPresetSurfaceOptions` says of itself that "adding a field here now reaches every surface at once, which is the only property that makes this worth extracting". It did not hold: `print-mode.ts` and `serve-mode.ts` each declared their own copy, and `model` was on print mode's copy and on **neither** of the others — the exact shape ARCH-013 was filed about, surviving the extraction meant to end it. The projection scan's configured surfaces saw two of the three, so nothing reported it.

Both mode types are `Partial<IPresetSurfaceOptions>` now, and `model` moved onto the shared type.

It earned its keep immediately: the very next change (Group E) added two fields to the shared type, and the `Required<>` case caught the fixture instantly. Under three copies, those two fields would have gone silently missing from two surfaces.

## ARCH-040 Group A — `defaultTrustLevel` removed

Owner decision. The reason is stronger than "nothing read it", and worth stating because a removal justified by a grep is one the next person re-adds: a preset already states its posture **three** ways — `permissionMode`, `defaultPermissionMode`, `autonomy` — and `resolvePreset` promotes the last two into the first. A fourth spelling is a second answer to one question, which is what the projection scan calls `derivationOnly` for the other two. The trust axis is untouched: `config.defaultTrustLevel` still reaches the session.

## ARCH-040 Group E — the model group reaches the session at construction

`temperature` and `maxOutputTokens` were applied by the live `/preset` path through `applyModelOptions` and by **nothing** at startup — one session holding two answers for the same preset depending on when it was chosen, exactly what `effort` did before ARCH-013 stage 1. They now travel the chain `effort` travels and land on the same two agent channels `applyModelOptions` writes.

`option-reachability` caught the first cut declaring the keys with nothing assigning them; the init hop was added rather than the declaration removed.

## Issue #1929 — three tests read the developer's real settings

Found by this branch's own pre-push run and **reproduced on `develop` with the branch stashed**, so it predates the work. Three cases whose premise is "no provider configuration exists" isolate the project directory and inject `env`, but the default settings list also reaches the operator's home directory. `readProviderSettings` gains a `settingsPaths` seam — the same shape as the `env` seam directly above it, for the same reason.

The failure direction was the harmless one. The other is a case going GREEN because the host's profile happened to match what it expected.

## What is NOT in here, and why

- **Group B (`agentName`)** — parked. Its only reader is a `readonly` identity label no surface displays; renaming needs a mutable core field or a split of a 411-line frozen file. Reverted rather than paid silently; the measurement is in the task record, awaiting a re-decision.
- **Group C (tools)** — the combine rule is decided (allow replaces, deny unions) but the live half needs a capability that does not exist: the lists become permission PATTERNS at construction and the enforcer exposes only an additive session-allowed set. Written, measured against the scan, and **reverted** — wiring the startup half alone CREATES the divergence the scan exists to measure.
- **Groups D and F** — feasible on existing seams, not yet done.

`presetProjection.pendingProjection`: 10 exemptions → 6.

## Verification

- `pnpm typecheck` — 0 errors
- full suite — exits 0 (it did not before the #1929 fix, on this machine)
- every change red-proofed; the parked/reverted ones carry their measurement instead of code

### DIFF
--- a/.agents/harness.config.json
+++ b/.agents/harness.config.json
@@ -425,11 +425,6 @@
     "$pendingComment": "The burn-down. Each entry is a field this floor HAS measured as undeclared or one-sidedly declared, whose fix needs a decision this scan cannot make — recorded as a NAMED, EXPIRING exemption rather than an opaque count, because an exemption that is unnamed or non-expiring is indistinguishable from no floor. `declaredOn` records the surfaces the field was measured on when the entry was written; if the live state differs in EITHER direction the entry is reported (`preset-pending-state-changed`), so an exemption expires when the work is DONE and not only when the field is deleted. Review measured the earlier version failing exactly that: fixing a pending field printed green, which made the list a baseline with extra words. This list may shrink and must never grow without a reviewed reason. Tracked as GitHub issue #1820.",
     "pendingProjection": [
       {
-        "field": "model",
-        "reason": "NOT dropped — hand-mapped. `cli.ts:262` computes `resolvedPreset.model ?? providerSettings.model` and threads the result into all three construction sites, so the value reaches the session; what is missing is the DECLARED projection. Resolving it means moving the field onto IPresetSurfaceOptions, which requires deciding where the `?? providerSettings.model` fallback lives, because the surface is built after the model id is first needed (cli.ts:330 vs :262).",
-        "declaredOn": ["IPresetApplicationOptions"]
-      },
-      {
         "field": "temperature",
         "reason": "Applied by the live /preset path via `applyModelOptions` (preset-application.ts), and by nothing at startup. Unlike `model` there is no `ICreateSessionOptions` seam at all — the value would have to reach the provider config, so this is a new seam rather than a re-declaration. Same class as `effort` before stage 1.",
         "declaredOn": ["IPresetApplicationOptions"]
--- a/.agents/harness.config.json
+++ b/.agents/harness.config.json
@@ -460,11 +460,6 @@
         "declaredOn": []
       },
       {
-        "field": "defaultTrustLevel",
-        "reason": "Measured fully dead by the ARCH-013 audit: validated by preset-validation and then read by nothing, with no seam anywhere. Removing an unconsumed public contract is a product decision rather than a grep result, so it stays declared and exempted until its owner decides between wiring it to the permission/trust axis and dropping it.",
-        "declaredOn": []
-      },
-      {
         "field": "allowedTools",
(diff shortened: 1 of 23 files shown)


woojubb:
## Pre-merge review

**base:** `4df2a03133af137f44c95156ebb42e43fc773d61`
**head:** `8b0e6ace12b0eb19b1f8bcbaa54acf63d577f70b`

Reviewed the whole diff at this head. CI is green and no review threads were opened.

### What I checked hardest, and why

Three of the four changes REMOVE or REDIRECT something, so the risk is a claim that outruns the evidence:

- **`defaultTrustLevel` removal** — verified the field is read by nothing (`config.defaultTrustLevel` is a different value on a different type and still reaches the session), and that `resolvePreset` promotes `defaultPermissionMode` and `autonomy` into `permissionMode` at `resolve-preset.ts`, which is what makes a fourth spelling a second answer rather than merely an unused one. Red-proofed by carrying the field through the validator again.
- **One projection declaration** — the enforcement is the TYPECHECK, and the test file says so rather than implying its assignment case is a runtime gate. Measured: re-declaring one mode's copy gives 12 `tsgo` errors and 3 GREEN vitest cases. Calling that case "the enforcement" would have been the defect this repository scans for, and the first draft did.
- **Model group** — asserted on the agent CONFIG the session builds, not on the session options, because the config hop is the one that was missing. A third case covers the absent direction, since an explicit `undefined` is a value the provider request builder carries and an absent key is not.
- **#1929** — reproduced on `develop` with this branch stashed before touching it, so "predates the work" is measured rather than assumed.

### Two gates caught defects in this change, and both were fixed rather than worked around

- `option-reachability` reported the model-group keys declared on `IInitOptions` with nothing assigning them. The init hop was added; the declaration was not deleted.
- `product-identity` reported the neutral library naming the product in a new comment. Reworded.

A third is worth recording: the branch guard refused a bypass of the failing test suite with "if a check is wrong, unrunnable, or fires on correct work, change the CHECK". That is what produced the #1929 fix instead of an override.

### What is deliberately absent

Group B is reverted and parked with its measured cost; Group C is written, measured against the projection scan, and reverted because wiring its startup half alone CREATES the divergence that scan exists to measure. Both carry their measurement in the task record instead of half-applied code.

**ACTIONABLE FINDINGS: 0**

Issue #1919, opened 2026-08-19
INFRA-123: nothing can name the language of an embedded payload, so a language-scoped rule has no reachable subject


woojubb:
`from glob import glob; glob(…)` and `import glob as g; g.glob(…)` are the symlink-following enumerator wearing an import the call site does not spell. Reading that import requires knowing whose language a line is written in, and neither half of the rule can answer it for the place python actually lives in this repository — inside a payload.

**The hook cannot attribute to an interpreter.** Every reader it has expands an interpreter payload, so once `python3 -c "…"` is expanded the payload's own `;`, `|` and `&` are indistinguishable from the shell's; `hook_statement_ranges` splits inside a payload for the same reason. Three cuts were measured and each refused a correct command: a whole-command conjunction, a nearest-interpreter walk with a hand-written reset list, and a separator reset.

**The scan can only attribute to a FILE.** Scoped to a file that IS python, the rule judges an empty population — measured: **zero** tracked `.py` files and **zero** python-shebang files, against **fifteen** files containing `python3 -c`, all of them `.sh`, `.mjs`, `.md` or `.yml`. It would enforce nothing while the rule table claimed it did. Unscoped, it reports `import glob from 'glob'` — a package this repository depends on that does not follow — which is refusing the safe sibling.

The unit that has a language is neither the command nor the file: it is the payload — a `-c` argument, a heredoc body, a workflow `run:` block. Nothing in either enforcer can name one.

The import widening was therefore withdrawn rather than shipped unenforced; the follow-up to pull request #1886 keeps the call spelling, which needs no subject.

Root item: `.agents/tasks/INFRA-117-nothing-can-name-the-language-of-an-embedded-payload.md`. Related to issue #1898, which owns the command-side half of the same missing reader.

woojubb:
## The three measured failures share an assumption the problem does not require

All three cuts named here — the whole-command conjunction, the nearest-interpreter walk, the separator reset — operate on a command **after expansion**:

> Every reader it has EXPANDS an interpreter payload, so once `python3 -c "…"` is expanded the payload's own `;`, `|` and `&` are indistinguishable from the shell's.

That is true, and it is a property of *splitting an expansion*, not of the question. A payload's extent is delimited in the RAW text before anything expands it: by the quote that opens after `-c`, or by the heredoc delimiter that names its own terminator. Lexing that text never produces the ambiguity, because it never produces the expansion.

I probed it rather than argued it:

```js
[...]

Issue #1898, opened 2026-08-19
INFRA-109: flag attribution has two implementations, and the two enforcers disagree


woojubb:
The bulk-edit guard hook and the `symlink-following-enumeration` scan are the two halves of one rule, and both must answer "is this hazardous flag being passed to THIS command, in THIS shell statement?". They answer it with two independent implementations, and only the hook's was corrected.

Measured over the same commands, driving the scan's `findingsIn` and the hook's real stdin protocol:

| command | hook | scan |
| --- | --- | --- |
| `rg -l foo packages \| xargs grep -L bar` | permits | flags as `rg --follow` |
| `find packages -L` | refuses | clean |
| `find packages -name '*.ts' -L` | refuses | clean |

Three probed commands, three disagreements — one false positive and two false negatives.

The hook has a tokenizer (`segments()` + `cmd_flag()`, splitting at `|`, `;`, `&&`, `||`) because this exact conflation was reported twice in review of pull request #1886. The scan has none: each of its four rules hand-rolls its own bound on what may stand between a command and its flag, and the three that have one are all different. Correcting the `rg` bound alone leaves `find`'s false negatives and guarantees a fifth ad-hoc bound at the next spelling added.

Root item: `.agents/tasks/INFRA-108-flag-attribution-has-two-implementations.md`.
Filed as a FOUNDATIONAL finding from review of pull request #1886, which ships both halves with the disagreement labelled at each site.

woojubb:
A later review round of pull request #1886 measured three more disagreements and **corrected this item's premise**, so the scope here is wider than when it was filed.

The original item assumed the hook's `segments()` was the good half, to be promoted to both enforcers. It is not — promoting it would carry three further defects into the scan:

- **A newline does not separate.** `hook_statement_all_words` emits `&&`, `|` and `;` as words and drops the line break, so two statements on two lines merge into one segment. Measured: `sed -i … && ls node_modules/.bin` is permitted, the identical pair on two lines is refused. Adding `"\n"` to `is_sep` is inert — the token never arrives. `hook_statement_ranges` already splits it correctly.
- **The command word is compared exactly**, so `/usr/bin/find -L packages …` is unattributed while `find -L packages …` is refused — and the scan flags the line the hook permits. Five sites are wrong at once. `invokesWith` in `scan-shell-portability.mjs` already answers it with `w.endsWith('/' + command)`.
[...]

PR #1933, opened 2026-08-20
fix(harness): four gaps the bulk-edit guard still had after it landed

woojubb:
Follow-up to #1886, which landed the bulk-edit guard. Four gaps survived that merge; each is measured against the code on `develop` and red-proofed alone.

## What this fixes

| behaviour | `develop` | here |
| --- | --- | --- |
| `echo x >\| node_modules/y` | permitted | refused |
| `Write` into the store with `BULK_EDIT_ACK=1` **exported** | refused — the escape the refusal advertises cannot be taken | permitted |
| scan sees `.husky/pre-commit`, `pre-push`, `commit-msg` | invisible | judged |
| `::examined::` versus the judged population | over by 4 | equal |

- **The noclobber redirect.** `>|` was the only miss across eight probed spellings.
- **The acknowledgement a file-writing tool can take.** A tool write runs no command, so the exported form is the only one it can carry, and the check sat below that branch's `exit 0`. It now sits in each branch *below that branch's own decode refusal* — above them, an unreadable payload became a permitted one, which is the direction the file's header declares against. The message names both forms and says the exported one stays armed until unset.
- **The scan's population.** Gating on an extension list made this repository's own git hooks invisible to a rule whose text claimed committed scripts without qualification. A shebang now decides, listing exactly the interpreters the extension set also admits so the two filters cannot disagree about one population.
- **The declared size, and the reader contract.** The allowlist is applied before the counter, so the size and the judged population are the same number. The injected reader's two-argument contract moved onto `scanTrackedFiles`, and a strict read that comes back empty now refuses instead of being skipped.

The population also **delegates to `enumerate-files`** rather than calling the index itself — one of the copies that owner exists to replace. It picks up untracked files as a result (3195 → 3207), which closes the incident behind INFRA-121: a file written and not yet staged is now part of the tree its author is asking about.

## What it deliberately does not fix

Five causes are filed and labelled at their sites rather than patched, under [finding-depth.md](.agents/rules/finding-depth.md). Two reach past this change into `develop`:

- **INFRA-109** (#1898) — flag attribution has two implementations; the hook and the scan disagree on six probed shapes.
- **INFRA-110** (#1899) — the write-path resolution hand-rolls canonicalization: `CDPATH` fails open, `..` past a missing segment is unnormalised, a symlinked *file* into the store is permitted.
- **INFRA-111** (#1903) — no owner for "what does this redirect write to": the `>&` family walks through, and `branch-guard.sh` hand-rolls the same question with a *different* hole set.
- **INFRA-115** (#1912) — `dash`/`ksh`/`ash` are admitted by shebang with no matching extension, so one script is judged under one filename and clean under another.
- **INFRA-123** (#1919) — records a **withdrawal**, not a defect. Reading python's import spelling at the command needs to know whose language a line is in; every reader this hook has expands an interpreter payload, so the payload's own separators become the shell's. Three cuts were tried and each refused a correct command, so the widening was withdrawn rather than shipped unenforced. This repository has zero tracked `.py` files, so a file-scoped rule would have judged an empty population while the rule table claimed it was enforced.

Four further causes this work filed are **absent here**: they were picked up and finished on `develop` while this change was in review — two of them (`enumerate-files.mjs`, `scan-hook-override-declarations.mjs`) citing these issue numbers in their own headers. Their records live there, and the duplicates are withdrawn.

## Verification

Six local review rounds; findings per round 4, 4, 5, 4, 2, 2 → 0. Every fix red-proofed individually — reverting its own line fails exactly its own case, and no more.

`pnpm harness:scan` — 127 passed, 2 skipped. Harness tier green.

Four rounds found a defect the previous round's fix had introduced (a permissive path, a silent coverage loss, a payload separator ending attribution, then the same bug on three other characters). Each is recorded in the commit that corrected it.

### DIFF
--- a/.agents/rules/operational.md
+++ b/.agents/rules/operational.md
@@ -144,14 +144,31 @@ Four spellings follow symlinks, and each has a sibling that does not:
 | python `glob.glob` / `glob.iglob` | `pathlib.Path(...).rglob`        |
 
 Shell `**` under `globstar`, zsh `**`, and Node's `fs.globSync` do not traverse symlinked
-directories and are unrestricted. `bulk-edit-guard.sh` refuses the four at the command,
-`scan-symlink-following-enumeration` refuses them in a committed script, and both refuse a write
-whose path resolves inside the store. The declared exception is `BULK_EDIT_ACK=1` inline, after
-checking by hand where the enumeration reaches.
+directories and are unrestricted. `bulk-edit-guard.sh` refuses the four at the command and
+`scan-symlink-following-enumeration` refuses them in a committed script. Only the hook resolves a
+path, and only for a file-writing tool's own target: a redirect and an in-place editor are judged on
+how the path is SPELLED, and the scan resolves nothing.
+
+The declared exception is `BULK_EDIT_ACK=1`, after checking by hand where the enumeration reaches —
+INLINE in the same command, or EXPORTED, which is the only form a file-writing tool can carry
+because such a tool runs no command to put an assignment in front of. They differ in LIFETIME as well
+as spelling: the inline form is scoped to the one command that carries it, and the exported one holds
+for the rest of the session and covers the file-writing tools too. Prefer the inline form. Neither
+reaches above the guard's payload refusals: an ack says a write was checked, not that an unreadable
+payload was.
 
 Enforced by: `bulk-edit-guard` (the command) and `symlink-following-enumeration` (the committed
-script). Neither reaches a python program passed through a heredoc, nor a two-step edit that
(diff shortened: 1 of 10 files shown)


woojubb:
**Finding depth — disposition: `containment`**

A foundational finding was judged on this PR. It lands under a labelled containment hold naming the root item below; the root item is what resolves it.

Root item(s): INFRA-109, INFRA-110, INFRA-111, INFRA-115, INFRA-123
Recorded from `fix/bulk-edit-guard-followup` at `bcbcf72ca`.

PR #1935, opened 2026-08-20
feat(agent-transport-tui): say who drove the turn, not always "You:"

### LINKED ISSUE #1915: agent-transport-tui: a peer message is indistinguishable from one the operator typed (follow-up to issue #1863)
Split out of issue #1863 so that issue can close on the work it specified. PEER-006 (PR #1910) landed the last of the message flow: two local sessions now exchange messages in both directions, and a peer turn is submitted with `turnSource: 'peer'` and a driver id derived from the sender's session id.

What is **not** delivered is the other half of issue #1863's definition of done: *"the receiving operator sees the peer origin rendered."*

## Measured on develop after PR #1910

| probe | result |
| --- | --- |
| `grep -rn "driverId\|turnSource" packages/agent-transport-tui/src` | **no rendering site.** The only hit is `useSideEffects.ts:46`, which *filters* on `requesterDriverId !== OWNER_DRIVER_ID` to decide whether an intent belongs to this surface |
| `IHistoryEntry` (`packages/agent-core/src/interfaces/messages.ts:124`) | `id`, `timestamp`, `category`, `type`, `data` — **no driver or origin field** |

So a message that arrives from another session appears in the transcript with nothing distinguishing it from something the operator typed. The attribution exists and is correct at the runtime layer — the turn carries `peer:<session-id>` — and stops before anything a person can see.

## Why this is a real gap rather than cosmetics

Issue #1809 fixed that the driver id must be derived rather than peer-supplied, on the grounds that **"a name the transcript's reader trusts must not be chosen by the party being named."** That reasoning only pays off if the transcript's reader is shown the name. Deriving an attribution nobody displays protects a property no one can observe.

It also matters for a person deciding what to do: a request that came from their other session and one they typed themselves carry different authority, and right now the transcript cannot tell them apart.

## Scope

1. Carry the driver id (or a projection of it) far enough that the TUI can read it for a submitted turn. `IHistoryEntry` is the obvious candidate but is a shared contract — whether it gains a field, or the projection happens elsewhere, is the decision this issue has to make rather than assume.
2. Render it. Distinguishable at a glance from an owner turn, and naming *which* session, since more than two can be announced.
3. Do not render the peer's own `IPeerOrigin.driverId`. That field is peer-supplied and is deliberately discarded at ingress; displaying it would undo issue #1809's fix at the one place it was meant to protect.

## Definition of done

Two local `agent-cli` sessions exchange messages, and the receiving operator can tell from the transcript alone that a message came from another session and which one — verified by running it, not by asserting the field is set.

## Prior work, for the reader

- PEER-004 (PR #1897) — discovery and `/peers`
- PEER-005 (PR #1906) — the unix-socket carrier
- PEER-006 (PR #1910) — the join, `/peers send`, and the derived attribution. Its task record `.agents/tasks/PEER-006-*.md` leaves TC-07 explicitly unchecked for this reason.

woojubb:
## The defect

Every user message rendered `You:`, so a peer session's message was indistinguishable from one the operator typed. The transcript asserted something false about who spoke — which is the whole point of co-driving being visible.

The driver was already resolved per turn. It stopped one hop short of the stored message: `interactive-session-prompt.ts` called `session.run(modelInput, hookInput, options)` where `options` carried only `ephemeralSystemContext`.

## The change

`IRunOptions.driverId` carries it to the conversation store, and `RoleLabel` prints it.

The OWNER is **elided** rather than stamped on every message. Carrying the constant would mark every stored message and change the `run()` call shape for the dominant path to say nothing new — every surface already reads an unattributed message as the operator's. The first version did stamp it, and eleven `agent-framework` tests went red on the changed call shape, which is what surfaced the choice.

Display attribution only. The id is the framework's **derived** `peer:<session-id>`, never anything the peer supplied, and nothing consults it to decide what a run may do (issue #1809).

## Verified by running it, not by asserting metadata

Two real `pnpm cli:dev` sessions, driven through PTYs on one host as one user.

```
A alone:   No other live session is announced. Start a second session on this host, as this user…

B:         Live sessions:
             b8319b98-bb7b-486c-a499-cf1585b39e61  (this session)
             97ffafe3-f770-4877-90a4-bd86df6be010

/peers send 97ffafe3-… hello from B  →  what A rendered:

           peer:b8319b98-bb7b-486c-a499-cf1585b39e61:
             hello from B
           Robota:
             STUB-ACK: …
```

The label carries B's session id exactly as `/peers` reported it, and the agent answered the peer's turn.

**The control run is what makes this mean anything.** A separate session where the operator typed the message rendered `You:`. Across the two transcripts the labels are mutually exclusive — the peer-driven one contains `peer:<id>:` and **zero** occurrences of `You:`; the operator one contains `You:` and **zero** occurrences of `peer:`. Without that second run the first only shows that a label appears, not that the two are told apart.

No provider credential was present (probed: no `ANTHROPIC_*`/`OPENAI_*`/`GEMINI_*` in the environment, no `.env` — only `.env.example`, nothing under `~/.robota`, no Ollama on :11434). A local OpenAI-compatible stub stood in for the model call only; discovery, delivery and attribution are the product's own code and are what the scenario observes.

## The split, and why the baseline moved

`file-size` refused the first push: the change put six files past frozen baselines they were each sitting exactly at. The rule's answer is "split instead of extending", and taking it seriously produced five modules that should have existed anyway — see the second commit. `packages/agent-core/src/interfaces/agent.ts` went 308 → 242 and left the baseline list entirely.

The baseline is regenerated in the same commit because three files ended **below** their frozen counts and the scan asks for the gain to be locked. Every row in that diff moves one direction; nothing was adopted upward. That is the check separating "locking a refactor" from "quietly adopting debt", and the scan's own source warns the two are otherwise indistinguishable.

## Red-proof

| reverted | what went red |
| --- | --- |
| the store seam in `execution-service.ts` | 2 of 3 core attribution tests |
| the label in `RoleLabel` | exactly the peer case; both control cases stayed green |

## Also in here

- An orphan export (`driverLabel`) found by reading the diff at the review gate, not by a scan.
- Two `progress-report-quantification` findings of mine recorded — a bare `3/4` twice, the second inside the message acknowledging the first. Transcripts are append-only; the scan's contract is to record, not to edit history.

## Verification

129 of 129 scans pass. 3,302 tests pass across agent-core, agent-session, agent-framework and agent-transport-tui. Workspace typecheck clean.

Closes #1915


woojubb:
## Review

Read the full diff against the base below, not the description of it.

**What I checked and what I found**

- **The attribution path, end to end.** `IRunOptions.driverId` → `buildRunContext` → `IExecutionContext` → `addUserMessage` metadata → `RoleLabel`. Each hop is a pass-through with no branch that could drop it silently. The `run-options-audit` test refuses a field with no registered consumer seam, and it did refuse this one until the seam was named — that gate fired as designed.
- **That the id cannot come from the peer.** The value reaching the label is the framework's derived `peer:<session-id>`, computed from the rendezvous entry. Nothing on this path reads `IPeerOrigin.driverId`. Issue #1809's constraint holds.
- **That it is not an authorization input.** No branch anywhere consults `driverId` to decide what a run may do. It reaches metadata and a `<Text>` node.
- **The owner-elision choice.** Verified the reason rather than the claim: stamping `owner` on every turn changed the `run()` call shape for the dominant path, and eleven `agent-framework` tests asserted the old shape. Eliding restores it and loses nothing, because an unattributed message already reads as the operator's on every surface.
- **The guard on the TUI echo.** Confirmed by test that an unguarded lookup loses the whole message when the session cannot answer — `display-contract` D1 went red on exactly that before the guard was added. Attribution is an addition to the message, not a precondition for it.
- **`driverLabel` scope.** Found it exported with no caller outside its own module and made it private. That is the one finding I raised against my own diff, and it is fixed in `700f9a22c`.
- **The baseline regeneration.** Every row of the `file-size-baseline.json` diff moves one direction — `305→304`, `323→322`, `616→613`, and `packages/agent-core/src/interfaces/agent.ts` dropped out at 242. No entry was adopted upward. That is the check that separates locking a refactor from quietly adopting debt, and the scan's own source says the two are otherwise indistinguishable.

**Evidence beyond the tests.** Two real CLI sessions were run and the transcripts compared. The peer-driven one contains `peer:<id>:` and zero occurrences of `You:`; the operator-typed control contains `You:` and zero occurrences of `peer:`. The control is what makes the first mean anything — without it the run shows only that a label appears, not that the two are told apart.

**Red-proof.** Reverting the store seam fails 2 of 3 core attribution tests; reverting the label fails exactly the peer case and leaves both control cases green.

CI: all checks pass, none pending, none failing. 129 of 129 scans; 3,302 tests across the four touched packages; workspace typecheck clean.

REVIEWED BASE: de8706575bb34fc5cbf4a3fd67d587f9ad9229f0
REVIEWED HEAD: 700f9a22c118fefd131e69cf4bc337000750d635
ACTIONABLE FINDINGS: 0

PR #1936, opened 2026-08-20
feat(packages): typecheck the examples directory CI already executes (INFRA-124)

### LINKED ISSUE #1902: packages/*/examples is typechecked by nobody, so a breaking signature change lands green (follow-up to PR #1901)
Found while diagnosing INFRA-108 (PR #1901), where a breaking signature change landed on develop and turned `quality` red on every branch rebased onto it.

## The gap

`packages/*/examples/` is typechecked by **nobody**.

| surface | what it actually covers |
| --- | --- |
| each package's `tsconfig.json` | `include: ["src/**/*"]` — `examples/` is outside it |
| root `examples:typecheck` script | `pnpm --filter "./examples/**" run typecheck` — the **root** workspace only |
| the `examples-typecheck` CI job | runs the script above, so it inherits that scope despite its name |

Nine packages carry such a directory:

`agent-command`, `agent-core`, `agent-executor`, `agent-framework`, `agent-product`, `agent-session`, `agent-transport`, `agent-transport-tui`, `agent-transport-ws`

## Why it matters more than an ordinary untypechecked folder

These are not illustrative snippets. Several are **executable scenario verifiers wired into `scenario:verify`**, which `quality` runs as a required check. So the code is executed by CI but never read by the compiler — the worst of both: it can break the gate, and the gate that would have caught it at compile time does not look at it.

INFRA-108 is the concrete instance. `createSession` went sync → `async` in PR #1885; the caller in `packages/agent-command/examples/semantic-command-role-scenario-helpers.ts` kept reading `.session` off the returned promise. A typechecker reading that file would have refused it on the spot. Instead it landed green and surfaced later as a runtime `TypeError` pointing at an unrelated cleanup line.

## What a fix has to establish first

Not "add `examples` to every `include`". `rootDir` is `src` and `declaration` is on, so widening `include` changes build output. The likely shape is a separate `tsconfig.examples.json` per package (or one at the root spanning `packages/*/examples`) plus a script the existing `examples-typecheck` job also runs.

**The blast radius has to be measured before the shape is chosen.** My own probe over `packages/*/examples/**/*.ts` was dominated by `TS7016` (missing declaration files) and is not trustworthy — the workspace resolves sibling packages through `exports`/`customConditions` that an ad-hoc config does not reproduce. A real measurement needs a full declaration build and the same resolution settings the packages use.

## Related, deliberately not fixed in PR #1901

The cleanup loop in `verify-semantic-command-roles.ts` does `session.shutdown()` over an array the `try` block may have pushed an `undefined` into. Its TypeError **replaces** the original error, so a genuine failure inside the scenario is reported as a cleanup crash at the wrong line. That masking is why INFRA-108 was first attributed to the wrong branch. It belongs with this issue rather than with the one-line await fix.

## Definition of done

- Every `packages/*/examples/` file is read by a typechecker that a required CI check runs.
- The measured error count at adoption time is recorded, and whatever is not fixed is frozen with a stated reason rather than left silent.
- Reintroducing the INFRA-108 defect (dropping the `await`) is caught at typecheck time, not at scenario runtime — verified by actually reintroducing it.

woojubb:
## The defect

`packages/*/examples/` was read by no typechecker. Nine packages carry such a directory, and several hold executable scenario verifiers wired into `scenario:verify`, which `quality` runs as a **required** check.

So CI executed that code while the compiler never looked at it — the worst of both. It could break the gate, and the gate that would have caught it at compile time did not read it.

## The measurement that decided the shape

Issue #1902 said the blast radius had to be measured against a full declaration build before the shape could be chosen, and recorded that my first probe was untrustworthy. Taken again after a real build, extending each package's own config rather than replacing it: **thirteen errors, not the hundreds the noisy probe implied.**

Wrong by orders of magnitude, and wrong in the direction that would have justified deferring this as a nine-package migration. That is why the issue asked for the second measurement.

Two of the thirteen were configuration rather than code: `agent-core` genuinely imports `../src/*.ts` by extension because tsx runs its source directly, so that package enables `allowImportingTsExtensions` rather than its examples being rewritten to lie about how they run.

## Red-proof

Injecting a type error into `packages/agent-session/examples/verify-offline.ts`:

```
examples/verify-offline.ts(79,7): error TS2322: Type 'string' is not assignable to type 'number'.
```

Before this change that file compiled against nothing. The probe was reverted and the revert verified by re-reading the file and re-running the typecheck, not assumed — a revert in this repo has silently failed before.

## Verification

Workspace typecheck clean across all packages, on a tree rebased onto current `develop`.

Closes #1902

### DIFF
--- a/.agents/tasks/INFRA-124-examples-are-typechecked-by-nobody.md
+++ b/.agents/tasks/INFRA-124-examples-are-typechecked-by-nobody.md
@@ -1,0 +1,91 @@
+---
+title: 'INFRA-124: packages/*/examples was read by no typechecker, so a breaking signature change landed green'
+status: in-progress
+created: 2026-08-20
+priority: high
+urgency: now
+area: packages
+depends_on: []
+---
+
+# INFRA-124: the compiler reads the directory CI executes
+
+## Objective
+
+Issue #1902. `packages/*/examples/` was typechecked by nobody. Nine packages carry such a directory,
+and several hold executable scenario verifiers wired into `scenario:verify`, which `quality` runs as a
+required check. So CI EXECUTED that code while the compiler never read it — the worst of both: it can
+break the gate, and the gate that would have caught it at compile time did not look.
+
+## The measurement the issue asked for, taken properly
+
+Issue #1902 said the blast radius had to be measured against a full declaration build before the shape
+could be chosen, and recorded that my first probe was untrustworthy — it was dominated by `TS7016`
+because an ad-hoc config did not reproduce the workspace's `exports`/type resolution.
+
+Taken again after `pnpm build`, with each package's own `tsconfig.json` extended rather than replaced:
+
+| package                                                                                                  | errors |
+| -------------------------------------------------------------------------------------------------------- | ------ |
+| agent-command                                                                                            | 2      |
+| agent-core                                                                                               | 3      |
+| agent-session                                                                                            | 8      |
+| agent-executor, agent-framework, agent-product, agent-transport, agent-transport-tui, agent-transport-ws | 0 each |
+
+**Thirteen, not hundreds.** The earlier estimate was wrong by orders of magnitude, and wrong in the
+direction that would have justified deferring this.
+
+Two of the thirteen were configuration rather than code: `agent-core` genuinely imports `../src/*.ts`
+by extension because `tsx` runs its source directly, so that package's config enables
+`allowImportingTsExtensions` rather than the examples being rewritten to lie about how they run.
+
+## What the other eleven were
+
+Every one a real defect a reader would have wanted to know about:
+
+- **`TDirectSession` read `.session` off a `Promise`.** ARCH-035 made `createSession` async; INFRA-119
+  awaited the CALL and left this alias one line over, resolving to `never` in silence. The same
+  defect, in the same file, surviving the fix for it — because nothing typechecked the directory.
+- **A command descriptor omitted `userInvocable`**, which `ICapabilityDescriptor` requires.
+- **Five message literals omitted `id` and `state`**, which `IBaseMessage` requires.
+- **Three terminal stubs omitted `writeError`, `prompt` and `select`**, which `ITerminalOutput`
+  requires. The `select` signature was read from the contract rather than guessed — my first stub had
+  a plausible generic shape the interface does not use.
+
+## Approach
+
+Each package gains a `tsconfig.examples.json` extending its own config, and its `typecheck` script
+runs it. Wiring into the EXISTING gate rather than adding a new one: `quality` already runs
+`typecheck` per affected package, so no workflow changes and no second thing to remember.
+
+`rootDir` is the reason a separate config is needed at all — the package config sets it to `src` with
+`declaration: true`, and widening `include` to `examples` conflicts with both.
+
+## Plan
+
+- [x] TC-01: measured all nine packages against a real declaration build — 13 errors, not the
+      untrustworthy earlier figure.
+- [x] TC-02: every error fixed, or its configuration cause named (`allowImportingTsExtensions`).
+- [x] TC-03: all nine packages typecheck their examples at zero.
+- [x] TC-04: `scenario:verify` still passes — the edits are type-level, and the behaviour is unchanged.
+- [x] TC-05: reintroducing the INFRA-119 defect is caught at TYPECHECK time, not scenario runtime —
+      the issue's definition of done, verified by actually reintroducing it.
+- [ ] TC-06: `pnpm harness:pre-push` green.
+
+## Test Plan
+
+The compiler is the test. What needs proving is that it now READS this directory, which a passing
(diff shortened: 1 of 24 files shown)


woojubb:
## Review

Read the full diff against the base below.

**What I checked**

- **That the directory really was unread.** `packages/*/examples/` appears in no `tsconfig.json` `include` on `develop`, while `scenario:verify` — a required check under `quality` — executes several of those files. Confirmed both halves rather than taking the issue's word.
- **That the new configs extend rather than replace.** Each `tsconfig.examples.json` extends the package's own `tsconfig.json` and adds `examples/**/*` alongside `src/**/*`, with `noEmit` and declarations off. A replacing config would have typechecked the examples under different settings than the code they import, which is the failure mode that makes this kind of change worthless.
- **The `agent-core` exception.** It enables `allowImportingTsExtensions` because its examples genuinely import `../src/*.ts` by extension — tsx runs the source directly. The alternative was rewriting those imports to lie about how they run. The exception is narrow, in one package, and stated.
- **The thirteen errors.** Each fix is in the example or its config, not in `src/`. No production type was loosened to make an example compile, which is the tempting wrong move here.
- **The scripts.** Each package's `typecheck` runs both configs, so the examples are checked wherever the existing gate already ran.

**Red-proof.** Injecting a type error into `packages/agent-session/examples/verify-offline.ts` produces `examples/verify-offline.ts(79,7): error TS2322: Type 'string' is not assignable to type 'number'.` Before this change that file compiled against nothing. The probe was reverted and the revert verified by re-reading the file and re-running the typecheck — a revert in this repo has silently failed before, so it is checked rather than assumed.

**On the measurement.** The issue recorded that my first blast-radius probe was untrustworthy and asked for it to be retaken against a full declaration build. Retaken: thirteen, not the hundreds the noisy probe implied — wrong by orders of magnitude, and wrong in the direction that would have justified deferring this as a nine-package migration. Worth stating because the second measurement is what made the change small.

CI: 24 checks, all pass, none pending, none failing. No inline review findings. Workspace typecheck clean on a tree rebased onto current `develop`.

REVIEWED BASE: de8706575bb34fc5cbf4a3fd67d587f9ad9229f0
REVIEWED HEAD: dd7b4ff74cee4501233ebc8cb397a08b3aafde3e
ACTIONABLE FINDINGS: 0

woojubb:
## Review — re-issued after the rebase onto current develop

The earlier verdict on this PR quoted `de8706575` / `dd7b4ff74`. Both moved when PR #1935 merged and this branch was rebased, so that verdict no longer describes what is being merged. This one does.

**What the rebase changed, verified rather than assumed.** The diff shrank from 683 lines to 550. Every one of the 133 lines is a deletion, all in `scripts/harness/progress-report-acknowledgments.json`: the two entries it carried reached `develop` through PR #1935 while this branch waited, so the rebase took develop's copy. **Nothing was added.** The new diff is a strict subset of the reviewed one, which is why the findings below still hold.

**What I checked in the substance**

- **That the directory really was unread.** `packages/*/examples/` appears in no `tsconfig.json` `include` on `develop`, while `scenario:verify` — a required check under `quality` — executes several of those files. Both halves confirmed, not taken from the issue.
- **That the configs extend rather than replace.** Each `tsconfig.examples.json` extends its package's own config and adds `examples/**/*` alongside `src/**/*`, `noEmit` on, declarations off. A replacing config would have checked the examples under different settings than the code they import, which is what makes this kind of change worthless.
- **The `agent-core` exception.** It enables `allowImportingTsExtensions` because its examples genuinely import `../src/*.ts` by extension — tsx runs the source directly. The alternative was rewriting those imports to lie about how they run. One package, stated.
- **The thirteen errors.** Every fix is in an example or its config. No production type was loosened to make an example compile, which is the tempting wrong move.

**Red-proof.** Injecting a type error into `packages/agent-session/examples/verify-offline.ts` yields `examples/verify-offline.ts(79,7): error TS2322: Type 'string' is not assignable to type 'number'.` Before this change that file compiled against nothing. The probe was reverted and the revert verified by re-reading the file and re-running the typecheck — a revert in this repo has silently failed before.

**On the measurement.** The issue recorded that my first blast-radius probe was untrustworthy and asked for it to be retaken against a full declaration build. Retaken: thirteen, not hundreds — wrong by orders of magnitude, and wrong in the direction that would have justified deferring this as a nine-package migration. The second measurement is what made the change small.

**Disclosure.** The rebase push used the documented inline `PRE_PUSH_ALLOW_UNREVIEWED=1`. The hook blocks any push to a PR reporting zero findings, because it cannot tell a rebase from new work. What it could not reach, I checked by hand: the diff comparison above. No new work was pushed.

CI: all checks pass, none pending, none failing. `mergeStateStatus: CLEAN`.

REVIEWED BASE: 45b54cc425c8ed7ee97c6da00d7a601bc971f78f
REVIEWED HEAD: aa7ac0eadf882eed7347c5e10908c6f352062512
ACTIONABLE FINDINGS: 0

Issue #1937, opened 2026-08-20
--task-file, --json-schema and --append-system-prompt are silently dropped in TUI and serve mode


woojubb:
Split out of #1820 (ARCH-040 Group D), which names it as "related and worth fixing with them".

## What

`buildAppendSystemPrompt(cwd, args)` composes the CLI-sourced system-prompt addition from three flags — `--append-system-prompt`, `--task-file` and `--json-schema`. It has **exactly one caller**: `packages/agent-cli/src/modes/print-mode.ts`.

So in interactive TUI mode and in serve mode, all three flags are parsed, validated, and then dropped. `packages/agent-cli/src/utils/cli-args.ts` advertises `--task-file` with no indication that it only works in print mode.

## Why it is filed rather than fixed inside #1820

ARCH-040 Group D has to decide the merge order between a preset's `appendSystemPrompt` and the CLI-sourced text. That decision is not expressible while the CLI-sourced text reaches one surface of three: "appends after the CLI text" has no meaning in the two surfaces where the CLI text never arrives.

Fixing the ordering first would encode a rule against a base that is missing two thirds of the time — so this is the prerequisite, not a companion.

## Fix direction

Compose the CLI-sourced text ONCE, where the preset surface projection is built, and carry it on `IPresetSurfaceOptions` so all three shells receive it — the property that projection exists to provide (ARCH-041). Fixing it by calling `buildAppendSystemPrompt` from two more places would be a third copy of the same call, which is the shape ARCH-041 just removed.

## Done when

- the three flags take effect in interactive, serve and print modes;
- a case asserts that, per surface, rather than asserting the helper's return value — the defect is that the helper is not CALLED, so a test of the helper passes while the flags do nothing.

## Context

`.agents/tasks/ARCH-040-preset-projection-decisions.md` carries the decision table. Group D's `systemPrompt` half is independent of this and proceeds; the `appendSystemPrompt` half waits on it.

woojubb:
Fixed in PR #1944, merged to `develop` as `7bb54097c`. Verified on the remote.

`buildAppendSystemPrompt` had exactly one caller, so the three flags were parsed, validated and then dropped in interactive and serve mode. Composed once now, inside `buildPresetSurfaceOptions` — the one shape all three shells already read. Calling the helper from two more places would have been the third copy of the same call, the shape ARCH-041 removed from that file.

## The naming, which the fix direction did not anticipate

[...]

Issue #1934, opened 2026-08-20
A preset's tool lists cannot be re-applied to a live session: permission patterns are fixed at construction


woojubb:
Split out of #1820 (ARCH-040 Group C). The combine RULE is decided; what is missing is a capability.

## What #1820 asked for, and what is actually blocking it

A preset carries `allowedTools` and `deniedTools`. Neither reaches a session today. The owner decided how they should combine on 2026-08-20:

- an **allowlist REPLACES** — it is a statement of the complete permitted set, so a later, more specific layer stating one supersedes the earlier answer rather than intersecting with it;
- a **denylist UNIONS** — a denial is not weakened by a later layer that forgot to repeat it.

That rule's home is `resolvePreset`'s `mergeDefined`, because precedence is the resolver's job: a union applied by one of three shells is a rule the other two disagree with.

## The blocker

`packages/agent-framework/src/assembly/create-session.ts` turns both lists into permission PATTERNS (`Name(*)`) and merges them into the enforcer's `allow`/`deny` sets **at construction**. `PermissionEnforcer` exposes only an additive `sessionAllowedTools` set — the "always allow" prompt path. **There is no way to REPLACE the configured rules on a running session.**

So the live `/preset` path cannot re-apply a preset's tool lists, while the startup path could.

## Why the startup half was written and then reverted

Wiring startup alone **creates** the divergence `scan-preset-projection` exists to measure: starting with a preset would apply its tool lists and switching to the same preset mid-session would silently not. The implementation was written, measured against that scan, and reverted. Half-applied would have been worse than absent, and the exemption entries in `.agents/harness.config.json` now carry this measurement instead of the original "needs a decision", which is no longer what is true.

## What this item has to decide, beyond the plumbing

A live permission-rule replacement has consequences a construction-time merge does not:

- what happens to a tool call already in flight when its rule changes;
- whether a denial can be added to a session that has already allowed that tool — and whether an earlier "always allow" answer survives it;
- whether the session's own accumulated `sessionAllowedTools` outrank a newly applied preset denial, or the reverse.

Those are user-visible and are the reason this is a capability rather than a wiring task.

## Done when

`allowedTools` and `deniedTools` are declared on BOTH projection surfaces, `presetProjection.pendingProjection` loses both entries, and the combine rule is asserted in both directions — an allowlist replacing and a denylist unioning — because a rule about combination can fail either way.

## Context

[...]

PR #1938, opened 2026-08-20
feat(preset): ARCH-040 groups D and F — seed the system prompt, apply the language

woojubb:
Two more groups of #1820. With these, six of the ten fields are resolved and the remaining four are split into their own items with the blocking cause named.

## Group F — `language`

Declared on the preset contract with no seam under any name. Owner decision: a PROMPT instruction, not a provider parameter — and the framework already composes a response-language section, so this wires what exists.

`language` joins `persona` and `selfVerification` as a value the rebuild closure retains and a live preset switch can override.

**Two things were built and then removed**, both second routes to one value:

- an `ICreateSessionOptions.language` key — a preset's language already ARRIVES at `config.language`, so this was a second path to the same place. `doc-examples` found it as a duplicate identifier before the reasoning did.
- a third copy of the live-prompt applier — `applyPersona` and `applySelfVerification` were the same five lines twice. They share one private `rebuildLivePrompt` now, which is also what made room under the size ratchet: the file is the same length with two more capabilities.

## Group D — `systemPrompt` SEEDS, it does not replace

The session already offers a REPLACE seam. Pointing a preset at it would silently drop the AGENTS.md, project-notes, skill and capability sections — context the person choosing a preset did not ask to lose. So it is a priority-4 section, above persona (5) and therefore above AGENTS.md (10).

The decisive case is the pair: *"the text appears"* is true of a replacement too. Only *"the text appears AND the surrounding sections survive"* tells them apart, so both are asserted — plus the ORDER against persona, by position rather than by the priority number.

**The two session options are one letter apart.** `systemPrompt` replaces, `presetSystemPrompt` seeds. A bare `...presetSurface` lands a preset's text on the replacing one and drops those sections invisibly. `toSessionOptions` renames it in ONE place, so the three shells cannot disagree about which option a preset means.

## A split the anti-monolith floor forced, and was right to

`interactive-session-options.ts` held THREE interfaces restating one option list — the public standard arm, the injected arm, and `IInitOptions` — so every preset field had to be added three times, and the floor refused the fourth addition. `IInitOptions` moves to its own module: what a CALLER may pass stays, what the async init hands onward leaves.

`option-reachability` **fail-closed on the move** rather than passing, which is exactly what that entry exists for.

## What is split out, and why it is not a deferral

- **#1934** — the tool lists. The combine rule is decided (allow replaces, deny unions); the live half needs a capability that does not exist, since the lists become permission PATTERNS at construction and the enforcer exposes only an additive set. Wiring startup alone CREATES the divergence the scan measures, so it was written, measured, and reverted.
- **#1937** — `appendSystemPrompt`. Its merge order is not expressible while `buildAppendSystemPrompt` has one caller and the CLI-sourced text reaches one surface of three. That is the prerequisite, not a companion.
- **Group B (`agentName`)** — parked awaiting an owner re-decision, with the measured cost recorded: its only reader is a `readonly` identity label no surface displays.

`presetProjection.pendingProjection`: 6 exemptions → 4.

## Verification

- `pnpm typecheck` — 0 errors
- full suite — exits 0
- red-proofed: dropping the seed section fails the seed cases, lowering its priority below persona fails exactly the ordering case, and dropping either live application fails exactly its live case

### DIFF
--- a/.agents/harness.config.json
+++ b/.agents/harness.config.json
@@ -372,9 +372,9 @@
     },
     {
       "name": "IInitOptions",
-      "file": "packages/agent-framework/src/interactive/interactive-session-options.ts",
+      "file": "packages/agent-framework/src/interactive/interactive-session-init-options.ts",
       "constructors": ["createInteractiveSession"],
-      "reason": "The INIT option shape, watched because ARCH-013 stage 3 dropped two fields in the ~40-field hand-map that builds it and stage 1's ICreateSessionOptions-only ratchet could not see the drop. Named IInitOptions rather than the published IInteractiveSessionStandardOptions because that is literally what it measures: the single production literal it reads is the createInteractiveSession(...) call, which takes IInitOptions. An earlier revision named the published type, and review measured the difference — 8 of the 10 keys it then reported ARE set by production code through buildRuntimeSession and the surface option builders, so the entry asserted unreachability it had not established, and burning it down would have pushed class-consumed options into the very hand-map this stage fixed."
+      "reason": "The INIT option shape, watched because ARCH-013 stage 3 dropped two fields in the ~40-field hand-map that builds it and stage 1's ICreateSessionOptions-only ratchet could not see the drop. Named IInitOptions rather than the published IInteractiveSessionStandardOptions because that is literally what it measures: the single production literal it reads is the createInteractiveSession(...) call, which takes IInitOptions. An earlier revision named the published type, and review measured the difference — 8 of the 10 keys it then reported ARE set by production code through buildRuntimeSession and the surface option builders, so the entry asserted unreachability it had not established, and burning it down would have pushed class-consumed options into the very hand-map this stage fixed. ARCH-040 split it into its own module: the original file held three interfaces restating one option list, so every preset field had to be added three times and the anti-monolith floor refused the fourth. This scan fail-closed on the move rather than passing, which is what the entry is for."
     }
   ],
   "productIdentity": {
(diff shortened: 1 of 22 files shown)


woojubb:
## Pre-merge review

**base:** `18dbba14a2e4c47dd55aea0b5d677c79ef5af6a4`
**head:** `6a2053910e8c8b61d23441f6cc9adc2a02a1b7c2`

Reviewed the whole diff at this head. CI green, no review threads opened.

### What I checked hardest

Both groups add a field to a projection that reaches three shells, so the risk is a value that stops at a boundary while the type says otherwise:

- **The seed-vs-replace distinction.** `systemPrompt` and `presetSystemPrompt` are one letter apart and mean opposite things. Verified that `toSessionOptions` is the only place the rename happens and that `cli.ts` no longer spreads the surface raw — a bare spread would land a preset's text on the replacing option and drop the AGENTS.md and capability sections with nothing reporting it. The test asserts the absence of `systemPrompt` from the mapped object, not just the presence of `presetSystemPrompt`, because only the pair rules out the leak.
- **The seed's placement** is asserted by ORDER against persona rather than by the priority number, so it survives a renumbering that preserves behaviour and fails one that does not.
- **Two second routes were built and removed**, not merely avoided: an `ICreateSessionOptions.language` key and a third copy of the live-prompt applier. Both are recorded in the commit message, because "we considered it" and "we built it and measured why not" are different claims.

### Three gates caught things here, and each changed the design rather than being worked around

- `doc-examples` found the duplicate `language` identifier before the reasoning did.
- The anti-monolith floor refused the fourth addition to a file holding three interfaces restating one option list — which produced the `IInitOptions` split, and the split is the right shape independent of the size number.
- `option-reachability` **fail-closed** on that move rather than passing, which is exactly what its entry exists for.

### Scope discipline

`appendSystemPrompt` and the tool lists are NOT in here. Each was traced to a prerequisite that does not exist yet — a one-caller helper (#1937) and a missing live-permission capability (#1934) — and wiring either half alone would have created the divergence the projection scan measures. The tool half was written, measured against that scan, and reverted for that reason.

### Verification at this head

- `pnpm typecheck` — 0 errors
- full suite — exits 0
- red-proofed: dropping the seed section fails the seed cases, lowering its priority below persona fails exactly the ordering case, and dropping either live application fails exactly its live case

**ACTIONABLE FINDINGS: 0**

PR #1939, opened 2026-08-20
docs(agent-cli): document /peers, the last thing issue #1863 was waiting on

### LINKED ISSUE #1863: agent-cli: local peer discovery and the command that addresses another live session (follow-up to #1809)

woojubb:
## What was missing

Issue #1863 lists four items. Three shipped — local discovery (PEER-004), the command (PEER-006), and the TUI attribution (PEER-007, issue #1915). The fourth was **"CLI documentation of the message flow and its concurrency behaviour"**, and `/peers` appeared in no README, no SPEC, and no command table. Every other slash command is listed; this one was reachable only by typing `/` and reading the popup.

## What this adds

- A `Sessions on this host` row in the slash-command table.
- A `Talking to another session` section: seeing who is there, sending, what the receiving operator sees, and what happens when that session is busy.

## The concurrency half is the part worth writing down

Because the interesting states are the ones that are **not** "delivered":

| state | what the sender is told |
| --- | --- |
| queued behind a running turn | reported as queued, deliberately **not** as delivered |
| displaced by a later message from the same sender | refused, reason `coalesced` |
| queue at capacity (32) | refused, reason `dropped` |
| queue cleared — abort, cancel, shutdown | refused, reason `cancelled` |

The first row is the one an operator needs: reporting a queued message as delivered would hide a wait they can otherwise see and act on. The rest matter because a submission that never became a turn says so, rather than leaving the sender to assume it ran.

## Read out of the code, not out of the issue

That distinction earned its keep. Issue #1863 says the sender learns "via a `coalesced` ack" — but `TPeerDeliveryState` has no `coalesced` member. The actual mechanism is `peer-message-ingress.ts` settling `refused` with reason `coalesced`. Documenting the issue's phrasing would have described a field that does not exist.

Sources for each claim:

- `describeSend` in `packages/agent-command/src/peers/peers-command.ts` — the four send outcomes and their exact wording;
- the enqueue path in `packages/agent-framework/src/interactive/interactive-session-pending-queue.ts` — the tail-coalesce rule and the depth of 32;
- the header of `packages/agent-framework/src/interactive/peer-message-ingress.ts` — how each queue outcome becomes a delivery state the sender sees.

Every string quoted in the new section was checked against the source it came from.

## Also: the user-execution evidence for PEER-004 and PEER-006

Recorded on both task records, including a **second run for both directions**. Issue #1863's definition of done says *both*, and one direction only shows that a label appears — not that it names the right peer:

| direction | label the receiver rendered | sender's id from `/peers` |
| --- | --- | --- |
| B → A | `peer:620c8924-9ee6-4058-9e65-2d0204b2a103` | B = `620c8924-…` |
| A → B | `peer:dfad4f56-7bc3-4eb4-b1e2-e86f10960333` | A = `dfad4f56-…` |

Each session labelled the other one. A single direction would have passed even with the label wired to the wrong end.

Both runs were real `pnpm cli:dev` sessions driven through PTYs on one host as one user, with a local OpenAI-compatible stub standing in for the model call only — no credential was present, and discovery, delivery and attribution are the product's own code.

## Verification

129 of 129 scans pass. Documentation and task records only; no source change.

Closes #1863

### DIFF
--- a/.agents/tasks/PEER-004-local-discovery-reaches-the-operator.md
+++ b/.agents/tasks/PEER-004-local-discovery-reaches-the-operator.md
@@ -68,21 +68,33 @@ session is, which is the question the registry keys its entries on.
       throwing or taking the other adapter down with it.
 - [x] TC-07: `pnpm harness:scan` green and the three touched packages' suites green.
 - [x] TC-08 (user-execution): two `agent-cli` sessions on one host, `/peers` in each showing the
-      other. **EXECUTED 2026-08-20** — two real `pnpm cli:dev` sessions driven through PTYs, same host,
-      same user. The first, alone, reported `No other live session is announced. Start a second session
-  on this host, as this user, and it appears here.` Once the second was up, its `/peers` listed both
-      with itself marked:
+      other. Executed 2026-08-20 — evidence below.
 
-      ```
-              Live sessions:
-                b8319b98-bb7b-486c-a499-cf1585b39e61  (this session)
-                97ffafe3-f770-4877-90a4-bd86df6be010
-              ```
+## User Execution Evidence
 
-              TC-05's distinction is what the first output shows: "no other session" reads as its own sentence
-              with a next step, not as an empty list a reader would have to interpret. Full transcripts and the
-              `/peers send` half are recorded in
-              [PEER-006](PEER-006-peers-send-reaches-the-other-session.md).
+**TC-08, executed 2026-08-20.** Two real `pnpm cli:dev` sessions driven through PTYs, same host, same
+user.
+
+The first, alone:
+
+```
+No other live session is announced. Start a second session on this host, as this user, and it
+appears here.
+```
+
+Once the second was up, its `/peers`:
+
+```
+Live sessions:
+  b8319b98-bb7b-486c-a499-cf1585b39e61  (this session)
+  97ffafe3-f770-4877-90a4-bd86df6be010
+```
+
+TC-05's distinction is what the first output shows: "no other session" reads as its own sentence with
+a next step, not as an empty list a reader would have to interpret.
+
+Full transcripts and the `/peers send` half, including both directions, are recorded in
+[PEER-006](PEER-006-peers-send-reaches-the-other-session.md).
 
 ## Test Plan
 
--- a/.agents/tasks/PEER-004-local-discovery-reaches-the-operator.md
+++ b/.agents/tasks/PEER-004-local-discovery-reaches-the-operator.md
@@ -67,8 +67,22 @@ session is, which is the question the registry keys its entries on.
 - [x] TC-06: the assembly wires the adapter, generates the id, and reports a refusal without
       throwing or taking the other adapter down with it.
 - [x] TC-07: `pnpm harness:scan` green and the three touched packages' suites green.
-- [ ] TC-08 (user-execution): two `agent-cli` sessions on one host, `/peers` in each showing the
-      other. Runnable once this ships.
+- [x] TC-08 (user-execution): two `agent-cli` sessions on one host, `/peers` in each showing the
+      other. **EXECUTED 2026-08-20** — two real `pnpm cli:dev` sessions driven through PTYs, same host,
+      same user. The first, alone, reported `No other live session is announced. Start a second session
+  on this host, as this user, and it appears here.` Once the second was up, its `/peers` listed both
+      with itself marked:
+
+      ```
+              Live sessions:
+                b8319b98-bb7b-486c-a499-cf1585b39e61  (this session)
+                97ffafe3-f770-4877-90a4-bd86df6be010
+              ```
+
+              TC-05's distinction is what the first output shows: "no other session" reads as its own sentence
+              with a next step, not as an empty list a reader would have to interpret. Full transcripts and the
+              `/peers send` half are recorded in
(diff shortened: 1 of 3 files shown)


woojubb:
## Review

Documentation and task records only; no source change. Read the full diff against the base below.

**What I checked, and it is the only thing that matters for a doc change: is every claim true of the code?**

Each string quoted in the new section was matched against the source it came from:

| claim | source |
| --- | --- |
| `No other live session is announced. …` | `packages/agent-command/src/peers/peers-command.ts` |
| `Live sessions:` / `Send to one: …` | same |
| `<id> has the message; it is waiting behind work already running there.` | `describeSend`, same file |
| `<id> had already seen that message.` | same |
| queue depth 32 | `MAX_PENDING_QUEUE_DEPTH` in `packages/agent-framework/src/interactive/interactive-session-pending-queue.ts` |
| `coalesced` / `dropped` / `cancelled` | `TTurnNotRunReason` in `packages/agent-interface-transport/src/turn-contracts.ts` |

**One place where the issue and the code disagree, and the code won.** Issue #1863 says the sender learns "via a `coalesced` ack". There is no `coalesced` member in `TPeerDeliveryState` — the actual mechanism is `peer-message-ingress.ts` settling `refused` with reason `coalesced`. Documenting the issue's phrasing would have described a field that does not exist. This is why the section was written from the source rather than from the issue text.

**Evidence recorded on the task records.** PEER-004's TC-08 and PEER-006's two scenarios now carry executed evidence, including a second run covering **both directions** — issue #1863's definition of done says *both*, and one direction only shows that a label appears, not that it names the right peer. B's message rendered in A as B's session id; A's rendered in B as A's. A single direction would have passed with the label wired to the wrong end.

**One fix after the first push.** `format-check` went red. Prettier and I disagreed about indenting a fenced block nested inside a `- [x]` item, and each pass indented it further — the third run had the block at twenty-two columns and a sentence escaped to column zero. Fixed by moving the evidence out of the checklist item into its own section, where it needs no nesting. Prettier now reports the file **unchanged**, which is the property that was missing: formatted once is not the same as stable on a second pass. Verified with the entry point CI actually uses (`node scripts/harness/verify-like-ci.mjs --only format-check`), not `pnpm format:check` — which passed locally while CI failed, and is why that mirror exists.

CI: all checks pass, none pending, none failing. `mergeStateStatus: CLEAN`. No inline review findings. 129 of 129 scans.

REVIEWED BASE: c2bf637865e7f280af1d88eac78bcd81278f28cc
REVIEWED HEAD: 917b0b437b4f7ce055b1e46faeb75d42dd2c4556
ACTIONABLE FINDINGS: 0

PR #1940, opened 2026-08-20
feat(preset)!: rename the agent when a preset switch says so (ARCH-040 Group B)

### LINKED ISSUE #1820: ARCH-013 stage 2 remainder: ten resolved-preset fields need a projection decision, not just a seam

woojubb:
Closes #1820.

The last of ARCH-040's ten fields that this repository can decide. The two that remain are split into their own items with the blocking cause named — see the end.

## The divergence

Startup applied the preset's `agentName`; the live `/preset` path did not. So starting with a preset set the name, and switching to the SAME preset mid-session left the old one — one preset with two answers, decided by WHEN it was chosen. Owner decision: `/preset` renames.

## This group was parked for a day on an analysis that was too narrow

The parked note offered two options — a mutable field on a core class, or splitting a 411-line frozen file — and called them the only two. **There was a third, and it is better than either.**

`Robota` declares `protected override config`, which means the BASE already declares `config`. So `name` becomes a getter on `RobotaBase` reading `this.config.name`, and both the field declaration and its constructor assignment leave the subclass. `robota.ts` is **three lines smaller**, and its ratchet is re-frozen at the gain.

Because `name` reads THROUGH `config`, `updateConfiguration({ name })` is the whole rename — there is no second copy on the instance to leave stale, which is exactly what the original "mutable field" sketch would have created. The seam stays narrow: a patch that is neither `tools` nor `name` still throws, and a case asserts it.

## One guard was satisfied rather than dodged

`product-identity` refused the single new call: `agent-session`'s agent field was named `robota` — the consumer's product name inside a neutral library — frozen at 30 occurrences with **no exemption mechanism**.

Two ways out: spell the call so it avoids the `.robota` marker, or fix what the marker points at. Re-spelling the thing a guard detects is the failure mode this repository has scans about, so the field is `agent` now and the count fell **30 → 3**. The remainder is the real `~/.robota` session directory, which is a genuine product path.

**BREAKING**: `SessionBase`'s protected `robota` member is `agent`. `getBuiltInAgent`-style external subclasses must rename their property.

## Why #1820 closes here

`presetProjection.pendingProjection`: **10 exemptions → 3**, and each of the three now has an owner:

| field | disposition |
| --- | --- |
| `allowedTools`, `deniedTools` | **#1934** — the combine rule is decided; the live half needs a capability that does not exist (permission patterns are fixed at construction). Written, measured against the projection scan, and **reverted**, because wiring the startup half alone CREATES the divergence that scan exists to measure. |
| `appendSystemPrompt` | **#1937** — its merge order is not expressible while `buildAppendSystemPrompt` has one caller and the CLI text reaches one surface of three. The prerequisite, not a companion. |

Everything ARCH-040 could decide is decided: `defaultTrustLevel` removed, the model group carried to construction, `language` applied as a prompt section, `systemPrompt` seeding rather than replacing, `model` declared, and `agentName` renaming live.

## Verification

- `pnpm harness:scan` — 128 passed, 3 skipped
- `pnpm typecheck` — 0 errors
- full suite — exits 0
- red-proofed: copying the name onto the instance instead of reading through `config` fails two of the four rename cases

### DIFF
--- a/.agents/harness.config.json
+++ b/.agents/harness.config.json
@@ -425,11 +425,6 @@
     "$pendingComment": "The burn-down. Each entry is a field this floor HAS measured as undeclared or one-sidedly declared, whose fix needs a decision this scan cannot make — recorded as a NAMED, EXPIRING exemption rather than an opaque count, because an exemption that is unnamed or non-expiring is indistinguishable from no floor. `declaredOn` records the surfaces the field was measured on when the entry was written; if the live state differs in EITHER direction the entry is reported (`preset-pending-state-changed`), so an exemption expires when the work is DONE and not only when the field is deleted. Review measured the earlier version failing exactly that: fixing a pending field printed green, which made the list a baseline with extra words. This list may shrink and must never grow without a reviewed reason. Tracked as GitHub issue #1820.",
     "pendingProjection": [
       {
-        "field": "agentName",
-        "reason": "The REVERSE divergence: startup declares it (IPresetSurfaceOptions) and the live /preset path does not, so starting with a preset sets the agent name while switching to the SAME preset mid-session leaves the old one. The owner decided on 2026-08-20 that /preset SHOULD rename. Implementation is parked on a measured cost the decision did not have in front of it: the field's only reader is `Robota.name`, a `public readonly` identity label that no surface displays, and making it renameable needs either a mutable core field or a split of a 411-line frozen core file. Re-decide against that cost before wiring — see ARCH-040.",
-        "declaredOn": ["IPresetSurfaceOptions"]
-      },
-      {
         "field": "appendSystemPrompt",
         "reason": "A session seam DOES exist (`ICreateSessionOptions.appendSystemPrompt`, projected by create-session-projection.ts), so the last hop is ready. The first hop is missing, and wiring it requires deciding the merge order against the CLI-sourced append text that `buildAppendSystemPrompt` produces — today that helper is called only from print-mode.ts, so `--task-file`/`--json-schema`/`--append-system-prompt` are themselves dropped in TUI and serve mode.",
         "declaredOn": []
--- a/.agents/tasks/ARCH-040-preset-projection-decisions.md
+++ b/.agents/tasks/ARCH-040-preset-projection-decisions.md
@@ -43,8 +43,8 @@ projection, and where the fallback lives is an implementation choice made in thi
 ## Plan
 
 - [x] Group A — remove `defaultTrustLevel` from `IResolvedPresetOptions` and its validator.
-- [ ] Group B — `agentName` on the live `/preset` path. **PARKED on a measured cost the decision did
-      not have in front of it** (see below); every other group is independent of it.
+- [x] Group B — `agentName` on the live `/preset` path. The parked cost turned out to have a third
+      option the first analysis missed — see below.
 - [ ] Group C — `allowedTools` (replace) and `deniedTools` (compose), resolved together. **BLOCKED on a
       missing capability**, see below. The RULE is decided and its home is identified; only the live
       seam is absent.
@@ -69,7 +69,7 @@ session/assembly seam"_. It is false today for `systemPrompt`, `language`, `temp
 `maxOutputTokens` and `defaultTrustLevel`. Changing the words without changing the fact is the
 defect, not the fix; this item makes it true.
 
-## Group B is parked, and why
+## Group B: parked, then unparked — the first analysis was too narrow
 
 The owner decided `/preset` should rename the agent. Implementing it turned up a cost the decision
 was not made against, so it is recorded here rather than paid silently.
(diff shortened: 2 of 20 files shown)


woojubb:
## Pre-merge review

**base:** `67c1366c4b31b3fbf2bf28d6a795760f6069a647`
**head:** `7a0a7c311fbc213519bafbf1f4097a07a460bbb2`

Reviewed the whole diff at this head. CI green, no review threads opened.

### What I checked hardest

This change makes a previously-immutable identity mutable, so the risk is a rename that moves one reader and not the others — the divergence the item is about, reproduced one layer down.

- `name` reads through `config`, so there is **no second copy** to leave stale. Verified `getConfig()`, `getStats()` and `getName()` all resolve through the same value, and the cases assert `getConfig().name === robota.name` after a rename rather than either alone.
- The seam stays narrow: `updateConfiguration` still throws for a patch that is neither `tools` nor `name`, with a case for it. Accepting `name` must not quietly turn it into a general config setter.
- Red-proofed: copying the name onto the instance instead of reading through `config` fails two of the four cases — which is precisely the design that was rejected.

### The rename in `agent-session` is a burn-down, not collateral

`product-identity` refused the single new call, and it was right: a neutral library's agent field was named after the consumer's product, frozen at 30 with no exemption mechanism. The two ways out were to re-spell the call so it avoids the `.robota` marker, or to fix what the marker points at. The first is the failure this repository has scans about. Count fell **30 → 3**, and the remainder is the genuine `~/.robota` directory.

### A correction worth recording

This group was parked for a day on my analysis that the only options were a mutable core field or splitting a 411-line frozen file. Both were wrong: `Robota` declares `protected override config`, so the base already owns `config`, and putting the getter there made `robota.ts` **smaller**. The parked note is updated in the task record rather than deleted, because the narrow analysis is the more useful thing for the next reader.

### Scope

`allowedTools`/`deniedTools` (#1934) and `appendSystemPrompt` (#1937) are deliberately absent — each traced to a prerequisite that does not exist, and the tool half was written, measured against the projection scan, and reverted because wiring its startup side alone creates the divergence that scan measures.

### Verification at this head

- `pnpm harness:scan` — 128 passed, 3 skipped
- `pnpm typecheck` — 0 errors
- full suite — exits 0
- `presetProjection.pendingProjection`: 10 → 3, each remaining entry owned by a filed issue

**ACTIONABLE FINDINGS: 0**

PR #1941, opened 2026-08-20
fix(agent-command): stop writing the budget through a path someone else can redirect

### LINKED ISSUE #1928: Security: address CodeQL high-severity file-system race in session command
## Observed
The merged promotion PR #1895 had a blocking review-gate finding and CodeQL security alert for a high-severity `js/file-system-race` at `packages/agent-command/src/session/session-command.ts:106`.

The review-gate log reported:
- `js/file-system-race [warning/security:high] packages/agent-command/src/session/session-command.ts:106`
- The PR also reported one high and four medium CodeQL security alerts in changed code.

## Scope
Investigate the path/file-system access around line 106, establish the race condition and a safe fix, then add regression coverage. The finding was observed during post-merge verification of #1895; the merge landed in `main` as `01a245c35740ea20d87bbc1aa4a6bd55200885d5`.

## Why not fixed inline
This was discovered during post-merge verification and is outside the completed issue/task-boundary documentation change; fixing it now would widen the work in flight and requires a dedicated security review.

## Not this
This issue does not represent the already-completed issue-versus-task/backlog boundary work. Depth and remediation disposition should be triaged when the issue is converted to a backlog item.

woojubb:
## Reproducing it found a second defect the report did not name

CodeQL flagged a high-severity `js/file-system-race` at `packages/agent-command/src/session/session-command.ts:106`. Reproducing rather than patching turned up **two** problems, and fixing only the reported one would have left the more useful half of the attack working.

### 1 — the reported race

```ts
function clearBudget(cwd: string): void {
  const file = join(cwd, BUDGET_FILE);
  if (existsSync(file)) {      // check
    writeFileSync(file, '{}'); // …use, line 106
  }
}
```

Swapping the path for a symlink between those two lines makes the write land on the symlink's target. Demonstrated, with a file outside the project overwritten by `{}`.

Removing the check closes the window. And the check was protecting nothing: `readBudget` reads an absent file, an empty document and malformed JSON all as "no budget set", so writing unconditionally says exactly what the check was guarding. The only visible change is that clearing on a project that never had the file now creates one holding `{}` — the same meaning, written down.

### 2 — what removing the check does *not* fix

A symlink planted **before** the call is still followed by a plain write. Re-running the same attack against the check-free version overwrote the target again.

So the race and the redirection are separate defects, and only one was reported. Stopping at the CodeQL finding would have produced a green alert and a live vulnerability.

`O_NOFOLLOW` closes the second: the open fails `ELOOP` when the final path component is a symlink, so the kernel decides at open time rather than a check that can go stale between asking and acting.

| | before | check removed | + `O_NOFOLLOW` |
| --- | --- | --- | --- |
| symlink swapped mid-window | target overwritten | n/a — no window | n/a |
| symlink planted beforehand | target overwritten | **target overwritten** | `ELOOP`, target intact |

## Decisions worth flagging

- **Both writers, one helper.** `writeBudget` had the same exposure. Fixing only the line CodeQL named would have left its sibling open.
- **Windows.** `O_NOFOLLOW` is undefined there, so the write falls back to the plain one rather than claiming a protection the platform cannot give. The test for it **skips** on such a platform rather than passing without testing — a test that passes by not testing is worse than one that says it did not run.
- **A refused write is a result, not a throw.** `success: false` naming the symlink, instead of an `ELOOP` escaping a synchronous command. "Could not write the budget" would send someone to look at permissions; the thing to look at is what the budget file points to.
- **`readBudget`'s own `existsSync` is gone too.** Same check-then-use shape, harmless there because the catch already handled it — which is exactly why it was redundant.

## Red-proof, separately for each half

One fix must not stand in for the other, so each was reverted alone:

| reverted | what went red |
| --- | --- |
| restore `existsSync` in `clearBudget` | exactly the clear-path test |
| drop `O_NOFOLLOW` | exactly the symlink test |

Nineteen of twenty stayed green in each case. The `clear` path had **no test at all** before this.

## Incidentally

`product-identity` for this package fell 8 → 7: the new code takes its directory from `BUDGET_FILE` rather than repeating the literal. Re-frozen in the same change, as that scan asks. Every row of that baseline diff moves down.

## Verification

126 of 129 scans pass (3 skipped). 265 tests pass in `agent-command`. Typecheck clean.

Closes #1928

### DIFF
--- a/packages/agent-command/src/session/__tests__/session-command-module.test.ts
+++ b/packages/agent-command/src/session/__tests__/session-command-module.test.ts
@@ -1,3 +1,15 @@
+import {
+  constants,
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+  symlinkSync,
+  writeFileSync,
+} from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
 import { describe, expect, it, vi } from 'vitest';
 import type { ICommandHostContext, ICommandSessionRuntime } from '@robota-sdk/agent-framework';
 import { InteractiveSession, SystemCommandExecutor } from '@robota-sdk/agent-framework';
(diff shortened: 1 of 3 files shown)


woojubb:
## Review

Security fix; read the full diff against the base below, and re-checked the claims by running the attack rather than by reading the code.

**The two defects are genuinely separate, and I verified that separately.**

| | before | check removed only | + `O_NOFOLLOW` |
| --- | --- | --- | --- |
| symlink swapped inside the check→write window | target overwritten | n/a, no window | n/a |
| symlink planted before the call | target overwritten | **target overwritten** | `ELOOP`, target intact |

The middle column is the finding that matters: fixing what CodeQL reported would have produced a green alert over a live redirection. Stopping there was the tempting move.

**That removing the check is safe** — `readBudget` reads an absent file, an empty document and malformed JSON all as "no budget set", so an unconditional write says exactly what the check was guarding. Checked all three paths, not just the first.

**That both writers are covered.** `writeBudget` had the same exposure and was not named by the alert. Fixing only the flagged line would have left the sibling open, which is the same shape as fixing only the reported half above.

**Windows.** `O_NOFOLLOW` is undefined there; the write falls back to the plain one rather than claiming a protection the platform cannot give, and the test **skips** rather than passing without exercising anything. A test that passes by not testing is the defect this repo files issues about.

**The failure is a result, not a throw.** `executeCostCommand` is synchronous and returns `ICommandResult`; an escaping `ELOOP` would have surfaced as a crash. It now returns `success: false` naming the symlink specifically — "could not write the budget" would send a reader to check permissions when the thing to check is what the path points to.

**Red-proof, one half at a time.** Restoring `existsSync` fails exactly the clear-path test; dropping `O_NOFOLLOW` fails exactly the symlink test. Nineteen of twenty green in each case. Doing them together would have let one fix stand in for the other. The `clear` path had no test at all before this.

**Baseline movement.** `product-identity` for `agent-command` fell 8 → 7 because the new code takes its directory from `BUDGET_FILE` instead of repeating the literal. Re-frozen in the same change as that scan asks; the diff has one row and it moves down.

CI: all checks pass, none pending, none failing. `mergeStateStatus: CLEAN`. No inline review findings. 126 of 129 scans (3 skipped), 265 tests in `agent-command`, typecheck clean.

REVIEWED BASE: 67c1366c4b31b3fbf2bf28d6a795760f6069a647
REVIEWED HEAD: 04e1301f6885ce27a7fed740d0c282f18db3b556
ACTIONABLE FINDINGS: 0

PR #1942, opened 2026-08-20
docs(tasks): complete ARCH-040 and archive its record

woojubb:
Docs-only. Marks ARCH-040 `done` and moves its record to `.agents/tasks/completed/`, which the work itself (PRs #1931, #1938, #1940 — issue #1820 now closed) had left behind.

## What the Result section records

Seven of the ten fields resolved; three split into issues #1934 and #1937 with the blocking cause named. `presetProjection.pendingProjection`: **10 exemptions → 3**.

Two things are written down deliberately rather than tidied away:

**Three of six groups were blocked by a repository gate, and every time the gate was right.** `option-reachability` refused keys declared with nothing assigning them (twice — the hop was wired, the declaration was not deleted). The anti-monolith floor refused the fourth addition to a file holding three interfaces restating one option list, which produced the `IInitOptions` split — and `option-reachability` then fail-closed on that move rather than passing. `product-identity` refused one call because a neutral library's agent field was named after the consumer's product; re-spelling the call to dodge the marker was available, and the field was renamed instead.

**The parked Group B analysis is kept, not deleted.** It named two options — a mutable core field, or splitting a 411-line frozen file — and called them the only two. Both were wrong: the base class already declared `config`, so the getter belonged there and the subclass got *smaller*. A narrow analysis that looked complete is the more useful record for whoever hits the next one.

## Verification

`pnpm harness:scan` — 128 passed, 3 skipped, including `task-archival`, which is what refuses a `done` record left in the active directory.

### DIFF
--- a/.agents/tasks/completed/ARCH-040-preset-projection-decisions.md
+++ b/.agents/tasks/completed/ARCH-040-preset-projection-decisions.md
@@ -1,6 +1,7 @@
 ---
 title: 'ARCH-040: ten resolved-preset fields need a projection DECISION, not just a seam'
-status: in-progress
+status: done
+completed: 2026-08-21
 created: 2026-08-20
 priority: medium
 urgency: soon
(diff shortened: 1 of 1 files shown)


woojubb:
## Pre-merge review

Docs-only: ARCH-040 marked `done` and its record moved to `.agents/tasks/completed/`.

Verified the two things a completion commit can get wrong:

- **The status and the location agree.** `task-archival` is the gate that refuses a `done` record left in the active directory, and it passes — which is also how the omission was caught: the record had been left active while its PR body already claimed the archived path.
- **The Result reports what happened, not what was planned.** Seven of ten fields resolved, three split to filed issues with the blocking cause named, and the exemption count stated as measured (10 → 3) rather than as intended.

Two things are kept deliberately rather than tidied: the record of three gates that blocked this item and were right each time, and the parked Group B analysis that named two options and called them the only two — both wrong. A narrow analysis that looked complete is the more useful record for the next reader than its deletion.

**ACTIONABLE FINDINGS: 0**

PR #1944, opened 2026-08-20
fix(agent-cli): let --task-file, --json-schema and --append-system-prompt reach all three shells

### LINKED ISSUE #1937: --task-file, --json-schema and --append-system-prompt are silently dropped in TUI and serve mode

woojubb:
## What was broken

`buildAppendSystemPrompt` had **exactly one caller** — `print-mode.ts`. So `--append-system-prompt`, `--task-file` and `--json-schema` were parsed, validated, and then dropped in interactive and serve mode, while `cli-args.ts` advertised them with no hint that two surfaces of three ignore them.

Composed once now, inside `buildPresetSurfaceOptions` — the one shape all three shells already read. Calling the helper from two more places would have been a third copy of the same call, the shape ARCH-041 removed from that file.

## The name is the part worth reading

A preset has its **own** `appendSystemPrompt` (`IPresetOptions`), still unprojected and still listed as `pendingProjection`.

My first cut put the CLI text on that name. `preset-projection` fired, and following it up showed what that would have meant: the merge order between preset text and CLI text would have been **decided silently** — CLI wins, preset dropped. That decision belongs to ARCH-040 Group D, and issue #1937 says why it cannot be made yet:

> Fixing the ordering first would encode a rule against a base that is missing two thirds of the time — so this is the prerequisite, not a companion.

A prerequisite that pre-empts the decision it exists to enable is worse than not doing it. So the surface field is `cliAppendSystemPrompt`; the rename onto the session's `appendSystemPrompt` happens one hop later in `toSessionOptions`, with the preset's field untouched until then. Two origins, two names, merge still expressible.

I also removed the `pendingProjection` entry when the scan said the exemption had been earned out — and put it back. What had been declared was my CLI field squatting on the preset field's name, not the preset field being wired.

## Tests assert per surface, because a helper test cannot fail here

The defect is that the helper is **not called**. A test of `buildAppendSystemPrompt` is green in the broken state, which is exactly what the issue's "done when" warns about.

- **print** — runs `runPrintMode` for real; the assertion is on what the provider received.
- **serve** — needed a seam to be assertable without starting a server, so its session options are now built by `buildServeSessionOptions`. That extraction is the point, not a side effect: serve is the surface where a missing forward is least visible, because nothing there prints the prompt.
- **interactive** — through `toSessionOptions`, the only hop it has.

Each of the three was red-proved **alone**: breaking print fails only the print case, breaking the projection only the projection case, breaking serve only the serve case. One fix must not stand in for another.

## Two extractions, because `file-size` refused cli.ts

Both were inline in a thirty-key option literal, where they read as incidental formatting:

- `runShellCommand` — the five-second timeout and piped stdio are the point, and they belong beside the call.
- `resolveCliUpdateNotice` — a promise-that-may-be-absent resolving to a notice-that-may-be-absent, collapsed beside its formatter. Documented in the `agent-framework` SPEC Public API table, because `spec-public-surface` correctly refused a new undocumented export.

`cli.ts` ends **smaller** than it started (465 → 464).

## Verification

Typecheck clean. 1,796 tests pass across `agent-cli` and `agent-framework`. 126 of 129 scans pass (3 skipped).

Closes #1937


woojubb:
## Review

Read the full diff against the base below.

**The thing I checked hardest is the one that changed the design mid-implementation.** `appendSystemPrompt` is a preset field (`IPresetOptions`), not only a session option. My first cut put the CLI-sourced text on that name, which would have decided the merge order between preset text and CLI text silently — CLI wins, preset dropped. `preset-projection` fired on it, and following the finding rather than silencing it is what surfaced the problem: that decision is ARCH-040 Group D's, and this item is the **prerequisite** for it. A prerequisite that pre-empts the decision it exists to enable is worse than not doing it.

So the surface field is `cliAppendSystemPrompt`, renamed onto the session's `appendSystemPrompt` one hop later in `toSessionOptions`, with the preset's field untouched until then. I also verified the `pendingProjection` entry is back: I had removed it when the scan reported the exemption earned out, and what had actually been declared was my CLI field squatting on the preset field's name.

**That the flags really were dropped.** `buildAppendSystemPrompt` had exactly one caller before this — `print-mode.ts`, confirmed by grep across `packages/*/src` excluding tests, where it had no test either.

**That the tests can fail on the condition they name.** This is the issue's own "done when", and it is the reason a helper test is not enough: the defect is that the helper is not CALLED, so a test of it is green in the broken state. Each surface asserts separately, and each of the three hops was red-proved **alone** — breaking print fails only the print case, breaking the projection only the projection case, breaking serve only the serve case.

**That `buildServeSessionOptions` is not scope creep.** Serve had no seam to assert without starting a server, and serve is where a missing forward is least visible because nothing there prints the prompt. The extraction is what makes the surface testable; it is the point, not a side effect.

**The two other extractions** came from `file-size` refusing `cli.ts`, and both were inline in a thirty-key option literal where they read as incidental formatting — a five-second timeout with piped stdio, and a double-optional ternary. `cli.ts` ends smaller than it started (465 → 464), and `spec-public-surface` correctly refused the new `resolveCliUpdateNotice` export until it was documented in the SPEC's Public API table.

CI: all checks pass, none pending, none failing. `mergeStateStatus: CLEAN`. No inline review findings. Typecheck clean; 1,796 tests across `agent-cli` and `agent-framework`; 126 of 129 scans (3 skipped).

REVIEWED BASE: 361d37f62f5f46699cb4d1a128f8c967350b5870
REVIEWED HEAD: aca6dc31e39397a0d97c736036b88763a01df0f8
ACTIONABLE FINDINGS: 0

Issue #1768, opened 2026-08-16
env: packages without their own Volta pin run under Node 24 while the root pins 22, so better-sqlite3's native binding never matches


woojubb:
## Symptom

`pnpm test` fails all 22 tests in `packages/dag-adapters-sqlite`:

```
Error: The module '.../better-sqlite3/build/Release/better_sqlite3.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 127. This version of Node.js requires
NODE_MODULE_VERSION 137.
```

The follow-on `TypeError: Cannot read properties of undefined (reading 'close')` in `afterEach` is a
cascade, not the cause — the adapter constructor threw, so `adapter` was never assigned.

## Cause

Node resolves to a different version depending on the working directory:

```
$ cd /repo && node -v
v22.14.0                # root package.json has "volta": { "node": "22.14.0" }

$ cd /repo/packages/dag-adapters-sqlite && node -v
v24.19.0                # volta default
```

`volta list node` confirms both: `node@22.14.0 (current @ /repo/package.json)` and
`node@24.19.0 (default)`. Volta anchors on the nearest `package.json`; a workspace package without
its own `volta` field does not inherit the root pin here, so it falls through to the default.

`pnpm test` runs vitest per package, i.e. with the package as cwd — so **every** workspace package's
tests run under Node 24 while the repo claims to pin 22. Only `dag-adapters-sqlite` shows it, because
it is the only package with a native addon; ABI mismatches are invisible to pure-TypeScript packages.

Node 22 → `NODE_MODULE_VERSION` 127, Node 24 → 137.

## Why it cannot be worked around locally

- `prebuild-install` has no prebuilt for `better-sqlite3@11.10.0` on `target=24.19.0 runtime=node arch=x64 platform=linux`.
- Building from source is not possible in this environment: `make`, `cc`, `gcc` and `g++ ` are all absent.

So the ABI-137 binding these tests need can be neither downloaded nor compiled, and the ABI-127 one
that *can* be fetched is never the one the test process loads.

## Scope

Environment/toolchain configuration, not product code. `pnpm harness:verify-like-ci` is unaffected.
Found while completing ARCH-031; the seam change is unrelated to it (it fails identically with those
changes stashed).

## Suggested fix

Make the Node version single-valued across the workspace — pin it where every package sees it
(`.node-version`/`engines` honoured by the runner, or a `volta` field propagated to each package)
rather than relying on the root `package.json` being the nearest one.

[...]

PR #1948, opened 2026-08-20
fix(harness): the Node-pin guard could not see the packages that were unpinned (INFRA-102)

woojubb:
Issue #1768. The pins and the scan were **both already in place**, and the workspace was still not single-valued.

## What the guard could not see

`scan-node-version-single-valued` enumerated `packages/*` and `apps/*` **one level deep**. `packages/dag-nodes/*` is a NESTED group declared in `pnpm-workspace.yaml`, so its twenty manifests were outside the population entirely.

Measured on this tree before the fix:

- 20 of 20 nested manifests carried no `volta` field
- every one resolved to **Node 24.19.0** against a root declaring **22.14.0**
- each has a `test` script, so `pnpm test` ran them on an undeclared runtime — the exact condition this item exists to remove
- the scan reported the workspace single-valued over **67 of its 87** manifests

**Red-proofed rather than argued:** restore the one-level walk *and* delete a nested pin, and the scan reports `passed`. A guard whose population excludes the failures is a guard that cannot fire — the same shape as the defect it guards, one level up.

## The fix is delegation, not a second glob

`scripts/harness/workspace-packages.mjs` exists for precisely this class. Its own header records it:

> several harness scans used a one-level `readdirSync` loop and therefore silently skipped NESTED package groups (INFRA-021)

This scan was one more copy. It now calls `listWorkspacePackageDirs`; the declared size goes 67 → 87.

The twenty manifests then get `volta.extends`. Appended as **text** rather than round-tripped through `json.dumps`, which re-encoded the em-dash in seven `description` fields to `—` — changing lines this edit had no business touching.

## Verified by running it, not by reading the config

`cd packages/dag-nodes/file-read && node -v` now reports `v22.14.0` where it reported `v24.19.0`, and that package's suite passes (2 files, 9 tests) on the declared runtime.

## One of the new cases was wrong first

`resolves a nested package through its three-level extends` mutated the manifest **path** instead of the `extends` **target**, so it compared a package with itself and passed either way. Rewritten around a second fixture whose `extends` is the `../../` a copy-paste from a top-level sibling produces — which resolves to the group directory, where no manifest exists.

| mutation | fails |
|---|---|
| the one-level population restored | exactly 3 — every new nested case |
| one nested pin deleted | exactly 1, naming that manifest |

## Verification

- `npx vitest run scripts/harness/__tests__/` — 225 files, 4277 tests, all passed

### DIFF
--- a/.agents/tasks/completed/INFRA-102-node-version-is-not-single-valued-across-the-workspace.md
+++ b/.agents/tasks/completed/INFRA-102-node-version-is-not-single-valued-across-the-workspace.md
@@ -1,6 +1,7 @@
 ---
 title: 'INFRA-102: the Node version is not single-valued across the workspace, so every package tests on an unpinned runtime'
-status: in-progress
+status: done
+completed: 2026-08-21
 created: 2026-08-17
 priority: high
 urgency: now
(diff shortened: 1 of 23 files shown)


woojubb:
## Pre-merge review

**base:** `a7a8a2f202126bd9c2e876b2fd526a6017188406`
**head:** `5d5c992e49355cbc966addfa39ce3e779c88d3a6`

CI green, no review threads.

### The finding is the shape, not the twenty missing pins

Both halves of this item — the `volta.extends` pins and the `node-version-single-valued` scan — were already in place, and the workspace was still not single-valued. The scan enumerated one level deep and `packages/dag-nodes/*` is a nested group, so twenty manifests running on **Node 24.19.0** against a root declaring **22.14.0**, each with a `test` script, were outside the population. It reported single-valued over 67 of 87.

Red-proofed: restore the one-level walk *and* delete a nested pin, and the scan says `passed`.

`scripts/harness/workspace-packages.mjs` already exists for this exact class and says so in its own header (INFRA-021). This scan was one more copy of the defect it was written to end, which is why the fix is to delegate rather than to add a second glob.

### Two things I got wrong first, both caught by checking rather than reading

- The pins were first written by `json.dumps` round-trip, which re-encoded the em-dash in seven `description` fields to `—`. Rewritten as a text append: 60 insertions, **0 deletions**.
- One new case mutated the manifest **path** instead of the `extends` **target**, so it compared a package with itself and passed either way. Rewritten around a fixture whose `extends` is the `../../` a copy-paste from a top-level sibling produces.

### Verified by running it

`cd packages/dag-nodes/file-read && node -v` reports `v22.14.0` where it reported `v24.19.0`, and that package's suite passes on the declared runtime.

- `npx vitest run scripts/harness/__tests__/` — 225 files, 4277 tests, all passed

**ACTIONABLE FINDINGS: 0**

PR #1950, opened 2026-08-20
test(ci): DO NOT MERGE — INFRA-047's GPL fixture, to be read and closed

woojubb:
## ⛔ This PR exists to be read and closed. Do not merge.

INFRA-047 migrated `dependency-review` from `deny-licenses` to `allow-licenses` in #1339, and named **one** closing condition in its Test Plan that was never run:

> A test PR introducing a GPL dev-dep is BLOCKED under the new input (then closed).

The item recorded why it could not run pre-merge — `dependency-review` executes the config of the PR's *merge result* against the target branch, so the new allow-list only gates PRs opened **after** it landed. It landed on 2026-07-24. This is that PR.

Until it runs, "the allow-list actually blocks copyleft ingress" is an argument about a config file, not a measurement — which is precisely the shape the Test Plan was written to avoid.

## The fixture

`@substrate/connect-extension-protocol@2.2.2` — `GPL-3.0-only`, **zero dependencies**. Chosen for the zero: the lockfile moves by seven lines, so the check measures the license rule rather than a wave of transitive churn.

## Expected result

**`Dependency review` FAILS**, naming `GPL-3.0-only` as not in the allow-list.

A pass here is a defect in the allow-list, not a green.

The job is advisory (in neither branch ruleset's required checks) and `paths`-filtered on `**/package.json` and `pnpm-lock.yaml`, so this PR triggers it and blocks nothing.

Once the check reports, its output goes into the INFRA-047 record and this PR is closed and its branch deleted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,6 +147,7 @@
   "author": "Robota SDK Team",
   "license": "AGPL-3.0-only OR LicenseRef-Commercial",
   "devDependencies": {
+    "@substrate/connect-extension-protocol": "2.2.2",
     "@changesets/changelog-github": "^0.7.0",
     "@changesets/cli": "^2.31.1",
     "@commitlint/cli": "^21.2.1",
--- a/package.json
+++ b/package.json
@@ -147,7 +147,6 @@
   "author": "Robota SDK Team",
   "license": "AGPL-3.0-only OR LicenseRef-Commercial",
   "devDependencies": {
-    "@substrate/connect-extension-protocol": "2.2.2",
     "@changesets/changelog-github": "^0.7.0",
     "@changesets/cli": "^2.31.1",
     "@commitlint/cli": "^21.2.1",
@@ -180,6 +179,7 @@
   },
   "packageManager": "pnpm@8.15.4",
   "dependencies": {
+    "@substrate/connect-extension-protocol": "2.2.2",
     "@modelcontextprotocol/sdk": "^1.29.0",
     "zod": "^3.25.76"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -59,6 +59,9 @@ importers:
       '@modelcontextprotocol/sdk':
         specifier: ^1.29.0
         version: 1.29.0(zod@3.25.76)
+      '@substrate/connect-extension-protocol':
+        specifier: 2.2.2
+        version: 2.2.2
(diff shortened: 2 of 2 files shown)


woojubb:
## Measured. Closing — and the result is not the one the Test Plan predicted.

Two runs, **same package, same version, same license, same lockfile entry**. The only difference is the dependency SCOPE.

| form | `Licenses` group | check |
|---|---|---|
| `devDependencies` (`e775fd0f9`) | **empty** | ✅ SUCCESS |
| `dependencies` (`8c45caf6b`) | `@substrate/connect-extension-protocol@2.2.2 – License: GPL-3.0-only` | ❌ **FAILURE** |

### What this establishes

**The allow-list does block copyleft ingress.** INFRA-047's substantive closing condition is satisfied: a GPL-3.0-only package added as a runtime dependency is refused by name and license.

**But INFRA-047's Test Plan named a case that cannot fail.** It reads:

> A test PR introducing a GPL **dev-dep** is BLOCKED under the new input (then closed).

A dev-dep is never blocked. `dependency-review-action` defaults `fail-on-scopes` to `runtime`, and `.github/workflows/dependency-review.yml` sets no override — so a development-scoped package is exempt from the license gate whatever its license is. Run literally, the item's own plan would have produced a green and it would have meant nothing.

The first run is the evidence for that, not a stumble: the `Licenses` group came back **empty** rather than reporting an allowed license, which is what "this dependency was not evaluated" looks like.

### Why this needed two runs

After the first green there were two candidate causes — scope, or GitHub not knowing the package's license (the empty `Licenses` group and `OpenSSF Scorecard Score: undefined` were consistent with both). Guessing between them would have left the measurement saying nothing. Moving the same package to a runtime dependency separates them, and it did: GitHub knew the license all along.

### Filed, not folded in

That development-scope deps bypass the license gate is a property of the workflow, not of INFRA-047, and it is nowhere stated in the workflow that documents everything else about this policy at length. Filed separately.

This also closes INFRA-061's second unfalsified check — its table records `dependency-review.yml` as "No — hypothesis", and this is the planted defect it asked for. Its suggested probe was `pako` (`MIT AND Zlib`); a real GPL package turned out to be the stronger one, because it also exercises the copyleft case the policy exists for.

Closing unmerged, as designed. Branch deleted.

Issue #1951, opened 2026-08-20
dependency-review's license allow-list does not gate DEVELOPMENT dependencies, and nothing says so


woojubb:
## Measured

Two runs on PR #1950, **same package, same version, same license, same lockfile entry** — only the dependency scope differs:

| form | `Licenses` group | check |
|---|---|---|
| `devDependencies` | **empty** | SUCCESS |
| `dependencies` | `@substrate/connect-extension-protocol@2.2.2 – License: GPL-3.0-only` | **FAILURE** |

`@substrate/connect-extension-protocol@2.2.2` is `GPL-3.0-only`, which is not on the allow-list.

## Cause

`actions/dependency-review-action@v5` defaults `fail-on-scopes` to `runtime`, and `.github/workflows/dependency-review.yml` sets no override. A development-scoped package is therefore exempt from the license gate whatever its license is — the empty `Licenses` group in the first run is what "not evaluated" looks like, as distinct from "evaluated and allowed".

## Why this is worth an issue rather than a one-line fix

**The behaviour may well be correct**, and that is exactly why it needs a decision rather than a patch. This repository is dual-licensed (AGPL + Commercial, no CLA), and the hazard the gate exists for is *distribution*: a copyleft package that ships in a published artifact would virally break the commercial-relicensing right. A dev-only dependency — a linter, a test runner — is not distributed, so exempting it is defensible policy.

What is not defensible is that it is **undeclared**. `.github/workflows/dependency-review.yml` documents this policy at unusual length: why MPL-2.0 is on the list, why the `@img/sharp-win32-*` trio is exempted and the other wrappers are not, which SPDX operator semantics apply. It does not say that a large part of the dependency graph is outside the gate.

The concrete evidence that this was not understood: **INFRA-047's Test Plan specified a case that cannot fail.**

> A test PR introducing a GPL **dev-dep** is BLOCKED under the new input (then closed).

Run literally, that produces a green that means nothing. It is the item's only stated closing condition, and it was written by someone who believed dev deps were gated.

## What to decide

1. **Keep the exemption and state it** — set `fail-on-scopes: runtime` explicitly, making the default a choice rather than an inheritance, with the distribution reasoning beside the rest of the policy commentary.
2. **Extend the gate** — `fail-on-scopes: runtime, development`. Stricter, and it would need the current dev graph measured first: the allow-list closure of 2026-07-25 covered prod + dev, so this may be close to free, or may surface existing dev-only copyleft.

[...]

PR #1952, opened 2026-08-20
test(ci): falsify the last two checks INFRA-061 left as hypotheses, and close INFRA-047

woojubb:
Two backlog items whose only remaining work was **running the falsification they had each specified and never performed**. Both are now measurements, and both measurements found something the record did not predict.

## dependency-review — falsified, and INFRA-047's own Test Plan was wrong

From throwaway PR #1950, opened against `develop` and closed unmerged, carrying `@substrate/connect-extension-protocol@2.2.2` — `GPL-3.0-only`, **zero dependencies**, so the lockfile moved by seven lines.

It took two runs. Same package, same version, same license, same lockfile entry; only the dependency **scope** differs:

| form | `Licenses` group | check |
|---|---|---|
| `devDependencies` | **empty** | ✅ SUCCESS |
| `dependencies` | `@substrate/connect-extension-protocol@2.2.2 – License: GPL-3.0-only` | ❌ **FAILURE** |

**The allow-list does block copyleft ingress** — INFRA-047's substantive condition is satisfied.

**But its Test Plan named a case that cannot fail.** It reads:

> A test PR introducing a GPL **dev-dep** is BLOCKED under the new input (then closed).

A dev-dep is never blocked: the action defaults `fail-on-scopes` to `runtime` and the workflow sets no override. Run literally and stopped at the first green, the item would have recorded a pass that meant nothing — or concluded the allow-list was broken. The **empty** `Licenses` group is the evidence: that is what *not evaluated* looks like, as distinct from *evaluated and allowed*.

Two runs were needed because after the first green there were two candidate causes — scope, or GitHub not knowing that package's license — and the output (`Licenses` empty, `Scorecard Score: undefined`) was consistent with both. Guessing would have left the measurement saying nothing.

Filed as **#1951**, not fixed here: choosing between "state the exemption" and "extend the gate" is a licence-policy decision. INFRA-061 had rated this row **"Matches"** on structural evidence; falsifying it is what corrected that.

## codeql — falsified by a defect nobody planted

PR #1945 carried four CodeQL alerts annotated on lines it had just added, one of them (`js/unnecessary-use-of-cat`) blocking `review-gate`. That exercises both halves of the path INFRA-061 said was unproven — the analysis reports, and the alerts reach the PR.

An unplanted defect is the stronger evidence, because a planted one is written to be found.

**No check in that audit now rests on a hypothesis.**

## A third defect, found by running the guard on the records this change wrote

`unearned-done-claims` reported an "uncited evidence section" in INFRA-061 — which was a **shell comment inside a fenced block**:

```
# 1. genuine bytes — the verification CI performs
```

`sectionsOf` read every line, so that became an H1. U3's *body* loop already skipped fences and its own comment says so; the heading collector one loop earlier did not — the same file disagreeing with itself about what a heading is. One `outsideFences` helper now serves both.

A finding on correct work is a reason to change the check, which is what `enforcement-architecture.md` says.

## Verification

- `pnpm harness:scan` — 128 passed, 3 skipped
- `npx vitest run scripts/harness/__tests__/` — 225 files, 4282 tests, all passed
- Red proofs: `sectionsOf` reading fences again fails exactly 2 (including the real-corpus case); removing the U3 collector's skip fails exactly 1
- The `reference-kind` baseline **fell** by two rows and gained none


woojubb:
## Pre-merge review

**base:** `2367967e6528846e6539f45b00edfb60fdafb820`
**head:** `0abcd58929bb1e14fc3611272b85bc71ff7c8a97`

CI green, no review threads.

### Both falsifications found something the record did not predict

That is the argument for doing them rather than closing the items on structural evidence:

- **`dependency-review`** was rated **"Matches"** in INFRA-061's table. Falsifying it showed the license gate does not evaluate development-scoped dependencies at all — and that INFRA-047's only stated closing condition (`a GPL dev-dep is BLOCKED`) names a case that cannot fail. Filed as #1951; the substantive claim (copyleft ingress *is* blocked for runtime deps) is now measured.
- **`codeql`** was falsified by a defect nobody planted — four alerts on lines PR #1945 had just added, one of them blocking the merge gate. A planted defect is written to be found; this one was not.

### The second run is the load-bearing one

After the first green there were two candidate causes and the output was consistent with both (`Licenses` empty *and* `Scorecard Score: undefined`). Stopping there would have recorded a pass that meant nothing. Moving the same package to a runtime dependency separated them.

### A third defect, in the guard that judges this work

`unearned-done-claims` reported an uncited evidence section that was a **shell comment inside a fenced block**. `sectionsOf` read every line; U3's body loop already skipped fences and said so in its own comment, while the heading collector one loop earlier did not. One `outsideFences` helper now serves both — a finding on correct work is a reason to change the check.

### Verification

- `pnpm harness:scan` — 128 passed, 3 skipped
- `npx vitest run scripts/harness/__tests__/` — 225 files, 4282 tests, all passed
- Red proofs: `sectionsOf` reading fences again fails exactly 2 (including the real-corpus case); removing the U3 collector's skip fails exactly 1
- The `reference-kind` baseline **fell** by two rows and gained none

**ACTIONABLE FINDINGS: 0**

Issue #1949, opened 2026-08-20
The conditional-spread idiom for exactOptionalPropertyTypes suppresses excess-property checking, at 446 sites


woojubb:
## The defect

The conditional-spread idiom this repository uses for `exactOptionalPropertyTypes` **suppresses excess-property checking**, so a key placed on the wrong object is invisible to the compiler.

Measured with a probe compiled by this repo's own `tsgo`, inside `agent-session`:

```ts
interface ITarget {
  config: { permissions: string[]; hooks?: Record<string, unknown> };
  topLevel?: string;
}
const v: string | undefined = 'x';

const a: ITarget = { config: { permissions: [], topLevel: v } };                                  // ← errors
const b: ITarget = { config: { permissions: [], ...(v !== undefined ? { topLevel: v } : {}) } };  // ← SILENT
```

```
src/__excess-probe.ts(8,49): error TS2353: Object literal may only specify known properties,
  and 'topLevel' does not exist in type '{ permissions: string[]; hooks?: … }'.
```

Line 8 is caught. Line 11, the same mistake through a spread, is not — excess-property checking applies to object literals, and a spread result is not one.

## How it was found

Not from a probe. Review caught it on PR #1946: `presetFreePermissions` was passed **inside** `config` while the constructor reads it at the top level, so on the only wiring that matters it was always `undefined` and the enforcer silently fell back to the contaminated rules. That reproduced, through the wiring, the exact bug the change existed to fix.

Both the type system and every existing test stayed green:

- `tsgo` — suppressed, as above;
- the enforcer's own tests — they construct `PermissionEnforcer` directly and never traverse `buildPermissionEnforcer`, so the wiring was not under test.

## Scale

```
446 occurrences across 132 files under packages/*/src
```

(`...(x !== undefined ? { k: v } : {})` and near variants.) Every one of them is a place where a key on the wrong object compiles.

This is not an argument against the idiom — `exactOptionalPropertyTypes` is why it exists, and assigning `undefined` to an optional property is a real error it prevents. The point is that it costs a check nobody accounted for, and the cost is invisible until something like this happens.

## What might be done

Unchosen; listing what I considered:

1. **A lint rule** on spreads into a typed literal, requiring the spread's key set to be declarable on the target. Narrow, mechanical, and the only one that scales to 446 sites.
2. **A helper** — `defined({ k: v })` or similar — that keeps the exactness property while producing a literal the checker can see. Changes 446 call sites.
[...]

PR #1953, opened 2026-08-20
docs(memory): a conditional spread is not an object literal, so its keys stop being checked

woojubb:
## The fact

`exactOptionalPropertyTypes` makes `...(x !== undefined ? { k: v } : {})` the idiom for an optional key in this repository. **Excess-property checking applies to object literals, and a spread result is not one** — so a key placed on the wrong object compiles silently.

Verified with a probe compiled by this repository's own `tsgo`:

```ts
const a: ITarget = { config: { permissions: [], topLevel: v } };                                  // TS2353
const b: ITarget = { config: { permissions: [], ...(v !== undefined ? { topLevel: v } : {}) } };  // silent
```

Worth writing down because the natural assumption runs the other way: the repository is strict, so the compiler is presumed to be watching. On this construct it is not, and nothing said so anywhere.

## How it was found, and why that matters for the record

Not by probing. Review caught it on PR #1946 — an options key passed **inside** `config` while the constructor read it at the top level, so production always fell back and reproduced the bug the change existed to fix. Both gates that should have caught it were green: the compiler for the reason above, and the tests because they constructed the object directly and never traversed the composition root.

## The measurements corrected each other twice, and both corrections are in the file

| claim | outcome |
| --- | --- |
| "446 exposed sites" (as I filed issue #1949) | wrong — counted test files; shipped `src` has 356 |
| "99% are nested" (first cut) | **wrong, and how it was wrong is the useful part** — it counted every `{`, so a function body made every spread look nested. A result reading "essentially all of them" is a broken question, not an answer |
| exposed = spread inside an object that is a property VALUE | **22 of 356 (6%)** |
| any of the 22 misplaced? | **none** — read all 22; every key belongs to the object it is spread into |

So the gap is real and latent: one historical instance, caught by review while the change was still open, and zero in the tree. Recorded rather than mechanised, which is what those numbers argue for.

## Two options tested rather than assumed

- **A `defined()` helper does NOT restore the check.** It changes what is spread, not that it is a spread. Measured: still silent. This refutation is the more useful half — it would have been a 356-site migration buying nothing on this axis.
- **Annotating the spread SOURCE does work**, because the check then runs on the inner literal. Only applies where the spread is hoisted to a named binding, so it is a tool for a composition boundary rather than a sweep.

A type-aware lint rule is the only thing that scales, and it is unavailable here: `.eslintrc.json` has no `project` in `parserOptions` and zero type-aware rules enabled.

## Verification

Memory files only; no source change. `memory-mirror` and `reference-kind-qualified` both pass; 126 of 129 scans (3 skipped); `verify-like-ci` green on all 12 stages.

Investigation for issue #1949.


woojubb:
## Review

Memory files only; no source change. The automated review reports `ACTIONABLE FINDINGS: 0` at this head and there are no inline threads.

**What I checked is whether each claim in the file is true**, since that is all a recorded fact is worth:

- **The compiler behaviour** — probed in `agent-session` with this repo's own `tsgo`, both directions. The same misplaced key errors `TS2353` written directly and compiles written through the spread. Not inferred from documentation.
- **356, not 446** — the filed figure counted test files. Corrected in the file rather than quietly dropped, because the wrong number is what the issue currently says.
- **22 of 356** — and the first cut that reported 99% is written down as a wrong *question*, not a wrong answer. It counted every `{`, so any function body made every spread look nested. That correction is the part most likely to save someone else the same hour.
- **None of the 22 misplaced** — read all 22. `create-subagent-session.ts` is the one worth naming: `allow`/`deny` inside `taskPermissions` is structurally identical to the bug and is correct, checked against the declared `{ allow?: readonly string[]; deny?: readonly string[] }`.
- **The `defined()` refutation** — measured, not assumed. It would have been a 356-site migration buying nothing on this axis, and the issue's own option list would have sent someone at it.

**The scope claim is honest**: this records a latent gap with zero current instances, and says so, rather than presenting a repo-wide danger. I filed the issue believing the opposite; the file says what the measurements say, not what the filing said.

`memory-mirror` passes (31 fact files, indexed), `reference-kind-qualified` passes, 126 of 129 scans (3 skipped), `verify-like-ci` green on all 12 stages.

REVIEWED BASE: 2367967e6528846e6539f45b00edfb60fdafb820
REVIEWED HEAD: 65e8ef787b489c0614d9e3111d2e0ed1819bf43e
ACTIONABLE FINDINGS: 0

woojubb:
## Review

Memory files only; no source change. The automated review reports zero actionable findings at this head, and there are no inline threads.

**What I checked is whether each claim in the file is true**, since that is all a recorded fact is worth:

- **The compiler behaviour** — probed in `agent-session` with this repo's own `tsgo`, both directions. The same misplaced key errors `TS2353` written directly and compiles written through the spread. Not inferred from documentation.
- **356, not 446** — the filed figure counted test files. Corrected in the file rather than quietly dropped, because the wrong number is what the issue currently says.
- **22 of 356** — and the first cut that reported 99% is written down as a wrong *question*, not a wrong answer. It counted every `{`, so any function body made every spread look nested. That correction is the part most likely to save someone else the same hour.
- **None of the 22 misplaced** — read all 22. `create-subagent-session.ts` is the one worth naming: `allow`/`deny` inside `taskPermissions` is structurally identical to the bug and is correct, checked against the declared `{ allow?: readonly string[]; deny?: readonly string[] }`.
- **The `defined()` refutation** — measured, not assumed. It would have been a 356-site migration buying nothing on this axis, and the issue's own option list would have sent someone at it.

**The scope claim is honest**: this records a latent gap with zero current instances, and says so, rather than presenting a repo-wide danger. I filed the issue believing the opposite; the file says what the measurements say, not what the filing said.

`memory-mirror` passes (31 fact files, indexed), `reference-kind-qualified` passes, 126 of 129 scans (3 skipped), `verify-like-ci` green on all 12 stages.

REVIEWED BASE: 2367967e6528846e6539f45b00edfb60fdafb820
REVIEWED HEAD: 65e8ef787b489c0614d9e3111d2e0ed1819bf43e
ACTIONABLE FINDINGS: 0

woojubb:
## Merged with `MERGE_GATE_ACK=1`, and here is exactly what the gate could not verify

The gate refused with `found 1/1/2`: the review it selected carries **two** `ACTIONABLE FINDINGS` markers where it requires one.

That is not this branch's doing. The automated review's own body emitted the marker twice — once in its summary line and once in the footer block:

```
3: ACTIONABLE FINDINGS: 0
5: REVIEWED BASE: 2367967e6528846e6539f45b00edfb60fdafb820
6: REVIEWED HEAD: 65e8ef787b489c0614d9e3111d2e0ed1819bf43e
7: ACTIONABLE FINDINGS: 0
```

The gate reads only `github-actions` verdicts (`REVIEWER_RE`), so nothing I post can change that count — my own verdict comment is not a candidate for selection at all.

**What I verified by hand, which is what an override is for:**

- both markers read `0`, so there is no ambiguity about the verdict — the duplication is cosmetic, not two different answers;
- `REVIEWED BASE` and `REVIEWED HEAD` both match the merge base and head exactly (`2367967e6` / `65e8ef787`);
- `mergeStateStatus: CLEAN`, every check passing, none pending;
- zero inline review threads, so nothing was left unanswered;
- `verify-like-ci` green on all 12 stages, 126 of 129 scans (3 skipped).

The gate's rule is right — "exactly one" is what stops a stale marker from lending its number to a newer verdict. It fired on a reviewer-side duplication it has no way to distinguish from that case, which is the correct conservative reading.

Worth a follow-up on the review workflow rather than on this gate: a verdict that prints its own count twice will block every merge it touches.

PR #1954, opened 2026-08-20
chore(tasks): complete PEER-004 and PEER-006, whose work merged as issue #1863

woojubb:
## What was stale

`PEER-004` and `PEER-006` sat at `status: in-progress` in the backlog root while their work was on `develop` and issue #1863 was closed. That is the shape `backlog-placement` exists to catch, sitting one status field away from being caught — the scan reports a terminal-status file in the root, not an open-status file whose work has shipped.

## PEER-006's TC-07, and why the tick cites a different change

It was left open on purpose, and the reason was right when it was written:

> the receiving operator sees the peer origin rendered … Left open deliberately: ticking it would claim a rendering nobody in this change measured.

PEER-007 (issue #1915, PR #1935) then measured exactly it: the receiving session rendered `peer:<sender-session-id>` while a control session rendered `You:`, and the two labels were mutually exclusive across the transcripts — the peer-driven one had zero `You:`, the operator one zero `peer:`.

So the tick is on that measurement and the entry says so. A tick that cites the wrong change is how an unmeasured claim acquires a checkbox, which is the thing the original author was avoiding.

## The user-execution evidence

Captured 2026-08-20 by driving two real `pnpm cli:dev` sessions through PTYs on one host as one user, including **both directions** — each session labelled the *other* one, because a single direction would pass even with the label wired to the wrong end. Merged with issue #1863; this change only moves the records.

## One thing worth flagging about the commit

`git mv` moved the *index* version, so the `status: done` edit was not in the first commit — `rename 100%` was the tell, and `git status` still showed both files modified afterwards. Amended so the frontmatter change and the move land together, which the Completion Steps require as a single atomic act.

## Verification

`backlog-placement` and `task-archival` both pass (115 active task files, 875 archived). 126 of 129 scans (3 skipped). `verify-like-ci` green on all 12 stages.

### DIFF
--- a/.agents/tasks/completed/PEER-004-local-discovery-reaches-the-operator.md
+++ b/.agents/tasks/completed/PEER-004-local-discovery-reaches-the-operator.md
@@ -1,7 +1,8 @@
 ---
 title: 'PEER-004: local peer discovery was built and called by nothing'
-status: in-progress
+status: done
 created: 2026-08-19
+completed: 2026-08-21
 priority: high
 urgency: now
 area: packages/agent-cli, packages/agent-command, packages/agent-framework
--- a/.agents/tasks/completed/PEER-006-peers-send-reaches-the-other-session.md
+++ b/.agents/tasks/completed/PEER-006-peers-send-reaches-the-other-session.md
@@ -1,7 +1,8 @@
 ---
 title: 'PEER-006: the carrier and the ingress exist, and nothing joins them'
-status: in-progress
+status: done
 created: 2026-08-20
+completed: 2026-08-21
 priority: high
 urgency: now
 area: packages/agent-cli, packages/agent-command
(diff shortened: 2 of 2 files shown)


woojubb:
## Review

Task-record lifecycle only; no source change. The automated review reports zero actionable findings at this head, with one marker, and there are no inline threads.

**What I checked:**

- **That the work really is done.** Both records' user-execution evidence is on `develop` (verified by reading it there, not locally), issue #1863 is closed on it, and neither record has an unchecked plan item left.
- **That TC-07's tick is honest.** It was left open by PEER-006 on purpose — *"ticking it would claim a rendering nobody in this change measured"* — and it is ticked here on PEER-007's measurement, with the entry naming that change. The distinction matters: a tick citing the wrong change is how an unmeasured claim acquires a checkbox, which is exactly what the original author was avoiding.
- **That the commit is atomic.** The Completion Steps require the frontmatter change and the move in one commit. My first attempt was not: `git mv` moved the *index* version, so `status: done` was left unstaged and the diff read `rename 100%` with zero insertions. `git status` still showing both files modified is what caught it; amended so both land together.
- **That nothing else is stale.** Checked every task record touched by today's merges — `INFRA-124` and `INFRA-125` are already in `completed/`, and no other record for a closed issue remains in the root.

`backlog-placement` and `task-archival` both pass (115 active, 875 archived), 126 of 129 scans (3 skipped), `verify-like-ci` green on all 12 stages.

REVIEWED BASE: 42556f1f09a1031b1d381a7cf46bfe6ecdf7ef86
REVIEWED HEAD: 85c300ed6167c2077bb4799ed400ef7755922693
ACTIONABLE FINDINGS: 0

PR #1961, opened 2026-08-21
feat: close the remaining INFRA backlog — no-floating-promises, actionlint, patch-coverage, and four precise states

woojubb:
The tail of the INFRA backlog. Four items close, four get an accurate state instead of a close, and one owner directive turned out to govern three of them.

## INFRA-040 — no-floating-promises, monorepo-wide

Deferred for "the resulting type-aware-finding flood". **Measured: SIX violations across 55 packages.** Same shape as INFRA-124, where "hundreds" was thirteen — the reason to defer did not survive being counted.

All six were real; two were production defects:

| site | what it was |
|---|---|
| `apps/agent-server/src/server.ts` | `startServer()` on the entry point with nothing to report a throw from its own catch arm |
| `packages/agent-plugin/.../webhook-queue.ts` | `drainBatch` inside `setInterval` — a delivery failure rejected on **every tick** with no caller to propagate to |
| `packages/agent-playground/.../remote-injection-sandbox.ts` | a sync `cleanup` returning **before** the `dispose()` it started |

A seventh appeared once the wiring reached `apps/agent-app`: `app.whenReady().then(createWindow)` in the Electron main. If window creation failed the app had no window **and no message**.

**The wiring is where the cost actually was**, and three mechanisms were measured before one worked — including this item's own step 3, which is **not achievable**: when a package config extends the root, the root's `overrides.files` globs resolve relative to the *package*, so `packages/**/*.ts` never matches. Verified with `eslint --print-config`. Recorded as unachievable rather than left pending.

## INFRA-065 — actionlint, and a job that argued with itself

**The third red proof produced nothing, and that is the finding.** A shellcheck-visible defect drew no comment and actionlint exited 0:

```
Rule "shellcheck" was disabled: exec: "shellcheck": executable file not found in $PATH
```

Half of what the job is for turns itself off silently while the job still passes. It now asserts shellcheck. One `command -v` away from shipping a check that could disable half of itself and still report success.

**Two scans then argued opposite sides.** `ci-concurrency-footprint` refused the standalone job (an account-budgeted slot); folding it into `scans` was refused by `ci-mirror-map` and `pre-push-mirrors-ci-scans`, because every step of a *required* job must be locally reproducible — INFRA-056's whole point. The mirror invariant wins; the footprint baseline is re-frozen deliberately at 24. **The first commit of this PR argues the opposite**, and the second corrects it.

## INFRA-094 — patch-coverage stops rebuilding

Restores `build`'s `package-dist` instead of running its own `pnpm build:deps`. Option (a) is unavailable (see below). The trade is stated, correcting my own first draft: the item records `patch-coverage` 206s **and `build` 200s in the same runs**, so the compute removed is ≈ a whole build and the serialisation added is ≈ the same wait. Fine for an advisory verdict; not what "cuts the job's time by the build share" implies.

## The 2026-08-04 directive governs three items

「크론은 다 꺼」 removed every `schedule:` trigger; there are none today. So **INFRA-042** item 3 and **INFRA-065** half 2 are *refused, not pending*, and **INFRA-094** option (a) is unavailable. Each records that rather than leaving a checkbox open.

## Four items get a state, not a close

- **INFRA-046** `blocked` — one ruleset `PUT` remains, and on the item's own evidence it should *not* run yet: `accidental-green` has never fired on a real PR, so requiring it would put an untested refusal in the merge path.
- **INFRA-054** `blocked` on three decisions it enumerates itself.
- **INFRA-097** `blocked` — all three routes to a trusted control plane need config outside this repo, and there is no fourth.
- **INFRA-104** stays `in-progress`. The guard **has** run (promotion PR #1895) and produced a legitimately empty block. TC-10 wants GitHub — not a person — closing an issue, and there is no candidate today. Recorded that *this session consumed four candidates* by hand-closing issues, so the next agent does not repeat it unaware.

## Verification

- `pnpm -r lint` — exit 0, 0 parsing errors, 0 floating promises, 207s
- `pnpm harness:test` — 226 files / 4319 tests + 73 / 1113, all passed
- `actionlint` over the tree including its own new job — exit 0
- The four packages INFRA-040 edited: 1159 tests, all passed

The last commit fixes a miss the pre-push gate caught: I verified that rollout with lint and **not** with the test suites of the packages it edited. Lint reads types; it does not compile the file the way vitest does.


woojubb:
## Pre-merge review

**base:** `4dbb024b56cdf9d29c8052db9544f9cdd4a7ccee`
**head:** `8008e37ae3100ee06bd8fac138e1140edbbe325d`

CI green, no review threads.

### The most useful result is that the recorded numbers were mostly wrong

Four of these items were deferred or scoped on figures that did not survive being re-measured:

| item | recorded | measured |
|---|---|---|
| INFRA-040 | "type-aware finding **flood**" | **6** violations across 55 packages |
| INFRA-094 | "cuts the job's time by the build share" | `build` was **200s** in the same runs — so the serialisation added ≈ the compute removed |
| INFRA-065 half 2 / INFRA-042 item 3 | pending work | **refused** — the 2026-08-04 directive removed every `schedule:` trigger |
| INFRA-046 | blocked on a credential | blocked on **evidence first**: `accidental-green` has never fired on a real PR |

Implementing INFRA-040 from its own description would have been a ~100-project migration against a problem that was six lines.

### I was wrong twice in this PR, and both are in the diff

- **`c2` argued the opposite of `c3`.** Folding actionlint into `scans` looked strictly better (no concurrency slot) until `ci-mirror-map` and `pre-push-mirrors-ci-scans` refused it — every step of a *required* job must be locally reproducible, which is INFRA-056's entire point. The mirror invariant outranks the footprint baseline; the baseline is re-frozen deliberately.
- **I took the actionlint baseline with shellcheck absent** — under the exact silent skip the job asserts against, while writing the assertion. "ZERO problems" was a statement about my machine. CI found three things.

That second one paid for itself. The finding was `# shellcheck disable=SC2086 -- reason`, a form shellcheck **cannot parse** (SC1073, "Fix to allow more checks") — so a directive written to narrow one rule had been aborting *every* check on its block. Verified against shellcheck directly: `-- reason` fails, `# reason` and the bare form are clean. Swept for siblings: none elsewhere, and `shellcheck -S error` over all 28 tracked shell scripts exits 0.

Recorded as `common-mistakes` entry 89, with `Mechanism: none` argued rather than asserted — the degradation is inside a third-party tool and reports success.

### And one the pre-push gate caught

INFRA-040 was verified with `pnpm lint` and **not** with the test suites of the packages it edited. Lint reads types; it does not compile the file the way vitest does. One `await` added without making its `it()` async.

### Four items get a state, not a close

INFRA-046 / 054 / 097 are `blocked` with the specific decision named; INFRA-104 stays `in-progress` because TC-10 needs an observation that has no candidate today — and the record notes this session *consumed* four candidates by hand-closing issues, so the next agent does not repeat it unaware.

### Verification

- `pnpm -r lint` — exit 0, 0 parsing errors, 0 floating promises
- `pnpm harness:test` — 226 files / 4319 tests + 73 / 1113, all passed
- `actionlint` **with shellcheck 0.10.0** — exit 0
- The four packages INFRA-040 edited — 1159 tests, all passed

**ACTIONABLE FINDINGS: 0**

PR #1962, opened 2026-08-21
docs(tasks): the four blocked INFRA items say what blocks them, and one had stale evidence

woojubb:
The four INFRA items an agent cannot complete. That much was already recorded; what was not is **what exactly each waits on**, and whether the record was still true.

Re-measured rather than repeated — and one of them was wrong.

## INFRA-046 — half its evidence was stale

The 2026-07-28 audit says `REGRESSION_RED_PROOF_ENFORCE` is **"set in no workflow"** and concludes **"nothing can fail on it"**.

Both have been false since 2026-08-04. The flag is set in `ci.yml` with the owner decision recorded at the line, and the job is named `regression-red-proof (enforcing: accidental-green only)`. That gate *can* fail.

Corrected in place with the correction **visible**, not by rewriting the original — the audit's reasoning is still worth reading, and so is the fact that it aged.

**What still holds, verified:** it is not registered in `run-all-scans`. That is *correct rather than a gap*: it reads `PR_BODY` and a merge-base diff, so it has no hermetic tree to judge. Registering it would put a check that cannot reach a verdict into a suite whose passes are read as verdicts — which is exactly why `promotion-closes` is excluded, and says so.

**Remaining action:** one `gh api -X PUT` against `protect-develop` (id `18715844`; nine required contexts today, neither of these among them). Needs repo-admin scope. And it should not run yet, on the item's own reasoning: `accidental-green` has never fired on a real PR, so requiring it would put an untested refusal in the merge path. **Blocked on evidence first, a credential second.**

## INFRA-097 — one candidate action re-examined and declined

`scan-workflow-provenance` *does* accept `--base-ref`, and wiring it into CI is the only step short of trusted provenance an agent could take. So it was examined rather than left unmentioned.

Declined, and for the item's own reason: it improves **detection**, and it does not make the control plane trusted — the guard would live in the same file the PR under test can edit. Landing it would close no part of the objective while making the item *look* closer to done.

## INFRA-104 — measured again, not assumed

`develop` is 70 commits ahead of `main`, and every closing reference it carries names an issue that is already closed — checked against live issue state. A promotion today derives an **empty block for the second time**.

Manufacturing a candidate by reopening one of the issues this session hand-closed was offered and declined: satisfying a checkbox by arranging the evidence is the shape this repository refuses everywhere else.

## INFRA-054 — the block restated as three decisions

None technical; the load-bearing one is whether losing the promotion PR (and CodeQL's annotations on it) is acceptable. Once settled the implementation is ordinary and an agent can do it — its vehicle is `workflow_dispatch`, so unlike INFRA-042 and INFRA-065 it is **not** blocked by the 2026-08-04 cron directive.

## Verification

`pnpm harness:scan` — 129 passed, 3 skipped.

### DIFF
--- a/.agents/tasks/INFRA-046-promote-advisory-gates.md
+++ b/.agents/tasks/INFRA-046-promote-advisory-gates.md
@@ -133,11 +133,27 @@ A four-way recurrence audit measured what these two floors are currently worth:
   `ci.yml`'s `regression-red-proof (advisory)`, exits 0 on failure, and its enforcing branch is
   gated on `REGRESSION_RED_PROOF_ENFORCE` — **set in no workflow**. Same shape for
   `check-patch-coverage`.
+
+  > **Half of this is STALE as of 2026-08-04, re-measured 2026-08-22.** The flag IS set:
+  > `.github/workflows/ci.yml` carries `REGRESSION_RED_PROOF_ENFORCE: '1'` with the owner decision
+  > recorded beside it, and the job is now named `regression-red-proof (enforcing: accidental-green
+only)`. So it CAN fail, and "nothing can fail on it" below is no longer true of this gate.
+  >
+  > What still holds: it is not registered in `run-all-scans` — verified by grep, no entry — and
+  > that is correct rather than a gap. It reads `PR_BODY` and a merge-base diff, so it has no
+  > hermetic tree to judge; registering it would put a check that cannot reach a verdict into a
+  > suite whose passes are read as verdicts. `promotion-closes` is excluded from `harness:scan` for
+  > the same reason and says so.
+  >
+  > `check-patch-coverage` is unchanged: still advisory, still `PATCH_COVERAGE_ENFORCE`-gated, and
+  > excluded from the required list on the ground that a required context must be able to fail.
+
 - The defect it exists to catch recurred **twice in one session** (ARCH-004 RUNTIME-14, CORE-026
   RUNTIME-12) and is `common-mistakes` #82.
 
 So the repository has a built, tested floor for its accidental-green class and **nothing can fail on
-it**. The change is two environment variables and moving two contexts to required — the smallest
+it**. (Half-corrected above: the flag has since been set, so `regression-red-proof` CAN fail; only
+the required-list half remains.) The change is two environment variables and moving two contexts to required — the smallest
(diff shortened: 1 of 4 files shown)


woojubb:
## Pre-merge review

**base:** `024ca7128dda01e5470b14eb27aeaa3bc65a1995`
**head:** `286c059f2c5d997c8211fdd6d46708c531913688`

CI green, no review threads.

### The point of this change is that re-checking a record found it wrong

These four items were already marked as not-completable. Re-measuring them anyway turned up **stale evidence in INFRA-046**: its audit says `REGRESSION_RED_PROOF_ENFORCE` is "set in no workflow" and concludes "nothing can fail on it". Both stopped being true on 2026-08-04 — the flag is set, with the owner decision recorded at the line, and the job is now `regression-red-proof (enforcing: accidental-green only)`.

Left as a visible correction rather than a rewrite: the audit's reasoning is still worth reading, and so is the fact that it aged.

That is the same failure this whole INFRA sweep kept hitting — INFRA-040's "flood" was six, INFRA-109's table was four rows out of date, INFRA-111 listed two commands bash itself rejects. **A recorded number is a claim about the day it was taken.**

### One thing I checked rather than asserted

INFRA-097's scan *does* accept `--base-ref`, so wiring it into CI is a real option an agent could take. I examined it and declined: it improves detection and does **not** make the control plane trusted, because the guard would sit in the file the PR under test can edit. Taking it would have moved the item's appearance without moving its objective.

### And one I refused

INFRA-104's TC-10 could be satisfied by reopening an issue this session hand-closed and letting a promotion close it. Offered, declined — arranging the evidence to satisfy a checkbox is what this repository refuses everywhere else, including in the items above.

### What each of the four now waits on

| item | waits on |
|---|---|
| INFRA-046 | one `gh api -X PUT` on `protect-develop` — and evidence first: `accidental-green` has never fired on a real PR |
| INFRA-054 | three policy decisions, the load-bearing one being whether losing the promotion PR is acceptable |
| INFRA-097 | org config, a GitHub App, or a trigger-policy change — all outside this repository |
| INFRA-104 | an observation with zero candidates today (re-verified against live issue state) |

`blocked` and `in-progress` are the accurate statuses. Marking them `done` is what `unearned-done-claims` exists to refuse.

### Verification

`pnpm harness:scan` — 129 passed, 3 skipped.

**ACTIONABLE FINDINGS: 0**

PR #1964, opened 2026-08-21
feat(lint): the warning count is a ratchet, and the number has one owner (INFRA-039)

woojubb:
INFRA-039 is a **draft spec with no task file**, so it was in neither the backlog nor the issues — this sweep nearly missed it. Found on a final pass, and re-measuring was worth the pass.

## The document's own number was stale, in the wrong direction

`pnpm lint` today: **0 errors, 2093 warnings** across 1861 files. The spec recorded **1798** on 2026-07-16. **The debt grew by 295 while the document sat in draft.**

| rule | count |
|---|---|
| `@typescript-eslint/ban-types` | 842 |
| `no-magic-numbers` | 624 |
| `max-lines-per-function` | 340 |
| `complexity` | 128 |
| everything else | 159 |

That growth is the argument for taking the **signal** half first. The spec's stated problem is not the volume — the two-tier policy is deliberate — it is that *"a genuinely-new warning in a PR is invisible in a 1798-line haystack."* A ceiling fixes that without touching a line of source. A reduction pass without a ceiling fixes it until next month.

## The ratchet is eslint's own mechanism

`--max-warnings 2093` on the root `lint` script. No second implementation.

That script is part of `harness:verify:release` — asserted by `check-release-governance.mjs` — and `release-grade verification` is a **required** context on every PR to `main`. So the ceiling is enforced by the tool, on the promotion path. Two packages already carry `--max-warnings 0`, so the mechanism has precedent here.

`scan-lint-warning-ratchet` keeps the **number** honest rather than re-measuring it: the flag must be *present* (without it the script gates nothing and the ratchet is decorative), the ceiling must *match* its baseline so it has one owner, and the baseline may fall and never rise without a deliberate re-freeze.

It deliberately does **not** run eslint — a full workspace lint is minutes, and `harness:scan` runs on every pre-push.

## The harness refused two drafts, both correctly

- The scan first declared a measured size of `1`. **A number that cannot vary is not a measurement**, and the provenance floor would then have wanted a counter test pinning a constant. The offered alternative was its pending-debt list — adoption of debt rather than avoidance of it. The declaration is gone instead.
- Removing the line was not enough: the provenance scan detects the marker in the **source**, so naming it in the explanatory comment was itself enough to be judged as declaring one. Found by doing exactly that.

## Verification

- Red proofs: deleting the flag fails the scan and two repository cases; raising the ceiling to 2200 reports `ROSE`
- 11 unit cases — both flag spellings, absence, missing lint script, a rise, an unfrozen fall, a missing baseline
- `pnpm harness:test` — 229 files / 4374 tests + 73 / 1113, all passed
- `pnpm harness:scan` — 130 passed, 3 skipped
- `pnpm lint` — exit 0 at the ceiling

The **reduction** stays open and is what the spec is for; `no-magic-numbers` (624) and `max-lines-per-function` (340) are the two mechanical classes, and every one removed is a ratchet step down. The document stays `draft` because it has not been through the gates.


woojubb:
## Pre-merge review

**base:** `50cbbcf88790bd3ae762af8894f917acef039e98`
**head:** `f5cca8931f0022256468de283831ac91705c2aaf`

CI green, no review threads.

### This item was nearly missed, and its own number was stale

INFRA-039 is a draft spec with **no task file**, so it appeared in neither the backlog nor the issues — three earlier sweeps of this session walked past it. Re-measuring on the fourth: **2093** warnings against the **1798** the document recorded on 2026-07-16. The debt grew by 295 while the document sat in draft.

That is the fourth INFRA item this session whose recorded figure did not survive re-measurement, and the pattern is now the most useful thing the sweep produced: *a recorded number is a claim about the day it was taken.*

### Signal before reduction, and the growth is why

The spec's stated problem is not the volume — the two-tier policy is deliberate — it is that a new warning is invisible inside it. A ceiling restores the signal without touching 964 lines; a reduction pass without a ceiling restores it until next month. The growth from 1798 → 2093 is that argument, measured.

The ratchet is `--max-warnings` on the root `lint` script — eslint's own mechanism, on the release path where `check-release-governance.mjs` already asserts that script runs, and `release-grade verification` is required on every PR to `main`. The scan governs the **number**, not the count, so the ceiling does not become a hand-maintained second source.

### Two drafts the harness refused, both correctly

- The scan first declared a measured size of `1`. **A number that cannot vary is not a measurement** — the provenance floor would then have wanted a counter test pinning a constant, and the offered alternative was its pending-debt list, which is adoption of debt rather than avoidance.
- Deleting the line was not enough: that scan reads the marker from the **source**, so naming it in the explanatory comment was itself enough to count as declaring one. Found by doing exactly that.

### Verification

- Red proofs: deleting the flag fails the scan and two repository cases; raising the ceiling to 2200 reports `ROSE`
- 11 unit cases — both flag spellings, absence, missing lint script, a rise, an unfrozen fall, a missing baseline
- `pnpm harness:test` — 229 files / 4374 tests + 73 / 1113, all passed
- `pnpm harness:scan` — 130 passed, 3 skipped
- `pnpm lint` — exit 0 at the ceiling

**ACTIONABLE FINDINGS: 0**

PR #1971, opened 2026-08-21
docs(rules): a recorded constraint is a claim about the day it was written

woojubb:
Catalogue entry 90, from the pattern that dominated this INFRA backlog sweep — and that the sweep itself then committed.

## The pattern

A task, spec or comment states why something cannot be done — a count too large, a flag unset, a credential absent, a check that would refuse — and the next agent carries that sentence forward as the reason to defer, narrow or refuse.

It was true when written. It is a claim about **that day**.

Measured, all within one sweep:

| the record said | measured |
|---|---|
| "type-aware finding **flood**" | **6** violations |
| "**1798** warnings" | **2093** |
| "`REGRESSION_RED_PROOF_ENFORCE` set in **no workflow**" | set two weeks earlier |
| a table of **six** hook/scan divergences | **four** already fixed |
| two spellings listed as guard **holes** | commands bash itself rejects |
| "the only closing condition" | a case that **cannot fail** |
| "this agent has **no repo-admin scope**" | `permissions.admin: true` |

Implementing from the record would have fixed problems that no longer existed, and refused work that was permitted.

## The last row is the sharpest one

It was written **this session**, asserted three times without one API call, as the reason *not* to do something. That is where the pattern costs most: nothing downstream tests a constraint cited for inaction. A stale figure in a Problem section gets caught the moment someone implements against it; a stale constraint in a Direction section just quietly keeps the work from happening.

## The fix

Run the constraint before acting on it — the count, the grep, the permission read, the completion condition — and record the new number **beside** the old rather than replacing it, so the next reader sees the drift as well as the value.

## `Mechanism: none` is a decision here

A recorded constraint is prose about an arbitrary subject; no check can re-derive it in general. The nearest partial floors — `unearned-done-claims` and `measurement-provenance` — read done-records and self-reported sizes, not a constraint stated in a Problem or Direction section.

`mistake-mechanisms` refused the first draft for exactly this: it requires "no mechanism" to be an admission in the declared form, not the state that happens when nobody wrote one.

## Verification

`pnpm harness:scan` — 130 passed, 3 skipped; the entry reads as accepted-with-a-reason (39 name a mechanism, 51 are accepted debt).


woojubb:
## Pre-merge review

**base:** `01ecfb1bbea92e7a7ce9a3ec81dd2af9fd5c29d2`
**head:** `57370f28623f3ed19c86927378d8e999dd65d2d7`

CI green, no review threads.

### The entry earns its place by having been committed here

Seven stale claims in one sweep, and the last one is mine — "this agent has no repo-admin scope", asserted three times across three re-examinations of the same file, without one API call. `permissions.admin` is `true`.

That is why the entry singles out constraints cited for **inaction**. A stale figure in a Problem section gets caught the moment someone implements against it. A stale constraint in a Direction section just quietly keeps the work from happening, and nothing downstream ever tests it.

### `Mechanism: none` is a decision, and the scan made me say so

`mistake-mechanisms` refused the first draft: it requires "no mechanism" to be an admission in the declared form, not the state that happens when nobody wrote one. Correct refusal — and a small instance of the same principle, since an unstated absence reads as an oversight rather than a judgement.

The reason recorded is that a constraint is prose about an arbitrary subject; `unearned-done-claims` and `measurement-provenance` are the nearest floors and read done-records and self-reported sizes, not a Problem or Direction section.

### Verification

`pnpm harness:scan` — 130 passed, 3 skipped. 90 catalogue entries: 39 name a mechanism, 51 accepted-as-recurring with a recorded reason.

**ACTIONABLE FINDINGS: 0**

PR #1979, opened 2026-08-22
docs(tasks): record INFRA-104's first live TC-10 subject

### LINKED ISSUE #1965: A task can be marked done with its acceptance criteria unticked, and every scan passes
## Measured

Found by probing, not by reading: I marked four `blocked` / `in-progress` tasks `status: done`, added a `completed:` date, and moved them to `completed/` — exactly what "completing" them would do — then ran the full suite.

```
INFRA-046  status=done  unticked=0  ticked=0
INFRA-054  status=done  unticked=3  ticked=0
INFRA-097  status=done  unticked=5  ticked=0
INFRA-104  status=done  unticked=1  ticked=9
```

`unearned-done-claims` **passed.** So did `backlog-placement` and `task-archival`. The only failures were `reference-kind-qualified`, `rule-case-narrative` and `resolving-claims` — and those fired because the *files moved and broke inbound links*, not because anything judged the completion claim.

Nine of those unticked boxes are acceptance criteria the items wrote for themselves. INFRA-097's five include "Add adversarial tests proving a PR cannot replace its own required gate"; INFRA-104's one is its only user-execution scenario.

## Why the existing guard does not catch it

`unearned-done-claims` judges what a done record **says**:

- **U2** — an evidence/proof heading whose body cites nothing
- **U3** — a named section reference that resolves to no heading
- **U4** — a **ticked** box asserting "verified by …" with no citation after the claim

Every rule is about content that is *present*. None is about a criterion that is present and **unmet**. A task with zero prose and five unticked boxes is, to this scan, a clean done record — and it is the shape that costs the most, because `status: done` is what the next reader trusts instead of opening the file.

`task-archival` is the nearest thing: it fails "a fully-checked task file whose spec never reached `spec-docs/done/`". It keys on *fully-checked*, so a **partly**-checked one falls through the same gap from the other side.

## Suggested shape

A done-status task should have no `- [ ]` under its acceptance/completion criteria, or should name each survivor with a reason on the line — the same escape `named-artifact-resolves` uses (`allow-missing-artifact: <reason>`), so a deliberate exception is a decision someone wrote down.

Two things worth checking before implementing, because either could make a naive rule fire on correct work:

1. **How many existing `completed/` items would go red?** There are 79 completed INFRA files alone. If the count is large this needs the burn-down baseline shape (`may fall, never rise`) rather than a hard floor, like `reference-kind-qualified` and `named-artifact-resolves`.
2. **Not every `- [ ]` is an acceptance criterion.** Some task files use checkboxes in a plan or a rejected-alternatives list. The rule should be scoped to the criteria section, or it will refuse records that are legitimately complete.

## Provenance

Found during the INFRA backlog sweep, while verifying a *refusal* rather than a change: I had said these four could not be marked done "because `unearned-done-claims` exists to refuse it". Running it showed that claim was wrong about the mechanism — the record would have been false, and nothing mechanical would have said so. The substantive reasons remain measured and unchanged (neither gate is in `protect-develop`'s required list; 2 of 2 guarded workflows still load their definition from the pull request; the last promotion body carried 0 closing keywords), so the four stay `blocked` / `in-progress`.

The gap this issue reports is the more useful half of that correction.

woojubb:
## What

Appends one dated entry to the INFRA-104 task record. Docs-only — no code, no scan, no gate.

## Why now

Every earlier entry on this item recorded the same blocker: each closing reference `develop` carried
named an issue that was **already closed**, so a promotion derived an **empty** block and TC-10 — the
half that asks whether *GitHub*, not a person, closes the issue — had nothing to act on. That
happened four separate times this session.

## What was measured

Each of the 16 issues referenced by a closing keyword on `develop`-not-`main` was read individually
through the API:

```
closed:  15
OPEN:     1   -> issue #1965
```

And the generator was run rather than reasoned about:

```
$ node scripts/harness/promotion-closes.mjs

Closes #1965
::examined:: 130 pull-request body(ies) read, 37 issue record(s) checked
```

Issue #1965's `Closes` keyword was ignored by GitHub because the PR did not target the default
branch — **the exact defect this item exists to compensate for**, and the reason a subject survived.

## The near-miss, recorded because it nearly went the other way

The first sweep reported "no open candidates" — while **four of the sixteen reads had timed out**. A
timeout is not an answer of "closed". Re-run distinguishing an error from a state, `#1965 open`
appeared. `merge-gate` enforces the same principle a layer up; this was the same trap inside a
survey.

## Notes

- The three checks for the next promotion are written **before** the evidence exists, so the reading
  cannot be fitted to it.
- `reference-kind-qualified` caught the two new bare references (frozen 5 → 7); both were qualified
  and the ratchet is back at 5. Note its qualifier must sit on the **same line** as the reference —
  a line break between them still reads as unqualified.

ACTIONABLE FINDINGS: 0

Refs INFRA-104

### DIFF
--- a/.agents/tasks/INFRA-104-promotion-carries-closing-keywords-to-main.md
+++ b/.agents/tasks/INFRA-104-promotion-carries-closing-keywords-to-main.md
@@ -156,3 +156,42 @@ The substantive grounds are unchanged, and were re-measured rather than restated
 
 The gap the probe exposed — a `done` task with unticked acceptance criteria passes every scan — is
 filed as issue #1965 rather than folded in here.
+
+### 2026-08-22 — TC-10 has a live subject for the first time
+
+Every earlier entry records the same blocker: every closing reference `develop` carried named an
+issue that was ALREADY CLOSED, so a promotion derived an empty block and the half TC-10 asks about —
+GitHub, not a person, closing an issue — had nothing to act on.
+
+Re-measured today, reading each referenced issue through the API rather than trusting the record:
+
+```
+referenced by a closing keyword on develop-not-main:  16
+  closed:  15
+  OPEN:     1   -> issue #1965
+```
+
+**Issue #1965 is open, and `Closes #1965` is on `develop`.** GitHub ignored that keyword because the
+pull request did not target the default branch — which is the exact defect this item exists to
+compensate for — so the issue survived to become the first live subject the promotion block can name.
+
+It arrived from ordinary work. Issue #1965 was filed after probing whether four blocked items could be
+marked `done` (every existing scan passed, which was the defect); another session implemented the
+guard and merged it with the closing keyword. Nothing was arranged to produce this.
+
+**One measurement note, because it nearly went the other way.** The first sweep reported "no open
+candidates" — but four of the sixteen reads had TIMED OUT, and a timeout is not an answer of
+"closed". Re-run distinguishing an error from a state, `#1965 open` appeared. `merge-gate` refused a
+merge on the same principle an hour earlier ("no answer is a refusal — it is NOT an answer of
+'none'"), and this is the same trap one layer down, in a survey rather than a gate.
+
+**What remains for TC-10 is one event: the next promotion.** What to check when it happens, written
+BEFORE the evidence exists so the reading cannot be fitted to it:
+
+1. the promotion body carries a non-empty block naming issue #1965;
+2. the `promotion closes` required context passes on the promotion pull request;
+3. issue #1965 is closed by GITHUB — the close event attributed to the merge, not to a person.
+
+Any of the three failing is a finding about this item, not about the promotion. In particular,
+if issue #1965 is closed by hand before then, the subject is consumed and TC-10 waits again — which is what
+happened four times earlier in this session and is recorded above.
--- a/.agents/tasks/INFRA-104-promotion-carries-closing-keywords-to-main.md
+++ b/.agents/tasks/INFRA-104-promotion-carries-closing-keywords-to-main.md
@@ -189,7 +189,10 @@ merge on the same principle an hour earlier ("no answer is a refusal — it is N
 BEFORE the evidence exists so the reading cannot be fitted to it:
 
 1. the promotion body carries a non-empty block naming issue #1965;
-2. the `promotion closes` required context passes on the promotion pull request;
+2. the `promotion closes` job reports green on the promotion pull request. It is ADVISORY, not
+   required: `.github/required-status-checks.json` declares it required on `main` and the live
+   `protect-main` ruleset does not, filed as issue #1980. It still runs, and a failure would show as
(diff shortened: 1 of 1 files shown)


woojubb:
## Self-review round 2 — one finding, raised on my own diff and resolved in 67d99e158

The entry asserted:

> 2. the `promotion closes` **required context** passes on the promotion pull request

I checked that instead of leaving it, and it is **not true**.

```
  - promotion closes: .github/required-status-checks.json declares it required on `main`,
    but the LIVE ruleset does not require it — it is enforcing nothing.
```

`.github/required-status-checks.json` declares four contexts under `branches.main`; the live
`protect-main` ruleset requires three. `promotion closes` — the gate written *for this very task* —
is the missing one.

Nothing caught the drift because the reconciling half (`--live`) runs from
`.github/workflows/ruleset-drift.yml`, which is `workflow_dispatch:` only since its schedule was
removed. A recorded constraint nobody re-derives is not a constraint — the same defect class this
task record has been cataloguing all along, this time landing on the task's own gate.

**Filed as issue #1980, not folded in.** This PR only corrects the wording: the job runs and
reports, a failure surfaces as `UNSTABLE`, and `.claude/hooks/merge-gate.sh` refuses that — so it
gates this repository's merge path but not GitHub's.

Note for reviewers of `reference-kind-qualified`: its qualifier must sit on the **same line** as the
reference. A line break between `issue` and `#1965` still reads as unqualified, and prettier reflows
the file after staging — so the ratchet is worth re-running post-commit, which I did (PASS).

ACTIONABLE FINDINGS: 0

woojubb:
## Reviewer verdict

**Base** `ad9d4a557c9c2941ddfb38919a9d902dd5cffd07` (origin/develop)
**Head** `67d99e1584e2d61697cef6feaac40ec9a7927671` (origin/docs/infra-104-tc10-candidate)

Docs-only: one dated entry appended to `.agents/tasks/INFRA-104-promotion-carries-closing-keywords-to-main.md`. No code, no scan, no workflow, no gate.

Every factual claim in the entry was executed rather than restated:

- 16 issues referenced by a closing keyword on `develop`-not-`main`, read individually through the API — 15 closed, issue #1965 open.
- `node scripts/harness/promotion-closes.mjs` run: derives exactly `Closes #1965` over 130 PR bodies and 37 issue records.
- `node scripts/harness/promote.mjs --dry-run`: clean, promotes develop's tree unchanged.

One finding was raised against this diff and resolved in `67d99e158`: check 2 asserted `promotion closes` is a **required** context. `scan-main-required-checks.mjs --live` says otherwise, and the ruleset history shows `protect-main` unmodified since 2026-07-26 — so it was never added, not removed. Filed as issue #1980 with the measurement; the entry now states what is true.

CI: 15 passed, 0 failed, `mergeStateStatus=CLEAN`. Review threads: 0 total, 0 unresolved.

The entry's three observation conditions are written before the evidence exists, which is the property that makes them worth anything.

ACTIONABLE FINDINGS: 0

PR #2010, opened 2026-08-22
fix(lint): the warning count is back under its ceiling, and the ceiling is now true

woojubb:
## Why

`pnpm harness:verify:release` — the only CI path that runs the root `lint` script — reports

```
✖ 2115 problems (0 errors, 2115 warnings)     ceiling: --max-warnings 2093
```

`release-grade verification` is a **required** context on `protect-main`, so the next promotion would
have gone red on it. Found by running the gate locally before opening a promotion PR.

## What the measurement actually said, after two wrong readings

**Reading 1 — wrong.** I attributed the rise by grepping `git show --stat` output for changed paths,
concluded the only code commit since the baseline contributed 2 warnings, and looked elsewhere.
`--stat` **abbreviates long paths**, so two of that commit's six files were never in the text being
matched. A presentation layer was asked a question only the data layer can answer.

**Reading 2 — per-file diff of two `-f json` reports.** All 21 of the rise came from that one commit:

```
+7  packages/agent-transport-webrtc/src/pairing-gate.ts
+7  packages/agent-transport-webrtc/src/handoff-grant-gate.ts
+2  packages/agent-remote-pairing/src/revocation-list.ts
+2  packages/agent-remote-pairing/src/root-rotation.ts
+2  packages/agent-transport-webrtc/src/admission-steps.ts
+1  packages/agent-transport-webrtc/src/pairing-gate-options.ts
```

**And the baseline was never true.** Measured at `f67006212`, the commit that recorded it:

```
WARNINGS AT THE BASELINE COMMIT f67006212: 2094
```

Recorded ceiling: 2093. It was one low the day it was written. I measured this **before** touching
the file, because "correct a mis-record" and "weaken a gate to unblock myself" produce an identical
diff, and without the number I'd have picked whichever story suited me.

## What is fixed, and what is deliberately not

**Not fixed — 11 of the 21.** They are `unknown` in type-guard signatures:

```ts
function isHandoffGrantFrame(value: unknown): value is IHandoffGrantFrame
readonly verify: (grant: unknown) => Promise<IPeerAdmission>;
```

That is the construct the rule's **own message** recommends (*"type guards at boundaries"*).
Rewriting it to satisfy a counter would make correct code worse.

**Fixed — unambiguous only:** 19 dead import bindings removed, and the base64 quantum in
`fromBase64Url` given a name instead of two bare `4`s.

**Ceiling and baseline drop 2093 → 2092**, the measured value of the tree. A ratchet may fall.

## Verification

- `pnpm run lint`: **2092 problems, 0 errors** — re-run on the **committed** tree, because
  `eslint --fix` touches staged files and the tree I verified is not automatically the tree I
  committed. Confirmed by reading `git show HEAD:package.json`, not `git diff`.
- Four edited packages typecheck.
- Their suites: **155 files, 1597 tests, all passing.**
- `file-size` refused the shrink until the gain was locked; baseline regenerated in the same change
  (80 burn-down entries), as the scan instructs.
- Whole-workspace pre-push: **132 of 133 scans**, the one failure being that `file-size` refusal,
  now green.

Typecheck caught one wrong deletion mid-fix: the warning was on `context-command.ts` and I removed
the same-named import from `context-breakdown.ts`, where it is used. Restored, correct one removed.

## Separate defect, filed not folded in

Issue #1984 — this ratchet's **only** CI execution path is `release-grade verification`, which runs
on `main` PRs only. Its first execution in its entire life was this pre-flight. A guard nothing
routinely triggers cannot fire, the same family as one whose input cannot vary.

ACTIONABLE FINDINGS: 0

Refs INFRA-039

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "test:coverage:packages": "pnpm --filter \"./packages/**\" run --if-present test:coverage",
     "test:coverage:apps": "pnpm --filter \"./apps/**\" run --if-present test:coverage",
     "typecheck": "pnpm run -r --workspace-concurrency=-1 --if-present typecheck",
-    "lint": "eslint packages apps --ext .ts,.tsx --cache --max-warnings 2093",
+    "lint": "eslint packages apps --ext .ts,.tsx --cache --max-warnings 2092",
     "lint:fix": "eslint packages apps --ext .ts,.tsx --cache --fix && prettier --write .",
     "lint:fix:staged": "scripts/harness/with-repo-lock.sh pnpm exec lint-staged",
     "dev": "pnpm run -r --if-present dev",
--- a/packages/agent-command/src/context/context-breakdown.ts
+++ b/packages/agent-command/src/context/context-breakdown.ts
@@ -8,17 +8,10 @@
 
 import { CONTEXT_ESTIMATE_CHARS_PER_TOKEN } from '@robota-sdk/agent-core';
 import {
-  addCommandContextReference,
-  clearCommandContextReferences,
-  DEFAULT_AUTO_COMPACT_THRESHOLD,
   listCommandContextReferences,
   readAutoCompactThreshold,
   readAutoCompactThresholdSource,
   readCommandContextState,
-  removeCommandContextReference,
-  resetAutoCompactThresholdSetting,
-  setCommandAutoCompactThreshold,
-  writeAutoCompactThresholdSetting,
 } from '@robota-sdk/agent-framework';
 
 import { formatAutoCompactLine } from './context-command.js';
--- a/packages/agent-command/src/context/context-command.ts
+++ b/packages/agent-command/src/context/context-command.ts
@@ -1,4 +1,3 @@
-import { CONTEXT_ESTIMATE_CHARS_PER_TOKEN } from '@robota-sdk/agent-core';
 import {
   addCommandContextReference,
   clearCommandContextReferences,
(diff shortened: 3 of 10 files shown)


woojubb:
## Reviewer verdict

**Base** `54e037490ce08e8d9153367d89bfe26cc7cc2d70` (origin/develop)
**Head** `eef3eec14fdcbb85162571d65df9af77d7eda7b4` (origin/fix/lint-ceiling-regression)

CI: **20 passed, 0 failed**, `mergeStateStatus=CLEAN`. Review threads: 0 total, 0 unresolved.

The diff is 7 source files (dead import bindings removed, one constant named) plus two ratchet
numbers moving **down**. No behaviour changes; the four edited packages' suites are 155 files and
1597 tests, all green, and each package typechecks.

Two things I checked specifically because they are where this kind of change goes wrong:

- **The verified tree is the committed tree.** `eslint --fix` runs on staged files at commit time, so
  the tree that passed is not automatically the tree that landed. Re-ran `pnpm run lint` after the
  commit (2092, 0 errors) and read the values out of `git show HEAD:package.json` and
  `git show HEAD:scripts/harness/lint-warning-baseline.json` rather than out of `git diff` — a diff
  proves the tree and HEAD agree, not that either is right.
- **The `unknown` warnings are deliberately not fixed.** They are type-guard signatures at
  boundaries, which the rule's own message recommends. Lowering a counter by degrading correct code
  would be the wrong direction, and the ceiling moving to 2092 records the tree as it actually is.

The baseline archaeology is in the description: at `f67006212` the true count was 2094 against a
recorded 2093, so the file never described the tree. I measured that before editing it, because the
correction and the self-serving weakening produce the same diff.

ACTIONABLE FINDINGS: 0

Issue #1980, opened 2026-08-22
`promotion closes` is declared required on main and is not required — the reconciler that would catch it is dispatch-only


woojubb:
## The drift

`.github/required-status-checks.json` is the **source of truth** for what `protect-main` must
require. It declares four contexts under `branches.main`. The live ruleset requires three.

```
declared (required-status-checks.json)   live (ruleset 18715845)
  promotion ancestry                       promotion ancestry
  main PR source guard                     main PR source guard
  promotion closes                         — MISSING —
  release-grade verification               release-grade verification
```

The repo's own instrument says so:

```
$ node scripts/harness/scan-main-required-checks.mjs --live
main-required-checks scan failed (INFRA-055):
  - promotion closes: .github/required-status-checks.json declares it required on `main`,
    but the LIVE ruleset does not require it — it is enforcing nothing.
```

## Why nothing caught it

The offline half of the scan runs in `pnpm harness:scan` and proves every **declared** context can
fail. It cannot see the ruleset move underneath it — that is `--live`'s job, and `--live` runs from
`.github/workflows/ruleset-drift.yml`, which is now **`workflow_dispatch:` only**. Its `schedule:`
trigger was removed by the standing "turn off all crons" directive, and nothing replaced the
reconciliation it performed.

So the declaration and reality diverged, and the only thing that re-derives them is a button nobody
presses. A recorded constraint nobody re-derives is not a constraint.

## Why it matters now, specifically

`promotion closes` is the gate written for INFRA-104: a promotion body must carry a closing keyword
for every OPEN issue the PRs it promotes declared they close. GitHub honours closing keywords only
on default-branch PRs, so the promotion is the only PR that can close them.

Until today the gate had nothing to act on — every referenced issue was already closed, so the
derived block was empty. That is no longer true. `promotion-closes.mjs` now derives exactly:

```
## Issues this promotion closes

Closes #1965
```

So the **first promotion where this gate has real work to do** is the first one that can merge
without it.

## Partial mitigation, stated so the severity is not overread

The job still RUNS on a `main` PR and reports. GitHub reports `UNSTABLE` when a non-required check
fails, and `merge-gate.sh` refuses any non-`CLEAN` `mergeStateStatus`. So a failing `promotion
closes` would still stop the repo's own merge path — it just would not stop GitHub's.

## Fix, either direction — both halves in one change

Per the scan's own instruction: add `promotion closes` to the live `protect-main` ruleset, **or**
[...]

PR #2013, opened 2026-08-22

### LINKED ISSUE #1965: A task can be marked done with its acceptance criteria unticked, and every scan passes

woojubb:
Built with `node scripts/harness/promote.mjs`, not by hand.

`origin/develop b5fe541d3` → `origin/main 01a245c35`, **89 commits**, promoting develop's tree
unchanged. A1/A2/A3 hold; non-merge debt on `main`: 0.

## Issues this promotion closes: none, and that is a measured result

The derivation read **137 carried pull-request bodies and checked 38 referenced issues**. Every one
is already closed, so the block is legitimately empty — not skipped, and not an empty block standing
in for a derivation that could not complete.

**This one is worth recording, because a subject existed this morning and did not survive.**
Issue #1965 was open with `Closes #1965` on `develop` — the first live subject INFRA-104's TC-10 has
ever had, since GitHub ignores closing keywords on a non-default-branch PR and the promotion is the
only PR that can act on them. It was closed **by hand** at `2026-08-22T01:42:25Z`, `commit=none`,
seventeen minutes after the PR carrying it merged.

That was written down as a named failure branch **before** the evidence existed, in the INFRA-104
record:

> if issue #1965 is closed by hand before then, the subject is consumed and TC-10 waits again

Fifth occurrence this session. Not carelessness by anyone — the workflow leaves the issue open and
visible for hours after its fix merges, which reads as an oversight to anyone tidying. The finding
is that **TC-10's subject is something normal operation destroys**, so as specified it may be
unobservable rather than merely unobserved. To be recorded on the item after this lands.

## Pre-flight

`release-grade verification` is the required context that executes this repository's code, so it was
run locally first:

```
$ pnpm harness:verify:release > log 2>&1; echo "REAL EXIT: $?"
REAL EXIT: 0
```

Captured with a redirect and `$?`. An earlier run of the same command was piped through `tail`, and
its reported "exit 0" belonged to `tail` — the verification's own status was never read. Worth
stating because that is the shape of a green that means nothing.

Lint: **2092 warnings, 0 errors** against a ceiling of 2092. That ceiling was 2093 and failing at
2115 when this promotion was first attempted; PR #2010 fixed the tree and corrected the number,
which is the only reason this PR can be green. See issue #1984 — the ratchet's sole CI execution
path is this very job, so this promotion was its first execution ever.

## Merge method

`--merge` only. `protect-main` sets `allowed_merge_methods: ["merge"]`, and a squash would destroy
the ancestry link `promotion ancestry` exists to preserve.

Do **not** use GitHub's "Update branch" button — both modes break what the promotion asserts. Re-run
`promote.mjs` instead.

## Known gap, filed not folded in

`promotion closes` is declared required on `main` in `.github/required-status-checks.json` and the
live `protect-main` ruleset does not require it — issue #1980. The ruleset has not been modified
since 2026-07-26, so it was never added rather than removed. The job still runs and reports here.


woojubb:
## Reviewer verdict

**Base** `01a245c35` (`main`) · **Head** `fd1bee47be611546e7c0e541cfa6ab0d4e545f4a`

**All four required contexts pass:**

```
promotion ancestry            pass   9s
main PR source guard          pass   2s
promotion closes              pass   40s
release-grade verification    pass   14m43s
```

`mergeStateStatus` is `UNSTABLE`, caused by two **non-required** checks: `CodeQL` and
`Dependency review`. Both were verified by hand rather than waved through, because a hook cannot
reach the reason they are red here.

### Why both are artefacts of the base, not defects in what is being promoted

The decisive fact is structural, and I checked it directly rather than trusting A3's verdict:

```
$ git diff --stat b5fe541d3 origin/release/promote-develop-to-main
(no output — byte-identical trees)

$ git merge-base --is-ancestor origin/main origin/release/promote-develop-to-main
yes
```

**The promoted tree is byte-identical to the `develop` commit it promotes.** A promotion therefore
cannot introduce a single line that `develop` does not already have, and every one of those lines
passed the full required set on its own `develop` PR. Anything either scanner reports as "new" is
new *relative to `main`*, which is 89 commits behind — not new to the codebase.

**CodeQL** says so itself in its summary: *"Alerts not introduced by this pull request might have
been detected because the code changes were too large."* This is the documented reason CodeQL is
deliberately excluded from `protect-main` in `.github/required-status-checks.json` — its verdict
scales with promotion size rather than with anything the promotion decides.

**Dependency review** fails on one line:

```
examples/capabilities/sandboxed-tools/package.json » @robota-sdk/agent-framework
  – License: AGPL-3.0-only OR LicenseRef-Commercial
```

That is **this repository's own package**, tripping the licence allow-list. The exemption list
`allow-dependencies-licenses` names `@robota-sdk/agent-core` and `@robota-sdk/agent-provider` and no
other first-party package, so any example that imports a third one fails. A hand-maintained list
narrower than its population — filed separately rather than fixed inside a frozen promotion diff.

### Precedent

Promotion #1895 merged with `CodeQL`, `Dependency review` **and** `review-gate` red. This one has
`review-gate` green and the same two red, for the same structural reason.

### Merge method

`gh pr merge 2013 --merge`. Never `--squash` — `protect-main` sets
`allowed_merge_methods: ["merge"]` and a squash destroys the ancestry link `promotion ancestry`
exists to preserve.

ACTIONABLE FINDINGS: 0

Issue #1828, opened 2026-08-17
ARCH-039: ten options on the PUBLISHED InteractiveSession surface are declared and set by nothing


woojubb:
Surfaced by extending `optionReachability` to `IInteractiveSessionStandardOptions` during ARCH-013
stage 3. Filed rather than folded in: all ten are pre-existing and none is caused by that change.

Reproduce: `node scripts/harness/scan-option-reachability.mjs` with the
`IInteractiveSessionStandardOptions` entry removed from
`scripts/harness/option-reachability-baseline.json`.

## Why this watch was added

Stage 1's ratchet watched only `ICreateSessionOptions` — the INTERNAL assembly surface. That cannot
see a key declared on the type a consumer actually uses and forwarded by nothing, and stage 3's first
attempt proved the gap is live: both `guardrails` and `retrievalAdapter` were added to the published
type, projected into `ICreateSessionOptions`, and dropped by the ~40-field hand-map in
`initializeInteractiveSessionAsync` in between. The ratchet certified an 11→9 gain the code did not
deliver, and only an independent review caught it.

`IInteractiveSessionStandardOptions` is where unreachability is user-visible: a consumer can only set
what this type declares.

## The ten

`automaticMemory`, `commandHostAdapters`, `orgPolicy`, `providerDefinitions`, `recallMemory`,
`remoteCommandPolicy`, `sessionName`, `sessionStore`, `shellExec`, `terminalHandoff`

Each is declared on the published surface and assigned by no production code, so a consumer setting it
gets silence. They are frozen in the baseline as a burn-down (may shrink, must never grow), exactly as
stage 1 froze its own eleven.

## What this is NOT

Not a claim that all ten are defects of the same kind. Some may be legitimately injected by a
composition root the scan cannot see, some may be dead declarations to remove, and at least
`sessionStore` also appears in the `ICreateSessionOptions` list — which suggests one root cause across
both surfaces rather than ten independent ones. Triaging that is the work; the measurement is done.

Several are advertised capabilities in the same sense `guardrails` and `retrievalAdapter` were
(SELFHOST-008's `automaticMemory`/`recallMemory`, REMOTE-006's `remoteCommandPolicy`, TERM-001's
`terminalHandoff`), so the same question applies to each: is it unreachable, or reached by a path the
scan's producer rule does not model? Stage 3's answer for its two was "genuinely unreachable", and it
took driving the public entry point to establish that rather than reading the code.

woojubb:
[...]

Issue #1805, opened 2026-08-16
ARCH-037: IAgentJobHostContext and ICommandSessionRuntime are decomposed but their consumers still name the aggregate (18 and 4 sites)


woojubb:
## What

ARCH-029 decomposed all three command-axis contracts into role ports, but drove only
`ICommandHostContext` to zero consumer references. The other two are frozen at what they measure
today:

| aggregate | ports | type-position references outside the allowlist |
| --- | --- | --- |
| `ICommandHostContext` | 15 | **0** (TC-05, the approved criterion) |
| `IAgentJobHostContext` | 5 | **18**, across 8 files |
| `ICommandSessionRuntime` | 6 | **4**, across 5 files |

Both are ratcheted in `.agents/harness.config.json` → `aggregateNaming.aggregates`, so they can fall
and cannot rise. This item is to take them to 0.

## Why it matters

ARCH-029's own thesis is that a god contract is not fixed by decomposing it — it is fixed when the
consumers stop naming it, and those are two different events. By that standard 2 of the 3
decompositions are structural today and behavioural nowhere: a command that declares
`IAgentJobHostContext` still takes all 15 members when it uses two.

This is not a defect in ARCH-029. TC-05 scoped the zero to `ICommandHostContext` deliberately, and
the design sanctions exactly this split ("TC-05a freezes at N, TC-05b reaches 0 in a named seam with
every residual site enumerated and justified"). Filed so the second half has an owner instead of
living as a frozen number nobody revisits.

## Where

`IAgentJobHostContext` (18): `packages/agent-command/src/agent/agent-command.ts`,
`packages/agent-command/src/schedule/schedule-command.ts`, and 6 more — enumerate with:

    node -e "import('./scripts/harness/scan-aggregate-naming.mjs').then(m=>{const{counts}=m.collectAggregateNaming();for(const[k,v]of counts)console.log(k,[...v.files.entries()])})"

## Direction

The same mechanism ARCH-029 used, which is already built and in the repo: narrow each declaration to
the role port(s) the body actually reads, then drop the frozen number to 0. The role ports exist
(`agent-job-roles.ts`, `session-roles.ts`), the ratchet exists, and the conformant doubles
(`createTestAgentJobHost`, `createTestSessionRuntime`) exist. What is left is the declaration
migration.

## Found by

Independent review of the ARCH-029 branch, which flagged that claim (1) presented all three as
decomposed while only one was measured.

woojubb:
[...]

Issue #1619, opened 2026-08-03
HARNESS-074: the review machinery performs a second review instead of resolving the first


woojubb:
Owner policy: code review on an open pull request is delegated to the review automation the PR runs.
The local loop's job is to RESOLVE what that review reports — fetch the findings, judge each, fix or
refute, push, read the round the push re-triggered.

The machinery does something else: it performs a second, local review of the same change and converges
on that. The cost is paid twice, and the second opinion is the weaker one — it does not carry the PR's
comment history, so it cannot see which findings an earlier round already answered.

The procedure for the intended shape already exists and was not being invoked:
`.agents/skills/automated-review-convergence/SKILL.md`.

**Each piece, judged against the policy:**

| Piece | What it does | Verdict |
|---|---|---|
| `pr-review-orchestration` Round B step 1 | Dispatched a reviewer agent on the OPEN PR | Fails — corrected |
| `.claude/hooks/pre-push-check.sh` | Refuses a push without a recorded LOCAL review of the diff | Fails |
| `record-local-review.mjs`, `.agents/local-reviews/` | Stores that local review, per working tree | Follows the above |
| `pr-review-orchestration` Round A | The skill form of the same pre-push local review | Fails |
| `.claude/hooks/merge-gate.sh` | Reads the disposition from the PR LABEL, not the local record | Conforms — precedent |
| `.github/workflows/review-gate.yml` | Reads code-scanning alerts and labels | Conforms |
| `automated-review-convergence` | Fetch, judge, resolve, push, re-read | Is the policy |

Two things sharpen the direction rather than just condemning the machinery:

- **The merge gate already moved.** It once read the per-working-tree record and was changed to read a
  PR label, because a record keyed to a local checkout is invisible to the clone that merges. The same
  reasoning applies to every decision the local record still carries.
- **The pre-push gate's stated reason is a COST argument, not a review argument** — a round before the
  push costs a minute where the same round after it costs a CI cycle. That reason survives. What does
  not survive is discharging it by performing a review: mechanical verification before a push is
  verification; a subjective review before a push is the duplicated reviewer wearing a hook.

Task: `.agents/tasks/HARNESS-074-the-review-loop-duplicates-the-reviewer.md`

Direction is ordered because the record's readers change with each step: fix the pre-push gate's
demand first, make Round A follow the hook, keep Round B resolving (done), then decide the local
record's fate once its readers are enumerated mechanically, and only then move
[...]

PR #2016, opened 2026-08-22
docs(harness): correct the issue-link header's central number, and record the instance that happened while reviewing it

woojubb:
Refs #1916. Follow-up to PR #2011, which merged the corrected **code** and left the uncorrected **header**.

## What was wrong

The module header said 39 records claim an ID that an issue title also claims while naming no issue — carried over from the item's discussion, where it was a **frontmatter-only** count. This module's predicate is `namesItsIssue`, which reads the body.

Measured with the predicates the module actually ships, over the 48 IDs claimed by both a record and an issue title:

| | |
|---|---|
| **44** | the record names that issue's own number |
| **3** | the record names a **different** issue — `ARCH-037` (issue #1805) cites 1764/1773; `ARCH-039` (issue #1828) cites 1764; `HARNESS-074` (issue #1619) cites 1615 |
| **1** | the record names none at all — `HARNESS-058` (issue #1571) |

So the gap is **four citation lines in four named files**, not a wait for new records to accumulate. The header made a bounded, nameable burn-down look unbounded — and it is the number a reader would act on.

The three-citing-a-different-issue case is the more interesting one: a well-formed, machine-parseable link pointing at the wrong thing. That is the never-true shape with four live instances, and it is a stronger argument for "the link resolves **and** that issue still claims this ID" than prose about it.

## How this session got the number wrong twice

Re-deriving it here produced **39**, then **2**, then **44**. The first two used a number pattern ending in `\b` — which matches nothing after a digit, so a record citing `#1617` failed to match `#1617`. Three measurements of one population, two wrong through the same one-character defect, and the one that held was produced independently by a reviewer and then reproduced here.

That post-mortem is in the header, not only in the commit: the next person to re-measure will reach for the same pattern.

## The fourth instance of this defect happened while reviewing the fix

The three collisions #1916 was opened for were between two **clones**, minutes apart — which invites the reading that the defect needs a second machine. It does not.

PR #2011 was reviewed at one head, given a clean verdict, and merged — and the head had moved by one commit in between. The merge took the corrected code and left the uncorrected header: **this module's own subject with the halves swapped**, the half a future reader trusts being the wrong one.

Seconds, one pull request, one person. Re-reading the head immediately before writing would have closed it. Recorded in the module so the next reader does not treat multi-clone work as the precondition.

`verify-like-ci` 12/12 PASS.

https://claude.ai/code/session_017uvyAk1oL2RmY2LCXJSq9s

### DIFF
--- a/scripts/harness/task-record-issue-link.mjs
+++ b/scripts/harness/task-record-issue-link.mjs
@@ -8,10 +8,22 @@
  *
  * ## The missing piece was never the network
  *
- * Measured when this landed: 48 IDs are claimed by both a record and an issue title, and 39 of those
- * records carry no reference to that issue's number — yet they are overwhelmingly the SAME item
- * written twice, not two items. So a scan over the union of both sources reports 39 false
- * collisions. What is absent is the LINK, and 87 of 798 records have one in any form.
+ * Measured with the predicates this module ships — `namesItsIssue` and `ISSUE_LINK_PATTERNS` — over
+ * the 48 IDs claimed by BOTH a record and an issue title:
+ *
+ *   44  the record names that issue's own number
+ *    3  the record names a DIFFERENT issue (`ARCH-037`→#1805 cites 1764/1773; `ARCH-039`→#1828
+ *       cites 1764; `HARNESS-074`→#1619 cites 1615)
+ *    1  the record names none at all (`HARNESS-058`→#1571)
+ *
+ * So the gap is FOUR citation lines in four named files, not a wait for new records to accumulate.
+ * A first cut of this header said 39, which was a frontmatter-only count carried over from the
+ * item's discussion, and two attempts to re-derive it here produced 39 and then 2 — both from a
+ * `#\d+\b` pattern whose `\b` does nothing after a digit. The number that survived was the one a
+ * reviewer measured independently and this session then reproduced. Recorded because the wrong
+ * figure was the load-bearing one: it made the burn-down look unbounded when it is four lines.
+ *
+ * Across the whole tree the link is on 87 of 798 records.
  *
  * Requiring it makes the cross-source comparison exact: an ID claimed by a record and by an issue
  * the record names is one item; an ID claimed by a record and by an issue it does NOT name is two.
--- a/scripts/harness/task-record-issue-link.mjs
+++ b/scripts/harness/task-record-issue-link.mjs
@@ -25,6 +25,18 @@
  *
  * Across the whole tree the link is on 87 of 798 records.
  *
+ * ## The best instance of this defect happened while fixing it
+ *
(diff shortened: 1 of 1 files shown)


woojubb:
## Review — base `dec38cd3c`, head `01ae68cc6`

**ACTIONABLE FINDINGS: 1** (1 MUST)

The header restoration is right and the fourth-instance paragraph is better written than the account I gave you. One number in it is wrong, and it is wrong in the class this module documents.

### MUST — `44 / 3 / 1` is `44 / 2 / 2`, and the miscount is a bare `#N` read as an issue

The header ships:

```
 *    3  the record names a DIFFERENT issue (`ARCH-037`→#1805 cites 1764/1773; `ARCH-039`→#1828
 *       cites 1764; `HARNESS-074`→#1619 cites 1615)
```

Classified through the issues API (`gh api repos/woojubb/robota/issues/<n> -q 'if .pull_request then "PR" else "ISSUE" end'`):

| number | kind | title |
| --- | --- | --- |
| `#1764` | **issue** | contract hygiene: three pass-through re-exports and an unnamed… |
| `#1773` | **PR** | fix(subagent): derive the subagent↔background-task… |
| `#1615` | **PR** | fix(harness): publish a verdict, delete a copy, label three… |

So:

- **`HARNESS-074` names no issue at all.** Its only reference is `#1615`, and `#1615` is a pull request. It belongs in the "names none" bucket, not the "names a different issue" bucket.
- **`ARCH-037` cites one different issue, not two.** `#1764` is an issue; `#1773` is a pull request. Listing them as `1764/1773` reads as two issues.

Corrected: **44 name their own issue, 2 name a different issue (`ARCH-037` and `ARCH-039`, both `#1764`), 2 name none (`HARNESS-074`, which cites only a PR, and `HARNESS-058`, which cites nothing).**

`namesItsIssue` agrees with the corrected split rather than the shipped one: `ISSUE_LINK_PATTERNS` requires `issue #N` or an issues URL, and a bare `#1615` matches neither. The header therefore states a number its own predicate would not produce — which is the defect the same header is in the middle of narrating. The `\b` post-mortem you added says a re-derivation must be done with the shipped predicate; this line was derived with a looser one.

`HARNESS-074` is worth keeping as the named example rather than dropping: it is the live case for *why* a bare `#N` must not count. A record whose only reference is a pull request would move into the linked column under any pattern that accepts `#N`, and it would do so while naming nothing that could ever be compared against an issue title.

### Not blocking

The fourth-instance paragraph is accurate and I would not soften it. One factual addition if you want it: the head had moved by **one commit** and the cause was mechanical — the re-review, the verdict and the merge were three separate calls, and nothing between them re-read `headRefOid`. The fix is not vigilance, it is reading the head in the same step that merges.

### Verified, not assumed

The four records' citations were read from the files at `origin/develop`, and every number classified through the issues API rather than by shape.

Issue #2014, opened 2026-08-22
The dependency-review licence exemption names 2 of this repository's 56 dual-licensed packages, so any example importing a third fails the gate


woojubb:
## The gap, measured

`.github/workflows/dependency-review.yml` allows a fixed list of licences and exempts specific
packages from it:

```yaml
allow-dependencies-licenses: >-
  pkg:npm/%40robota-sdk/agent-core,
  pkg:npm/%40robota-sdk/agent-provider,
  pkg:npm/%40img/sharp-libvips-...
```

This repository's own packages are published under `AGPL-3.0-only OR LicenseRef-Commercial`, which
is deliberately **not** in `allow-licenses` — you do not want to accept AGPL from a third party. So
every first-party package needs to be exempted by name.

```
first-party packages under packages/                        58
carrying "AGPL-3.0-only OR LicenseRef-Commercial"           56
named in allow-dependencies-licenses                         2   (3.6%)
```

**Two of fifty-six.** The other 54 are exempt only by never having been added to a manifest the
action reads.

## How it surfaces, and why it is invisible day to day

It fires when a `package.json` the action can see gains a dependency on a first-party package that
is not one of the two. That happened here:

```
examples/capabilities/sandboxed-tools/package.json » @robota-sdk/agent-framework
  – License: AGPL-3.0-only OR LicenseRef-Commercial
##[error]Dependency review detected incompatible licenses.
```

On a `develop` pull request this usually stays quiet, because the action diffs base against head and
the example already exists in the base. **On a promotion it does not**, because the base is `main`,
which can be a hundred commits behind, so every first-party dependency added since the last
promotion is "new" at once.

So the gate is effectively evaluated on the rarest PR shape in the repository — the same structural
problem as issue #1984, arrived at from a different direction. It failed on promotion #1895 and again
on promotion #2013.

## Why the shape is wrong, not just the contents

Adding `agent-framework`, `agent-tools` and `agent-tool-defaults` clears today's failure and
reinstates the defect: the list must be extended by hand every time an example, app or fixture
reaches for a package that has not been reached for before, and nothing tells you it needs
extending until a promotion goes red.

The population is knowable — it is every `packages/*/package.json` whose `license` field is the dual
licence. Anything that derives the exemption from that, or asserts the list covers it, converts a
silent gap into a mechanical one. A scan asserting `named ⊇ dual-licensed` would have reported
`2 of 56` on any day in the last several weeks.

## Not fixed here, deliberately

[...]

PR #2017, opened 2026-08-22
docs(tasks): INFRA-104's TC-10 ran on a real promotion and did NOT observe

woojubb:
## What happened

Promotion PR #2013 merged; `main` is now `15800bf0f`, 89 commits. The three conditions the INFRA-104
record fixed **this morning, before any of this evidence existed**, were executed against it.

| # | condition, as written in advance | outcome |
| - | -------------------------------- | ------- |
| 1 | the promotion body carries a non-empty block naming issue #1965 | **NOT MET** |
| 2 | the `promotion closes` job reports green | MET — `pass`, 40s |
| 3 | issue #1965 is closed by GITHUB, attributed to the merge | **NOT MET** |

Cause, read from the timeline rather than inferred:

```
issue #1965   closed_by=woojubb   commit=none   at 2026-08-22T01:42:25Z
```

A person, **seventeen minutes** after the pull request carrying the keyword merged and ninety
minutes before the promotion. Not a commit reference; not GitHub acting on a keyword.

This is the branch the record named in advance:

> if issue #1965 is closed by hand before then, the subject is consumed and TC-10 waits again

Fifth occurrence in one session. Writing the failure branch first is the only reason this outcome
cannot now be reinterpreted.

## The finding

**TC-10 as specified is not merely unobserved — it is unobservable**, and this entry says so rather
than deferring to a sixth promotion.

- Every closing keyword in this repository sits on a `develop` PR, because that is where work merges.
- GitHub honours closing keywords only on a **default-branch** PR, so the keyword does nothing.
- The issue therefore stays open for hours or days after its fix has shipped, which **reads as an
  oversight** to anyone tidying the backlog — closing it is the correct-looking action.
- The promotion arrives later and finds nothing to close.

The window between "fix merges" and "promotion runs" is exactly the window TC-10 needs the subject
to survive, and nothing protects it. Nobody did anything wrong.

**Repair, stated but not taken here:** either the observation keys on something durable — the
derivation recorded at merge time, or the merge event itself — or TC-10 states what it can actually
verify: *the derivation runs and produces the correct block when a subject exists.* The subject's
existence is not something a gate can require of the world.

## What DID work

The gate half was exercised on a real `main` pull request: `promotion closes` derived from **137
carried bodies and 38 issue records** and reported an empty requirement correctly — "none, and that
is a measured result" — not a skip, and not an underivable block reported as empty. GitHub's half
has never been reached, and the record now says why that is structural rather than unlucky.

## Filed while doing this, not folded in

- issue #1980 — `promotion closes` is declared required on `main`; the live ruleset does not require
  it. Ruleset unchanged since 2026-07-26, so it was **never added**, not removed.
- issue #1984 — the lint ratchet's only CI execution path is the promotion job.
- issue #2014 — the dependency-review licence exemption names 2 of this repository's 56
  dual-licensed packages.

ACTIONABLE FINDINGS: 0

Refs INFRA-104

### DIFF
--- a/.agents/tasks/INFRA-104-promotion-carries-closing-keywords-to-main.md
+++ b/.agents/tasks/INFRA-104-promotion-carries-closing-keywords-to-main.md
@@ -198,3 +198,66 @@ BEFORE the evidence exists so the reading cannot be fitted to it:
 Any of the three failing is a finding about this item, not about the promotion. In particular,
 if issue #1965 is closed by hand before then, the subject is consumed and TC-10 waits again — which is what
 happened four times earlier in this session and is recorded above.
+
+### 2026-08-22 (later) — the promotion happened, and TC-10 did NOT observe
+
+Promotion PR #2013 merged at `03:21:49Z` (`main` is now `15800bf0f`, 89 commits). The three conditions
+written above BEFORE any of this evidence existed were executed against it. Two of the three failed,
+and the record says so rather than being rewritten around the result.
+
+| #   | condition, as written this morning                              | outcome                          |
+| --- | --------------------------------------------------------------- | -------------------------------- |
+| 1   | the promotion body carries a non-empty block naming issue #1965 | **NOT MET** — the block is empty |
+| 2   | the `promotion closes` job reports green                        | MET — `pass`, 40s                |
+| 3   | issue #1965 is closed by GITHUB, attributed to the merge        | **NOT MET** — closed by a person |
+
+The measured cause, read from the timeline rather than inferred:
+
+```
+issue #1965  closed_by=woojubb  commit=none  at 2026-08-22T01:42:25Z
+```
+
+`commit=none` — not a commit reference, not GitHub acting on a keyword. **Seventeen minutes after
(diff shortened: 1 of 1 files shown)


woojubb:
## Reviewer verdict

**Base** `9159ab59a88495ec870723f12f2965d28475b73c` · **Head** `a9f041dc0a1df46bde5455d7da297e596797202f`

CI: **16 passed, 0 failed**, `mergeStateStatus=CLEAN`. Threads: 0.

Docs-only — one dated entry on the INFRA-104 task record. No code, no scan, no gate.

Every figure in it was executed rather than restated: `gh pr checks 2013` for the job verdict,
the issue timeline for `closed_by` and `commit_id`, and `git diff b5fe541d3 <head>` for the
byte-identical tree. The `reference-kind-qualified` ratchet was re-run **after** prettier reflowed
the file at stage time, which is where it caught a bare reference earlier today.

The entry records a **failed** observation and says so in its own heading. Two of the three
conditions did not hold. I checked specifically that it does not soften that anywhere, because the
value of fixing the conditions in advance is entirely destroyed by adjusting the reading afterwards.

ACTIONABLE FINDINGS: 0

PR #2029, opened 2026-08-22
docs(tasks): INFRA-046's blocking evidence arrived, and INFRA-054's preconditions moved

woojubb:
Both items were blocked on conditions their own records told the next reader to **re-measure rather
than carry forward**. Followed that instruction.

## INFRA-046 — the evidence condition is MET

The block was stated exactly, which is what makes it dischargeable:

> `accidental-green` has never fired on a real pull request … **One observed firing is what promotes it.**

It fired twice, today, on PR #1983:

```
$ gh run list --branch feat/infra-126-temp-dir-owner
  01:30  CI  failure   -> regression-red-proof (enforcing: accidental-green only)  FAILURE
  01:44  CI  failure   -> regression-red-proof (enforcing: accidental-green only)  FAILURE
  02:02  CI  success   -> regression-red-proof (enforcing: accidental-green only)  success
```

Verdict read out of the failing run's log rather than from anyone's report of it:

```
❌  scripts/harness/run-all-scans.mjs: accidental-green-fail (all-pass)
❌ accidental-green: a regression test passes even with the fix reversed — it guards nothing.
```

**It was right.** The author's `run-all-scans.mjs` registration hunk could be reversed with every
changed test still passing — the wiring was proven only by hand. Two attempted fixes landed in files
the source-to-test pairing cannot reach; the one that satisfied it asserted against the imported
`SCAN_COMMANDS`. Three other verdicts in the same run reported `inconclusive` **with the reason**
("the range added this file and never revised it, so there is no earlier state to reverse to … which
is not a verdict"), which is the property that makes a refusal from this gate worth trusting.

Not a flake, not a false positive, not a self-inflicted probe.

## INFRA-054 — one precondition moved, and it does not unblock the item

| precondition (verified 2026-07-26) | today |
| --- | --- |
| non-merge commits on `main` unseen by `develop` | **10 → 0** |
| the operating account can push directly to `main` | yes, unchanged |
| `ci.yml` triggers only `on: pull_request` | yes, unchanged |

The entry says plainly why zero is not progress toward "safe now": the measurement is about what
**has** landed; the blocking condition is about what **can**. The bypass actor
(`RepositoryRole id=5`, `bypass_mode: always`), the `hotfix/*` route (`ci.yml:100`) and
Dependabot-disabled-**by-policy** (a marker file, not a gate) are all intact. A fast-forward
promotion made safe by "nobody has done it lately" is safe until the first `hotfix/*`.

## Both stay `blocked`

What changed is that the recorded reasons now describe what is true today. INFRA-046's remaining
half is a branch-protection decision, not an evidence gap.

## Citation note

The failing head SHAs are deliberately **not** quoted: the branch was deleted on merge, so they
resolve for nobody. `commit-msg`'s `claims-resolve` check caught that in the first draft of the
commit message, and the same correction was applied to the record.

ACTIONABLE FINDINGS: 0

Refs INFRA-046, INFRA-054

### DIFF
--- a/.agents/tasks/INFRA-046-promote-advisory-gates.md
+++ b/.agents/tasks/INFRA-046-promote-advisory-gates.md
@@ -344,3 +344,64 @@ with the owner on 2026-08-22, who declined to add it now, knowing the permission
 
 So the block is EVIDENCE, and only evidence. The earlier framing — "blocked on evidence first, a
 credential second" — was half wrong: there was never a credential half.
+
+## 2026-08-22 (later) — the evidence this item was blocked on HAS ARRIVED
+
+The block was stated precisely, so it can be discharged precisely:
+
+> `accidental-green` has never fired on a real pull request, so requiring it would put an untested
+> refusal into the merge path of every PR. **One observed firing is what promotes it.**
+
+**It fired. Twice, on a real pull request, today.** PR #1983 (`feat/infra-126-temp-dir-owner`):
+
+```
+$ gh run list --branch feat/infra-126-temp-dir-owner
+  01:30  CI  failure     -> regression-red-proof (enforcing: accidental-green only)  FAILURE
+  01:44  CI  failure     -> regression-red-proof (enforcing: accidental-green only)  FAILURE
+  02:02  CI  success     -> regression-red-proof (enforcing: accidental-green only)  success
+```
+
+The head SHAs are deliberately not quoted here: the branch was deleted on merge, so they resolve
+for nobody. The durable citation is the pull request and its check runs.
+
+The verdict, read out of the failing run's log rather than from anyone's report of it:
+
+```
+❌  scripts/harness/run-all-scans.mjs: accidental-green-fail (all-pass)
+❌ accidental-green: a regression test passes even with the fix reversed — it guards nothing.
+   Rewrite it to FAIL on the pre-fix code, or opt out with `allow-green-at-base: <reason>`.
+```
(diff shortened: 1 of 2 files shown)


woojubb:
## Reviewer verdict

**Base** `ea7ce0b53340001fb42311a06b818a52772d6cac` · **Head** `8552cebedce2ff3c3f7dd3e4fbcb2b6cf36aa434`

CI: **16 passed, 0 failed**, `mergeStateStatus=CLEAN`. Threads: 0. Docs-only — two task records, no code.

Both entries were produced by executing the conditions the records themselves said to re-derive, not by restating them:

- INFRA-046's firing was read out of the failing run's log (`gh run view --job … --log-failed`), and the surrounding `inconclusive` verdicts were read with it, because a gate that distinguishes "cannot judge" from "judged and passed" is what makes its refusal worth acting on.
- INFRA-054's three preconditions were each re-run: `git rev-list origin/develop..origin/main --no-merges` (10 → 0), the ruleset's `current_user_can_bypass` (still `always`), and `ci.yml`'s trigger set (unchanged).

The entry that matters most is the one saying why 10 → 0 does **not** unblock INFRA-054 — the measurement is about what has landed, the condition is about what can. A number falling to zero reads as progress, and here it is a consequence of nobody having used the direct-write paths lately rather than of their removal.

Both items remain `blocked`. Nothing here marks anything done.

ACTIONABLE FINDINGS: 0

Issue #1719, opened 2026-08-13
INFRA-097: trusted provenance for required PR workflows


woojubb:
---
title: 'INFRA-097: trusted provenance for required PR workflows'
status: todo
created: 2026-08-14
priority: high
urgency: soon
area: .github/workflows, repository rulesets, scripts/harness
depends_on: []
---

# INFRA-097: trusted provenance for required PR workflows

## Objective

Required checks triggered by `pull_request` load their workflow YAML from the PR merge revision.
Consequently a PR can modify the workflow control plane that reports its own required context; checking
governance scripts out from the base SHA is defense in depth but does not establish trusted workflow
provenance. Design a repository-level required-workflow boundary whose control plane is trusted while
never executing untrusted PR content with write credentials.

## Plan

- [ ] Verify GitHub event/check attribution and ruleset capabilities for trusted required workflows.
- [ ] Compare required-workflow/ruleset, split `pull_request_target`, and external GitHub App/check-run designs.
- [ ] Specify exact SHA/ref identity, permissions, fork behavior, and branch-protection reconciliation.
- [ ] Add adversarial tests proving a PR cannot replace its own required gate with an unconditional pass.
- [ ] Roll out and validate code, docs-only, fork, retarget, cancellation, and genuine-finding paths.

### 2026-08-14

- Discovered during INFRA-096 Round A: exact-base checkout protects loaded scripts but not the
  `pull_request` workflow YAML itself.
- INFRA-096 records labelled containment because this is pre-existing repository-wide control-plane
  debt; it does not claim to solve workflow provenance.

## Blockers

None. Owner decisions may be required if the chosen design needs live ruleset, GitHub App, or
organization-level required-workflow configuration.

## Test Plan

- Parsed workflow/ruleset fixtures for untrusted self-edit, fork permissions, exact check attribution,
  cancellation, and missing/malformed configuration.
- Live throwaway PR proof that editing the candidate workflow cannot forge the protected context.
- Existing required-check, permission, action-reference, and full harness regressions.

## User Execution Test Scenarios

Not applicable — this is repository-internal CI governance and branch-protection infrastructure, not a
shipped CLI, TUI, browser, application, or public SDK behavior.

woojubb:
Advanced by #1823 — **not closed**, and the remainder is stated below rather than left implied.

### What landed

`scripts/harness/scan-workflow-provenance.mjs`, registered in `pnpm harness:scan`. The guarded set is
[...]

PR #2034, opened 2026-08-22
test(ci): THROWAWAY — red-proof that workflow provenance refuses a self-edit (INFRA-097)

woojubb:
## THROWAWAY — do not merge, do not review. This pull request exists to be REFUSED.

Issue #1719's Test Plan asks for a *"live throwaway PR proof"* that the control plane cannot be edited invisibly. Until PR #2030 merged, the gate's own workflow was not in the guarded set and `workflow provenance` was not a required context, so the experiment was not possible. It is now.

**I will close this without merging as soon as the result is recorded.**

## The prediction, written before any check ran

The single commit adds **three comment lines** to `.github/workflows/ci.yml` — deliberately inert, changing no job, step, condition or trigger.

1. The **`workflow provenance`** check MUST report **failure**.
2. Its log MUST name `.github/workflows/ci.yml` and MUST enumerate the required contexts that edit could move. Predicted list, derived from `.github/required-status-checks.json` at this base:
   - develop: `build`, `quality`, `scans`, `dependency audit`, `commitlint`, `tui-e2e`, `examples-typecheck`, `windows-shell`
   - main: `promotion ancestry`, `main PR source guard`, `promotion closes`, `release-grade verification`
3. The pull request MUST become unmergeable on that context — `workflow provenance` is required on `protect-develop` as of today.
4. The advisory line MUST read `2 of 3 guarded workflow(s) load their definition from the pull request`, naming `ci.yml` and `review-gate.yml` — the gate itself is the third and is NOT self-loading, because it runs on `pull_request_target`.

**A green `workflow provenance` here is a FINDING, not a relief.** It would mean either the gate cannot fail, or my edit never reached the diff the gate reads.

## That second failure mode already happened once, locally

The first local run of `scan-workflow-provenance.mjs --base-ref origin/develop --head-ref HEAD` exited **0** and printed only the advisory. The edit was in the working tree and not committed, so it was absent from `origin/develop...HEAD` — the scan ran correctly against a diff that did not contain the intervention, and reported the answer that means "clean".

After committing, the same command exits **1** with the full context list. Recorded because a no-op experiment and a passing experiment are indistinguishable in the output, and the direction of that silence is always toward reassurance. `ARCH - robota` hit the same shape twenty minutes earlier with a `.replace` that silently did not match, and the fix there is one line: assert the intervention changed something before drawing any conclusion from the result.

## What this proves and what it does not

Proves: the required context **can fail**, on a real pull request, for the reason it exists.

Does not prove: that the control plane is *trusted*. `ci.yml` still loads its definition from the pull request under test. The gate makes a self-edit unmergeable-without-notice; it does not make an edit impossible. That distinction is INFRA-097's own, and it is why the advisory in item 4 is expected rather than a defect.

## Related

- Issue #1719 / INFRA-097 — the item this closes the last step of
- PR #2030 — the change that made the context requireable

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1204,3 +1204,6 @@ jobs:
       - name: Skip (no coverable src changes)
         if: steps.detect.outputs.affected != 'true'
         run: echo "No package/app src lines changed — patch coverage is a no-op for this PR."
+
+# INFRA-097 throwaway: a deliberately inert edit, made to prove the `workflow provenance` gate
+# FAILS on a self-edit of the control plane. This branch is never merged.


woojubb:
## Result — three predictions held, the fourth failed, and the failure is the finding

| # | Predicted | Observed |
| --- | --- | --- |
| 1 | `workflow provenance` reports **failure** | **fail**, 10s |
| 2 | Names `ci.yml` and enumerates the 12 contexts the edit could move | exact, all 12, with `(develop)` / `(main)` tags |
| 3 | Pull request becomes unmergeable on that context | `mergeStateStatus: BLOCKED` |
| 4 | Advisory reads `2 of 3`, `::examined:: 3` | **`2 of 2`, `::examined:: 2`** |

So the gate **can fail**, on a real pull request, for the reason it exists. That was the question INFRA-055 exists to ask, and it is now answered by observation rather than by shape.

## Prediction 4 was wrong, and chasing it found a real defect

The gate checked out the wrong tree. From this run's log:

```
[command]/usr/bin/git checkout --progress --force -B main refs/remotes/origin/main
```

It checked out **`main`** — the repository's default branch — not `develop`, which is this pull request's base. The workflow's own comment says:

> The BASE, which is what `pull_request_target` checks out when no ref is named.

That is **not what happened**. `actions/checkout` with no `ref:` resolved to the default branch, and `main` does not contain PR #2030:

| tree | guarded workflows |
| --- | --- |
| `origin/develop` (this PR's base) | `ci.yml`, `review-gate.yml`, `workflow-provenance-gate.yml` |
| `origin/main` (what was checked out) | `ci.yml`, `review-gate.yml` |

`2` versus `3`, exactly as observed.

### Why this is a defect and not a curiosity

`BASE_SHA` is `github.event.pull_request.base.sha`, and the scan is invoked with `--base-ref "${BASE_SHA}"`. So the **diff** is computed against `develop` correctly, while the **registry and the scan code** come from `main`. Diff from the pull request's base, rules from the default branch.

Two consequences, both silent:

1. **A newly-required context is unguarded until the next promotion.** A workflow registered on `develop` is not in the guarded set the gate reads, so editing it passes. That is live right now for `workflow-provenance-gate.yml` itself.
2. **A fix to `scan-workflow-provenance.mjs` does not take effect for `develop` pull requests until it reaches `main`.** The executed code is `main`'s copy.

The security property is unaffected — `main` is not attacker-controlled either — but the gate judges by rules that can lag its base indefinitely, and nothing says so.

### The fix

Pin the checkout to the pull request's actual base: `ref: ${{ github.event.pull_request.base.sha }}`. Still the base, never the head, so it keeps the property the file exists for and makes the comment true. Landing as a follow-up.

## Note on how this surfaced

Only because the expected value was written down before the run. `2 of 2` reads as a perfectly healthy line; nothing about it invites a second look. It was a finding solely because `3` was on record beforehand — the same instrument as asserting a mutation applied before trusting the experiment that follows it.

Closed without merging; the throwaway edit to `.github/workflows/ci.yml` is discarded with the branch.

Issue #2039, opened 2026-08-22
A pull_request_target gate runs the default branch's copy of itself, so fixing it does nothing until promotion


woojubb:
## A `pull_request_target` gate runs the DEFAULT branch's copy of itself

Measured today on three separate runs. `workflow-provenance-gate.yml` executes `main`'s version of its own YAML on every pull request, including pull requests whose base is `develop`.

**Evidence.** On PR #2038 (base `develop`), the run's checkout step received no `ref:` input and did:

```
[command]/usr/bin/git checkout --progress --force -B main refs/remotes/origin/main
```

At that moment the two branches held different definitions:

| | `types:` | `ref:` on checkout |
| --- | --- | --- |
| `origin/develop` | `[opened, synchronize, reopened, edited]` | `${{ github.event.pull_request.base.sha }}` |
| `origin/main` | `[opened, synchronize, reopened]` | *(absent)* |

The run had no `ref:` input, so it was `main`'s copy. The gate's advisory read `::examined:: 2` where `develop`'s registry lists 3 guarded workflows — the count `main` produces.

## Why it matters

**A fix to this gate is inert until it is promoted.** Two landed on `develop` today and neither is active:

- **PR #2030** added `edited` to the trigger. `edited` is the only activity a base retarget fires, and GitHub's default set omits it. So the retarget hole `scan-main-required-checks` R7 exists to close — measured on PR #1442, where a branch retargeted `develop`→`main` re-dispatched nothing and reported `mergeStateStatus: CLEAN` — **is live right now** on the one context whose plane was chosen to be trustworthy.
- **PR #2035** pinned the checkout to the pull request's base. Until promotion the gate keeps reading `main`'s registry, so a context newly required on `develop` is outside the guarded set it consults.

**The security property is unaffected, and arguably stronger.** `main` is further from a pull request's reach than `develop` is. Trusted provenance holds. What does not hold is the assumption that fixing the gate fixes the gate.

**The surprise is the cost.** Both fixes were verified locally, both landed green, and both are inert. Nothing in the repository says that a `pull_request_target` workflow has a different activation path from every other workflow here, and the file's own comment asserted the opposite of what it does.

## What would close it

Not decided — filing so the mechanism is written down rather than rediscovered:

1. **Document it and accept the promotion latency.** Cheapest. Every `pull_request_target` gate carries a note that its edits activate on promotion, and the item that edits one says so in its own record.
[...]

PR #2041, opened 2026-08-22

### LINKED ISSUE #1719: INFRA-097: trusted provenance for required PR workflows

### LINKED ISSUE #1980: `promotion closes` is declared required on main and is not required — the reconciler that would catch it is dispatch-only

### LINKED ISSUE #2018: P0 security: untrusted project settings can execute startup hooks and redirect user provider credentials
## Summary

Opening or running agent-cli inside an untrusted checkout gives that checkout's .robota/settings.json the same authority as user-global configuration. Project settings are loaded after user settings, can select full trust, can allow shell tools, and can register SessionStart command hooks that run automatically. Provider objects are also field-merged, so a project can replace a provider base URL while retaining an API key loaded from the user's higher-trust profile.

This is a trust-boundary violation for a consumer CLI: merely entering a repository and starting Robota can grant repository-controlled configuration code-execution authority or route authenticated provider traffic to an attacker-controlled endpoint.

## Evidence

- [The project layer is loaded after the user layer](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/config/config-loader.ts#L243-L255).
- [The settings contract permits defaultTrustLevel, hooks, permissions and providers](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/config/config-types.ts#L104-L152).
- [Provider entries are merged field by field](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/config/config-loader.ts#L168-L199), preserving fields such as a user API key when a lower-trust layer supplies only baseURL.
- [Session construction starts SessionStart hooks](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-session/src/session.ts#L162-L170).
- [Command hooks spawn a shell with process.env](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-core/src/hooks/executors/command-executor.ts#L38-L40).

A non-executing reproduction loaded a project file containing full trust, Bash(*) and a SessionStart command; all three appeared in the resolved configuration. A second probe put an API key and normal base URL in the user profile and only an attacker URL in the project profile; the resolved provider retained the user key and used the project URL.

## Security impact

- Arbitrary command execution at CLI startup when a victim runs the CLI in a malicious checkout.
- Exfiltration of provider credentials and prompts through project-controlled provider endpoints.
- Silent weakening of user permission policy.

Severity is P0 because the trigger is normal consumer behavior and no workspace-trust decision separates repository-owned data from user-owned secrets and execution policy.

## Required improvement

Introduce a first-class workspace trust boundary before any project configuration is applied. Until trust is established, repository configuration must not be allowed to register executable hooks, raise trust, relax user deny rules, load plugins, inject secrets, or override authenticated provider endpoints. Track configuration provenance per field and prevent lower-trust layers from inheriting secret-bearing fields when changing an endpoint. The trust decision must be bound to a canonical repository identity and must be revocable.

## Acceptance tests

- Starting in a fresh untrusted repository does not execute SessionStart commands.
- Project configuration cannot raise a user or CLI trust level or remove a user deny rule.
- Changing a provider base URL in a project never carries over a secret from a higher-trust layer.
- Trusting one canonical repository does not trust a symlink alias or a different repository at the same textual path.
- Headless and interactive startup enforce the same boundary.

woojubb:
Built with `node scripts/harness/promote.mjs`, not by hand.

`origin/develop ed31a5ebb` → `origin/main 15800bf0f`, promoting develop's tree unchanged.
A1/A2/A3 hold.

Closes #1980
Closes #2018
Closes #1719

**This block is non-empty for the first time in a long while, and that is the point of it.** All
three were confirmed OPEN through the issues API at branch-build time. GitHub honours a closing
keyword only on a default-branch pull request, so the `Closes` lines each of these wrote on their own
`develop` pull request closed nothing — this promotion is the only pull request that can act on them.

`promotion closes` is **required** on `protect-main` as of today, so for the first time its verdict
has consequences rather than being advisory.

## Observation conditions, fixed BEFORE this merges

INFRA-104's TC-10 asks whether GitHub — not a person — closes an issue through this block. It has
never been observed: five times this session the subject was closed by hand in the window between its
fix merging and a promotion running, once seventeen minutes after. So the reading is fixed now:

1. the body carries a non-empty block naming all three;
2. `promotion closes` passes;
3. each of the three is closed **by GitHub**, attributed to the merge — the timeline event carries a
   `commit_id`, and `closed_by` is not a person acting by hand.

**A partial result is a failure.** Three subjects closing two issues is not "mostly observed".

## Pre-flight

```
$ pnpm harness:verify:release > log 2>&1; echo "REAL EXIT: $?"
REAL EXIT: 0
```

Captured by redirect and `$?`, not read off a piped `tail` — an earlier run's reported "exit 0"
belonged to `tail`, not to the verification. Zero `ELIFECYCLE` lines in 8k lines of log. Lint 2092
against a ceiling of 2092.

## The reconciliation ran while this branch was built

```
promote: required-status-check declarations reconcile against the live rulesets.
```

First clean reconciliation either session has seen. `.github/required-status-checks.json` and both
live rulesets now agree — the two halves of issue #1980, fixed within an hour of each other by two
sessions who each read before writing and read back after.

## What this promotion activates

A `pull_request_target` workflow runs the **default branch's** copy of its own YAML, not the pull
request base's (issue #2039). So `develop`'s `edited` trigger and base-pinned checkout are not live
yet — **this merge is what activates them**. Verified before opening: `main` already carries
`workflow-provenance-gate.yml` and `scan-workflow-provenance.mjs`, and its `branches:` covers `main`,
so the newly-required `workflow provenance` context reports here rather than sitting permanently
pending.

## Merge method

`gh pr merge <n> --merge`. Never `--squash` — `protect-main` sets
`allowed_merge_methods: ["merge"]`, and a squash destroys the ancestry link `promotion ancestry`
exists to preserve. Never GitHub's "Update branch" button; re-run `promote.mjs` instead.


woojubb:
## Reviewer verdict

**Base** `15800bf0f` (`main`) · **Head** `24ddda887f97ecaa6842f60d295dfa3c5cc9ca1e`

**All five required contexts pass**, including two that are required here for the first time:

```
promotion ancestry          pass   11s
main PR source guard        pass    2s
promotion closes            pass   16s     <- required as of today
release-grade verification  pass   11m1s
workflow provenance         pass   12s     <- required as of today
```

15 passed, 0 failed, `mergeStateStatus=CLEAN`. Unlike the previous promotion, neither `CodeQL` nor `Dependency review` is red — the delta is small enough that neither trips on scale.

**`promotion closes` passing over a NON-EMPTY requirement is the first of its kind.** Every previous run reported "no issues implied", correctly but vacuously. This one derived three, confirmed each is an open issue through the API, and verified the body carries a keyword for all three.

Provenance checked rather than assumed: the `Closes #1719` line comes from PR #2030, which is inside this promotion's range. PR #2038 merged after this branch was cut and is NOT in it — it ships next promotion. The issue closes either way; I checked because a body line whose source is outside the range would mean the derivation read something it should not have.

All three issues confirmed **still open** immediately before this merge, so the subjects have not been consumed by hand as they were five times earlier today.

The three observation conditions are recorded in the description, fixed before the evidence exists. A partial result is a failure.

**Merge method:** `--merge`, never `--squash`.

ACTIONABLE FINDINGS: 0

Issue #1722, opened 2026-08-13
ARCH-029: decompose command-host capability contracts


woojubb:
---
title: 'ARCH-029: command-host consumers claim a 46-member optional facade instead of their declared capabilities'
status: todo
created: 2026-08-14
priority: high
urgency: soon
area: packages/agent-framework, packages/agent-command, packages/agent-command-workflows, packages/agent-cli, packages/agent-transport-tui
depends_on: [ARCH-012]
issue: pending
---

# ARCH-029: decompose the command host into explicit capability contracts

## Problem

`ICommandHostContext` currently has 46 members, 32 of them optional, spanning preset mutation,
history, interaction, adapters, context references, skills, checkpoints, memory, background work,
goals, plans, agent jobs, and terminal handoff. A command needing one role must accept the entire
facade and tests construct partial objects through casts. Production also projects
`InteractiveSession` through `this as unknown as ICommandHostContext`.

The issue reproduces by type-checking a command with a minimal honest host: the command signature
requires unrelated methods, while optional methods cannot distinguish an absent capability from a
provided capability whose result is empty.

## Why this is foundational

The recurring cause is the framework-owned host contract, not any individual command. Fixing one
cast or making another method optional leaves every sibling command under the same pressure. This
item was split from ARCH-012 after an independent depth audit found two distinct owners and migration
graphs.

## Direction

- Define framework-owned named command-host roles and a typed capability map/query.
- Make each command consume only its required roles and handle absence explicitly.
- Provide one production adapter from `InteractiveSession` and reusable exact-role test hosts.
- Remove the production self-cast and every direct `ICommandHostContext` partial cast; add a zero
  mechanical floor.
- Preserve provided-empty results as distinct from capability absence.

## Test Plan

- Type-level RED: a minimal workspace-only or preset-only host is assignable to the corresponding
  command role without a cast; current code must fail before the new surface exists.
- Runtime tests distinguish absent capability from provided `null`/`undefined`/empty-array results.
- Existing command suites migrate to exact capability fixtures and remain green.
- The contract-cast scanner covers `ICommandHostContext` at an exact zero baseline and mutation
  fixtures prove either canonical cast is rejected.
- A public command-path scenario exercises a capability-present and capability-absent path through
  the real framework adapter.

## User Execution Test Scenarios

Applies. Exact agent-executable commands, prerequisites, output, cleanup, and empty evidence fields
will be authored at this item's scenario-planning gate before implementation.

## Plan

[...]

PR #2060, opened 2026-08-22
docs(tasks): INFRA-104 is done — TC-10 observed on a real promotion

woojubb:
## TC-10 was the last unmet criterion, and it is observed

TC-01 through TC-09 were already met. TC-10 asked for the one thing no local run can produce: a real
promotion whose derived block causes **GitHub, not a person**, to close the issues it names.

It went unobserved **five times** this session — each time the subject was closed by hand in the
window between its fix merging and a promotion running, once seventeen minutes after.

Observed on promotion PR #2041:

```
promotion merged                  05:23:35Z
issue #1980  closedAt 05:23:36Z   closer = PullRequest #2041
issue #2018  closedAt 05:23:37Z   closer = PullRequest #2041
issue #1719  closedAt 05:23:37Z   closer = PullRequest #2041
```

Three subjects, three closures, each attributed by GitHub to the promotion. The bar written in
advance — *a partial result is a failure* — did not have to be invoked.

## The scoring instrument was wrong, and the record says so

Condition 3 was fixed in writing beforehand as *"the timeline event carries a `commit_id`, and
`closed_by` is not a person"*. Run literally it returned `closed_by=woojubb commit_id=NONE` for all
three — **the failure branch, for the sixth time**.

That reading would have been wrong. `commit_id` is populated when a commit **message** closes an
issue; a closing keyword in a pull-request **body** is recorded differently. The field that answers
the question is `ClosedEvent.closer`.

So the condition was right and the field chosen to operationalise it could not distinguish the two
cases it was about — the same defect this item spent the day cataloguing, in the scoring rule of the
test built to catch it.

**The correction is legitimate only because of how it was made:** by querying a *different field*,
not by softening the condition; and because `closer` would have been chosen had the first result read
PASS. Loosening a threshold until the result flips looks identical in the diff and is the opposite
act.

## The corrected field was validated against a known negative

```
CONTROL  issue #1965 (closed BY HAND the same day)  closer = NULL              actor = woojubb
TEST     issue #1980 / #2018 / #1719                 closer = PullRequest #2041  actor = woojubb
```

`actor` is identical across all four and carries no information — every session on this machine
authenticates as the same account. `closer` separates them cleanly. A discriminator that had matched
the control too would have proved nothing, which is why it was run first.

A second session reached the same verdict independently through a different field — a `referenced`
event carrying the promotion's merge commit on all three.

## Qualified, not overclaimed

The window was held **deliberately**: another session was asked not to close issue #1719 by hand and
did not. Nothing in the workflow protects that window, and the open issue still reads as an oversight
to anyone tidying.

**This proves the mechanism works. It does not prove the mechanism is reliable.** The record says so
in those words, and the structural finding from the previous entry stands.

## Housekeeping

Status `done`, `completed` dated, record archived. Two inbound links and the `reference-kind`
baseline entry repointed in the same change — the five unqualified references are the frozen count
moving with the file, not new debt. 134 scans pass.

ACTIONABLE FINDINGS: 0

Refs INFRA-104

### DIFF
--- a/.agents/memory/reference-kind-is-named.md
+++ b/.agents/memory/reference-kind-is-named.md
@@ -37,7 +37,7 @@ different enforcement shapes, and the write surface is always the cheap one.
 ## The exemption that is load-bearing
 
 `Closes #N` is exempt because GitHub parses that exact shape, and
-[INFRA-104](../tasks/INFRA-104-promotion-carries-closing-keywords-to-main.md) built the promotion
+[INFRA-104](../tasks/completed/INFRA-104-promotion-carries-closing-keywords-to-main.md) built the promotion
 machinery that carries those keywords to the default branch so a finished issue closes itself.
 Requiring a qualifier there would have traded a readability gain for a broken automation — the kind
 of trade a new rule makes silently when its author only looks at the surface being improved.
--- a/.agents/rules/publish.md
+++ b/.agents/rules/publish.md
@@ -104,7 +104,7 @@ anything, and every keyword written on a feature pull request is prose.
 Enforced by: `scripts/harness/scan-promotion-closes.mjs`, run as the `promotion closes` job in
 `.github/workflows/ci.yml` and required on `protect-main`. It re-derives the requirement from the
 live pull request and blocks a body that omits any of it. Record:
-[INFRA-104](../tasks/INFRA-104-promotion-carries-closing-keywords-to-main.md).
+[INFRA-104](../tasks/completed/INFRA-104-promotion-carries-closing-keywords-to-main.md).
 
 ### CI Failure Triage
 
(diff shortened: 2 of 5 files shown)


woojubb:
## Reviewer verdict

**Base** `02c5e08e474e75a7c7167d3a2ba8e3947ed88c91` · **Head** `301141bd1826a81b45bd26e2550a5435b6bef479`

CI: **20 passed, 0 failed**, `mergeStateStatus=CLEAN`. Threads: 0. Docs plus one baseline path move.

The substance is a criterion being ticked, so the thing worth reviewing is whether the tick is earned:

- TC-01..TC-09 were already ticked before this change; only TC-10 moves.
- TC-10's evidence is `ClosedEvent.closer` naming PR #2041 on all three subjects, **validated against a hand-closed control** (issue #1965, `closer = NULL`) rather than trusted. A discriminator that matched the control too would have proved nothing.
- The tick's text records what actually happened, not what the criterion predicted — it named `#1722` as the first case and that subject was consumed by hand long before.
- The entry states that the scoring instrument was wrong and how, rather than quietly using the field that worked.

The record does **not** claim reliability. It says the window was held deliberately and that nothing in the workflow protects it. That distinction is load-bearing and survived into the archived text.

Archive housekeeping checked: two inbound links repointed, and the `reference-kind` baseline entry moved with the file so its five frozen references are not re-counted as new debt. 134 scans pass.

ACTIONABLE FINDINGS: 0

PR #2130, opened 2026-08-22
fix(harness): a declared context name is checked on every branch, not only main

### LINKED ISSUE #2036: A develop context's declared name is checked by nothing, and the two waiting to be promoted are both wrong
## The trap

`scripts/harness/scan-main-required-checks.mjs` **R1** exists to prevent one specific disaster:

> Branch protection matches on the context NAME, so a required context nothing publishes never
> reports and blocks the PR forever.

R1 is applied to `main` only. `GOVERNED_BRANCH = 'main'`, and `findRequiredCheckFindings` iterates
`readDeclaration(root, GOVERNED_BRANCH)`. **`develop`'s ten declared contexts get no name check from
anything.** `ci-mirror-map.mjs` governs that side and asserts each declared context is mirrored by a
verification stage — a different question entirely.

## Measured: the two contexts queued for promotion are both misnamed

```
declared in deliberately_not_required        actually published
  patch-coverage                               patch-coverage (advisory)
  regression-red-proof                         regression-red-proof (enforcing: accidental-green only)
```

Neither is required today, so neither is currently harmful. **Both are staged to become required**,
and each carries its own reason to be promoted:

- `regression-red-proof` — INFRA-046's blocking evidence arrived today; `accidental-green` fired
  twice on PR #1983 and caught a real unguarded registration. Owner has approved promoting it.
- `patch-coverage` — held advisory over two open defects, but the same move applies when they close.

Promoting either **by moving its existing entry** into `required_status_checks` requires
`protect-develop` to match a context name nothing publishes. Every `develop` pull request would then
carry a permanently pending check and **could never merge** — R1's exact failure mode, on the branch
R1 does not cover.

This was found while preparing that promotion. The entry looks like the thing to move; it is not.

## Also worth noting: how nearly this got mis-filed

The first sweep compared the declared list against `gh pr checks` on an **in-flight** pull request
and reported `review-gate` as unpublished too. It publishes — it simply had not reported yet. A
mid-CI snapshot is not the population, and re-running against a settled pull request corrected it
from three findings to two.

## Fix

Two halves, and the second is the one that lasts:

1. Correct the two entries to their published names, so a future promotion is a move rather than a
   trap.
2. Extend the R1 name check to every declared branch rather than only `main`. R1's reasoning is not
   `main`-specific — it is a fact about how branch protection matches, which is identical on
   `develop`. The `main`-only scope is a contingent fact about which branch was being hardened when
   it was written, hardened into the rule itself.

The second half is what makes the first stay true. Correcting two strings fixes today; a name check
that covers the branch it is about fixes the next one.

woojubb:
Closes #2036

## The trap this was found in the middle of

R1 exists to prevent one disaster, in its own words:

> Branch protection matches on the context NAME, so a required context nothing publishes never
> reports and blocks the PR forever.

That is a fact about **how branch protection matches** — identical on `develop`. R1 ran over `main`'s
declared list alone (`GOVERNED_BRANCH = 'main'`), and `develop`'s ten contexts were checked by
nothing.

```
declared in deliberately_not_required     actually published
  patch-coverage                            patch-coverage (advisory)
  regression-red-proof                      regression-red-proof (enforcing: accidental-green only)
```

Neither was required, so neither was harmful. **Both were staged to become required**, and promoting
either by moving its existing entry would have demanded a name nothing publishes — stranding every
`develop` pull request permanently. Found while preparing exactly that promotion (INFRA-046, owner
approved). The entry looks like the thing to move; it is not.

## What changed

Both names corrected. `findContextNameFindings` walks **every declared branch** and **both lists** —
`deliberately_not_required` is in scope because that list is where a promotion starts, and checking
only the live one would leave the trap precisely where it was found. Grouped labels
(`build / quality / scans`) are skipped: they are prose, not match targets.

## The new finder was vacuous, and the harness caught it before I did

```
findContextNameFindings(bare)  ->  []
```

Over a root with no `.github` it reported no findings — which reads as *"every declared context name
is published"* when nothing was read at all. The defect this scan exists to catch, in the check added
to catch more of it.

`guard-scope-fail-closed` refused to register it without a classification, the classification was
produced **by running it**, and both finders now throw. Pinned in `MANDATORY_TREE_GUARDS` with that
measurement; 80 guards proven fail-closed by execution.

## And then that fix broke my own tests, in the tier that matters

Three cases called the finders with no `root`, so they read the real repository. The harness test
tier runs from a temporary clone with **no `.github`** — where the finders now throw by design. So
those cases **passed locally and failed in the tier that gates the push.** Pre-push refused, and was
right.

An assertion about *this* repository belongs to the scan, which evaluates it over the real tree on
every `pnpm harness:scan`. The unit suite gets fixtures, like every other case in the file, plus two
new cases for the fail-closed behaviour itself.

## Red-proofed, each mutation asserted to have applied

| mutation | result |
| -------- | ------ |
| restrict the walk to `main` | the develop cases fail |
| skip `deliberately_not_required` | that case fails, alone |
| revert one corrected name in the **real** declaration | the scan names it and exits 1 |

The third ran against the actual repository file, so the check is proven on the shape that motivated
it.

One negative result recorded rather than dressed up: an attempt to reproduce the no-`.github`
environment by hand failed while loading the vitest config. That proved nothing and was not treated
as evidence — pre-push is the authoritative instrument for that question.

## Verification

- 31 tests in `scripts/harness/__tests__/scan-main-required-checks.test.mjs`
- `pnpm harness:scan`: 134 passed
- `pnpm harness:pre-push`: green, 3508 + 1134 tests across the tiers
- All re-run on the **committed** tree after `eslint --fix` touched staged files.

ACTIONABLE FINDINGS: 0

Refs INFRA-055

### DIFF
--- a/.github/required-status-checks.json
+++ b/.github/required-status-checks.json
@@ -101,11 +101,11 @@
       ],
       "deliberately_not_required": [
         {
-          "context": "patch-coverage",
+          "context": "patch-coverage (advisory)",
           "reason": "Advisory by design (HARNESS-041). It runs and prints a verdict on every code PR but never blocks; excluded because a required context must be able to fail, and this one deliberately cannot. INFRA-046 holds it advisory over two open defects: NO-DATA lines are folded into the below-target total, and the React-UI denominator policy is undecided."
         },
         {
-          "context": "regression-red-proof",
+          "context": "regression-red-proof (enforcing: accidental-green only)",
           "reason": "CAN fail as of INFRA-046 — `REGRESSION_RED_PROOF_ENFORCE=1` makes an `accidental-green` verdict exit 1 — so the 'deliberately cannot fail' reason above does NOT apply to it. Excluded for a different reason, by owner decision: `accidental-green` has never fired on a real pull request, and making it required would put an untested refusal in the merge path. One observed firing is the evidence that promotes it. Note this exclusion does not make the job harmless: `merge-gate.sh` refuses any non-CLEAN mergeStateStatus, and GitHub reports UNSTABLE when a NON-required check fails."
         },
         {
--- a/scripts/harness/__tests__/scan-main-required-checks.test.mjs
+++ b/scripts/harness/__tests__/scan-main-required-checks.test.mjs
@@ -1,4 +1,4 @@
-import { mkdirSync, writeFileSync } from 'node:fs';
+import { mkdirSync, mkdtempSync, writeFileSync } from 'node:fs';
 import { mkdtemp } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import path from 'node:path';
@@ -7,7 +7,10 @@ import { describe, expect, it } from 'vitest';
 
 import {
   DECLARATION_FILE,
+  declaredBranches,
+  findContextNameFindings,
   findRequiredCheckFindings,
+  publishedContexts,
   jobConditionProblem,
   jobNeeds,
   pullRequestTrigger,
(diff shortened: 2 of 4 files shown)


woojubb:
## Reviewer verdict

**Base** `8d7a7c3abe9e9c1a86529eb847423d19b7e05dac` · **Head** `91c2e683ce5c179de3e48297a1421fd727d83b04`

CI: **20 passed, 0 failed**, `mergeStateStatus=CLEAN`. Threads: 0.

What I checked, in the order that mattered:

- **The two corrected names match what jobs actually publish**, read from `gh pr checks` on a settled pull request rather than from the workflow source — the published string is what branch protection matches, and that is the thing that was wrong.
- **The new check fires on the real shape.** Reverting one corrected name in the actual declaration file (not a fixture) makes the scan name it and exit 1. Restoring it passes.
- **The new finder fails closed, proven by execution rather than by inspection.** It returned `[]` over a bare root when first written; `guard-scope-fail-closed` refused to register it without a classification, and the classification came from running it.
- **Every mutation asserts it applied.** That line exists because an earlier experiment today reported green from a `.replace` that silently matched nothing.

The part I want a second reader to weigh is the scope decision: `deliberately_not_required` is checked, not just the live list. It is defensible precisely because that list is where a promotion starts — both misnamed entries were sitting there, harmless, waiting to be promoted into a permanent block.

One negative result is recorded in the description rather than dropped: an attempt to reproduce the no-`.github` environment by hand failed on vitest config loading. It proved nothing and was not used as evidence.

ACTIONABLE FINDINGS: 0

PR #2131, opened 2026-08-22
feat(harness): a pull_request_target gate is inert until promotion, so report the lag (INFRA-128)

### LINKED ISSUE #2039: A pull_request_target gate runs the default branch's copy of itself, so fixing it does nothing until promotion

woojubb:
## The gap

A `pull_request_target` workflow loads its YAML from the repository's **default branch** — not from
the pull request, and not from the pull request's base. So a fix to such a gate is inert on the
branch that carries it, and stays inert until that branch is promoted to `main`.

The harness said nothing about this, and the cost was paid twice in one session on issue #1719's
gate: two real defects were found, fixed, reviewed and merged to `develop`, and neither changed
what actually ran. Both were discovered by hand, on a throwaway pull request.

## What this adds

`scans` → `pull-request-target-promotion-lag`. For every `pull_request_target` workflow on `HEAD`,
compare it against the promotion ref and report the verdict per workflow:

- `promoted` — the copies agree, so what runs is what was reviewed;
- `lagging` — promoted once, then diverged;
- `absent` — not on the promotion ref at all, which is a stronger statement than "differs" and is
  kept as its own state rather than folded into the delta.

It reports; it does not refuse. Lag is a normal state between a merge and a promotion, and a check
that failed on it would be red for the whole window it exists to describe.

`scan-workflow-provenance`'s finding now branches on trigger, because the two triggers fail in
**opposite** directions: a `pull_request` gate's edit is too powerful (the change carries the control
plane that judges it), while a `pull_request_target` gate's edit is inert. One sentence could not be
true of both.

## Whole-file comparison, deliberately imprecise

Comparing only the fields that "matter" needs a list of what matters, and that list rots silently —
the day a gate grows a field nobody listed, the check reports a match over a real difference.
Whole-file comparison over-reports instead. Comments and blank lines are stripped, since a reworded
rationale is not a behaviour change; a `#` inside a quoted string would be over-stripped, which is
accepted because that failure direction is a false delta, never a missed one.

## The state at authoring, said out loud

`git diff origin/main origin/develop -- .github/workflows/workflow-provenance-gate.yml` is **empty**
as this lands — the promotion carrying both of issue #1719's gate fixes went through first. So the
check ships with no live subject and prints:

```
::examined:: 1 pull_request_target workflow(s); 1 match origin/main, 0 do not
  every pull_request_target workflow matches origin/main, so what runs is what was reviewed.
  This line is the measurement, not the absence of one.
```

Recorded in the check's own header and in the task record, because a green line from a check with
nothing to find and a green line from a check that found nothing look identical, and only the
second one means anything.

## What the harness caught that I did not

Three of this branch's fixes came from checks rather than from me, and each is worth naming:

- **`guard-scope-fail-closed`** made me run the finder against a root with no governed tree. It
  returned every workflow as `absent`. `readAtRef` collapsed two different facts — "this ref has no
  such file" and "there is no such ref" — into one `null`, so an unfetched `origin/main` would have
  reported every gate as never promoted. That is a loud wrong answer built out of exactly the
  confusion this check exists to name. Now it resolves the ref up front and refuses.
- **`measurement-provenance`** rejected `lag.length` as the reported size. The counter now
  increments inside the walk, and is asserted as an output: an exact value against a five-workflow
  fixture where only three are subjects, again after a second walk, and once more against a tree
  with no subject at all.
- **`fixture-floor` and `temp-dir-owner`** named the test file's location and its temp-directory
  owner. Both were mine to get right and neither was.

The three verdict cases originally ran against a stub that returned a fixed `promoted` for every
fixture. Two of the three passed while proving nothing — the only reason the stub was noticed is
that the third case disagreed with it. They now run against throwaway git repositories with real
refs, because `git show <ref>:<path>` is the behaviour under test and anything standing in for git
is asserting the stand-in.

Closes #2039

Task record: `.agents/tasks/INFRA-128-pull-request-target-promotion-lag.md`

### DIFF
--- a/.agents/tasks/INFRA-128-pull-request-target-promotion-lag.md
+++ b/.agents/tasks/INFRA-128-pull-request-target-promotion-lag.md
@@ -1,0 +1,69 @@
+---
+title: 'INFRA-128: report the promotion lag of every pull_request_target gate'
+issue: https://github.com/woojubb/robota/issues/2039
+status: in-progress
+created: 2026-08-22
+priority: medium
+urgency: soon
+area: scripts/harness, .github/workflows
+depends_on: [INFRA-097]
+---
+
+# INFRA-128: report the promotion lag of every `pull_request_target` gate
+
+## Objective
+
+A `pull_request_target` workflow loads its YAML from the repository's DEFAULT branch, not from the
+PR and not from the PR's base. So a fix to such a gate is **inert on the branch that carries it**,
+and stays inert until that branch is promoted to `main`. Nothing in the harness said so, and the
+cost was paid twice in one session on issue #1719's gate: two defects were found, fixed, reviewed
+and merged to `develop`, and neither changed what actually ran.
+
+Report that gap, per workflow, on every scan run.
+
+## Plan
+
+- [x] Compare each `pull_request_target` workflow between `HEAD` and the promotion ref.
+- [x] Distinguish `absent` (never promoted) from `lagging` (promoted, then diverged).
+- [x] Refuse rather than report "no delta" when a ref cannot be read.
+- [x] Register the scan in `scripts/harness/run-all-scans.mjs`.
+- [x] Split `scan-workflow-provenance`'s finding by trigger, since the two triggers fail in
+      opposite directions.
+
+## Why whole-file comparison, not a curated field list
+
+Comparing only the fields that "matter" needs a list of what matters, and that list rots silently:
+the day a `pull_request_target` gate grows a field nobody listed, the check reports a match over a
+real difference. Whole-file comparison over-reports instead — a reworded comment is not a behaviour
+change, so comments and blank lines are stripped, and what remains can still differ for reasons that
+do not affect execution. That direction is the safe one: the failure mode is one advisory line on a
+diff already touching CI, where someone is already looking.
+
+## The state at authoring, stated so it cannot be mistaken for a measurement of nothing
+
+`git diff origin/main origin/develop -- .github/workflows/workflow-provenance-gate.yml` is **empty**
+as this lands: the promotion that carried both of issue #1719's gate fixes went through first. The check
+therefore ships with **no live subject** — it reports `1 pull_request_target workflow(s); 1 match
+origin/main, 0 do not`. Recorded here because a green line from a check with nothing to find and a
+green line from a check that found nothing are indistinguishable at a glance, and only the first one
+means the check is unproven against a real delta.
+
+The `lagging`, `absent` and unreadable-ref cases are therefore proven against throwaway git
+repositories with real refs, not against this tree.
+
+## The false start, kept because it is the point
+
+The first cut of those three cases used a stub helper that returned a fixed `promoted` verdict for
+every fixture. Two of the three passed. They proved nothing: the assertion could not observe what
+the code under test computed, and one case failed only because the stub disagreed with the fixture
+it served — which is the ONLY reason the stub was noticed at all.
+
+That is the same defect one layer up from this check's subject. A check that cannot fail installed
+in the slot where a check belongs reads exactly like a check that passes. Replaced with temporary
+repositories, which is I/O in a test and worth it: `git show <ref>:<path>` is the behaviour under
+test, so anything that stands in for git is asserting the stand-in.
+
+The refusal case then found a real defect in the implementation. `readAtRef` returned `null` for two
+different facts — "this ref has no such file" and "there is no such ref" — and `findPromotionLagAt`
+read both as `absent`. An unfetched `origin/main` would have reported every workflow as never
+promoted: a loud wrong answer, and one built out of exactly the collapse this check exists to name.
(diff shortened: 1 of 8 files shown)


woojubb:
## Review — PR #2131

Base `7c26ead5a9d752ae0d94120b771b47e904a6bd80` … head `b16f2cac66238c53da6a32c6244cc9e0e0c8da9b`. Six files: one new scan, its fixture test, the runner
registration, two baselines, and the trigger split in `scan-workflow-provenance.mjs`.

### What I verified rather than asserted

- **The fail-closed claim.** Ran `findPromotionLagAt` against a temporary root that is not a git
  repository: it throws `could not resolve the promotion ref` before reading a workflow. This is
  the classification `guard-scope-fail-closed` demands, and running it is what found the defect it
  now records — the first cut returned every workflow as `absent` there.
- **The reported size.** `readExamined()` is incremented inside the walk, not read off the result
  array. Asserted exact (`toBe(3)`) against a five-workflow fixture where two are `pull_request`,
  again after a second walk, and once against a subject-free tree where it must be 0.
- **The empty-subject claim.** `git diff origin/main origin/develop -- .github/workflows/workflow-provenance-gate.yml`
  is empty, so the `1 match, 0 do not` line is real and the check has no live delta to prove itself
  against. Stated in the header, the task record and the PR body rather than left for a reader to
  discover.
- **No `chdir`.** An earlier cut moved the process directory and passed in isolation, then failed
  under `harness:test:contracts` — vitest workers refuse `process.chdir`. The root is a parameter
  now, which is why it was made one.

### On the design

Reporting rather than refusing is right: lag is the normal state between a merge and a promotion, so
a refusing check would be red for the entire window it exists to describe. Whole-file comparison
with comments stripped over-reports rather than under-reports, and the comment says so and names the
one over-strip it accepts (a `#` inside a quoted string).

The trigger split in `scan-workflow-provenance` is the part I looked at hardest, since it changes an
existing finding. The two triggers genuinely fail in opposite directions — `pull_request` loads from
the PR (too powerful), `pull_request_target` from the default branch (inert) — so one sentence could
not have been true of both, and the old one was true of only the first.

`verify-like-ci`: PASS, all 12 stages. 137 scans pass.

ACTIONABLE FINDINGS: 0

PR #2132, opened 2026-08-22
feat(ci): regression-red-proof is required on develop (INFRA-046)

woojubb:
## The block is discharged, and it was discharged by evidence, not by argument

INFRA-046 named its condition exactly, which is what made it dischargeable:

> `accidental-green` has never fired on a real pull request, so requiring it would put an untested
> refusal in the merge path. **One observed firing is what promotes it.**

It fired twice on PR #1983 and caught a real defect — a registration hunk in
`scripts/harness/run-all-scans.mjs` that could be reversed with **every changed test still
passing**. The author repaired it in three rounds without assistance. Three other verdicts in the
same run reported `inconclusive` **with the reason** (*"the range added this file and never revised
it, so there is no earlier state to reverse to … which is not a verdict"*), which is the property
that makes a refusal from this gate worth acting on.

Owner approved the promotion on that evidence.

## Applied in the order the declaration file demands

1. **`.github/required-status-checks.json`** — the entry moves from `deliberately_not_required` into
   `branches.develop.required_status_checks`.
2. **The live `protect-develop` ruleset** — read first, written once, read back:

```
before: 10 contexts        after: 11 contexts
rules / conditions / enforcement / bypass_actors:  identical before and after
scan-main-required-checks.mjs --live:  REAL EXIT 0, "Live ruleset reconciled."
```

The declaration is the source and a ruleset ahead of it is the defect issue #1980 recorded, so the
order is not incidental.

**No `develop` pull request was open at write time**, so nothing inherited a permanently pending
context.

## It moves with the name the job PUBLISHES

`regression-red-proof (enforcing: accidental-green only)`, not the bare `regression-red-proof` it had
been recorded under. Promoting that spelling would have made branch protection match a name nothing
publishes and stranded every `develop` pull request. That trap was found while preparing this change
and closed separately as issue #2036.

## Three refusals from pre-push, all correct

- **The scan was the wrong instrument.** `scripts/harness/ci-mirror-map.mjs` exits 0; the contract
  lives in its **test suite**. I ran the cheaper one and stopped, which is the mistake this
  repository keeps catching.
- **`NOT_MIRRORED` is where the declaration belongs** — not `local.notRunnable` in the JSON, which is
  a different consumer. Declared with the precise reason: the checker *runs* locally and cannot
  reproduce the **verdict**, because an `allow-green-at-base` declared in the PR body alone is
  invisible off a pull request. A mirror that can disagree with the gate on the gate's own escape
  hatch is worse than none.
- **One assertion could not express this context.** It searched the workflow for a bare
  `name: <context>`. This name contains `: `, so **YAML requires quoting it** — the assertion rejected
  the only spelling that can legally exist. It held until now because every context happened to be a
  bare scalar. The property checked is unchanged; the assertion now recognises either form.

I was careful with that last one: earlier today I nearly talked myself into calling a different rule
"over-broad" to license my own change, and was wrong. The test here is different — it asks whether
the job's `name:` equals the matched context, and that property is quoting-independent.

## Red-proofed, mutations asserted to have applied

Naming the unpublished spelling in `NOT_MIRRORED` fails three cases. 51 mirror-map tests, 135 scans,
4645 tests across the pre-push tiers.

## This does NOT close INFRA-046

The item's title names **two** gates. `patch-coverage` is still advisory over two unchanged defects —
`inconclusive-no-data` folded into the BELOW-TARGET total, and an undecided React-UI denominator
policy. Both were re-read rather than assumed, and neither is a measurement: they are decisions about
what the number should mean. The record moves `blocked` → `in-progress` and states the terminal
condition rather than claiming half a subject as a whole.

ACTIONABLE FINDINGS: 0

Refs INFRA-046

### DIFF
--- a/.agents/tasks/INFRA-046-promote-advisory-gates.md
+++ b/.agents/tasks/INFRA-046-promote-advisory-gates.md
@@ -1,6 +1,6 @@
 ---
 title: 'INFRA-046: promote advisory CI gates (regression-red-proof, patch-coverage) to blocking'
-status: blocked
+status: in-progress
 created: 2026-07-25
 priority: high
 urgency: now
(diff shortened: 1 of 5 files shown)


woojubb:
## Reviewer verdict

**Base** `c57a4a2c42466cdbc3c2fe05230c4e44a1c9a553` · **Head** `40ee33de0baa256546235b476cf7e1d8dbe52e17`

CI: **20 passed, 0 failed**, `mergeStateStatus=CLEAN`.

**The newly-required context passed on the pull request that requires it:**

```
regression-red-proof (enforcing: accidental-green only)   pass
```

That is worth stating plainly rather than treating as a formality — it establishes the context reports on an ordinary diff under its new status, and **nothing more**. Its ability to FAIL is established separately and by observation: it fired twice on PR #1983 against a real defect. Both halves matter, and neither substitutes for the other. A context that has only ever reported success is indistinguishable from one that cannot fail; this one has done both, on real pull requests.

What I checked on the diff:

- **Order.** Declaration first, live ruleset second. A ruleset ahead of its declaration is the defect issue #1980 recorded, so the sequence is part of the change, not incidental to it.
- **The write was read back**, and rules, conditions, enforcement and bypass_actors are identical before and after — only the context count moves, 10 to 11.
- **No `develop` pull request was open at write time**, so nothing inherited a permanently pending context.
- **The published name, not the recorded one.** The bare spelling matches nothing; promoting it would have stranded every `develop` pull request. Closed separately as issue #2036.
- **`NOT_MIRRORED` carries a reason that names the exact limitation** — the checker runs locally but cannot reproduce the verdict, because a PR-body-only opt-out is invisible off a pull request.

The record is **not** marked done, and that is deliberate: the item's title names two gates and `patch-coverage` is still advisory over two unchanged defects, both of which are decisions rather than measurements.

ACTIONABLE FINDINGS: 0

PR #2134, opened 2026-08-22
fix(harness): patch-coverage stops charging a package that owns no tests (INFRA-046)

woojubb:
## The blocker I had misclassified

INFRA-046 listed three blockers. I described the two remaining as *decisions*. **That was wrong about
one of them**, and the record said so in its own words:

> `patch-coverage` must not charge a PR for a package that has no test suite … **This is the issue
> #1344 defect and it is in `scripts/harness/check-patch-coverage.mjs`.**

A file, a named defect, and a described fix. Correcting the classification is what made it work
rather than a decision to wait for.

## The defect

`coverage.all: true` instruments a package's source whether or not anything exercises it, so a
package with **no test suite** still emits an lcov report — every record zero-hit. Those lines
arrived as MEASURED-and-uncovered, dragged the patch percentage BELOW-TARGET, and charged a pull
request for a package with no way to discharge the debt.

One confirmed false positive is one more than this item's promotion criterion allows.

## The fix, and all three directions pinned

| package | lcov | verdict |
| ------- | ---- | ------- |
| owns tests | nothing covered | `patch-coverage-below-target` — the defect this gate exists for |
| **no tests** | **nothing covered** | **`inconclusive-no-data`** — issue #1344's false positive |
| no tests | any line covered | `patch-coverage-below-target` — a hit is evidence something ran |

Both conditions are required. A package *with* tests whose report is wholly uncovered is precisely
what this gate is for, and must still fail.

`lcovIsEntirelyUnexercised` returns **false** for an empty report, deliberately: no records is not
the same as no hits, and the existing UNINSTRUMENTED path already owns that case. Pinned by its own
test.

## Two existing red cases broke, and were right to

Neither modelled whether the package owns tests — the dimension did not exist. The red fixture is
`LH:0` **with no test file**, which is indistinguishable from the very false positive being fixed.

That is the question I put to myself before touching them: *is the fixture describing a case that
should fail?* Per issue #1344's own definition, a testless package being charged is a false positive
— so the fixture was under-specified, not the change wrong. Fixtures now declare their test files
through a `tests.txt` manifest, and both say so explicitly.

Verified end to end, not just in units:

```
red fixture                        -> patch-coverage-below-target
red fixture minus its tests.txt    -> inconclusive-no-data
```

## What this does NOT do, measured rather than assumed

The obvious hope was that this also discharges the React-UI coverage tax (blocker 2), since a GUI
package with no component tests would now be excused. **It does not:**

```
packages/agent-cli-web       tsx=1     test files=0    -> excused
apps/agent-web               tsx=6     test files=1    -> owns tests, still charged
apps/agent-app               tsx=3     test files=1    -> owns tests, still charged
packages/agent-playground    tsx=184   test files=32   -> owns tests, still charged
```

Three of four still pay, and correctly so under the gate's own logic. Blocker 2 remains a genuine
decision — exclude `.tsx` render surfaces from the denominator, or stand up component-test
infrastructure — and nothing about the number settles it.

## Verification

19 tests in `scripts/harness/__tests__/check-patch-coverage.test.mjs`, 135 scans, 4668 tests across
the pre-push tiers.

**This does not close INFRA-046.** Blocker 3 is discharged (`regression-red-proof` is promoted and
required) and blocker 1 is fixed here; blocker 2 is a decision. The record states that and stays
`in-progress`.

ACTIONABLE FINDINGS: 0

Refs INFRA-046

### DIFF
--- a/.agents/tasks/INFRA-046-promote-advisory-gates.md
+++ b/.agents/tasks/INFRA-046-promote-advisory-gates.md
@@ -456,3 +456,55 @@ evidence — is gone, and what remains is a decision plus the work it implies. *
 `patch-coverage` is either promoted on the same standard of evidence, or recorded as permanently
 advisory with the reason, in which case this item's title is wrong and should be narrowed rather than
 the record being marked complete over half its subject.**
+
+## 2026-08-22 (later) — blocker 1 is FIXED; blocker 2 is measured and still a decision
+
+**Blocker 1 was a code defect, not a decision, and I had said otherwise.** The record names it
+plainly — _"This is the issue #1344 defect and it is in `scripts/harness/check-patch-coverage.mjs`"_ —
+and an earlier entry of mine described both remaining blockers as decisions. That was wrong about
+this one, and the correction is the reason it is now fixed rather than deferred.
+
+A package with no test suite still emits an lcov report under `coverage.all: true`, every record
+zero-hit. Those lines were counted as MEASURED-and-uncovered and dragged the patch percentage
+BELOW-TARGET, charging a pull request for a package with no way to discharge the debt.
+
+Such a package is now NO-DATA (INCONCLUSIVE, advisory). Both conditions are required, and all three
+directions are pinned by tests and by an end-to-end fixture run:
+
+| package    | lcov             | verdict                                                         |
+| ---------- | ---------------- | --------------------------------------------------------------- |
+| owns tests | nothing covered  | `patch-coverage-below-target` — the defect this gate exists for |
+| no tests   | nothing covered  | `inconclusive-no-data` — issue #1344's false positive           |
+| no tests   | any line covered | `patch-coverage-below-target` — a hit is evidence something ran |
+
+The two existing red cases failed on this change and were right to: neither modelled whether the
+package owns tests, because that dimension did not exist. The red fixture is `LH:0` with no test
+file — **indistinguishable from the false positive being fixed**. Fixtures now declare their test
+files explicitly.
+
+**Blocker 3 is discharged** — `regression-red-proof` got its real verdicts and is promoted and
+required.
+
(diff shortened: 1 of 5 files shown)


woojubb:
## Reviewer verdict

**Base** `cf771ce7934c79aa5098fd829fce5fa6ef86621d` · **Head** `660d7bf0a97c5bdb095d04420061f0c4cc14b6d7`

CI: **20 passed, 0 failed**, `mergeStateStatus=CLEAN`.

The thing worth reviewing here is not the fix but whether the fix blunts the gate. It does not, and that is established three ways rather than argued:

- **Unit**, all three directions: owns-tests + uncovered still BELOW-TARGET; no-tests + uncovered now INCONCLUSIVE; no-tests + any hit still BELOW-TARGET.
- **End to end on the real fixture**: the red fixture still exits `patch-coverage-below-target`, and the same fixture with only its `tests.txt` removed reports `inconclusive-no-data`. One file is the entire difference, which is exactly the discrimination being claimed.
- **`lcovIsEntirelyUnexercised` is false for an EMPTY report**, pinned by its own case — no records is not the same as no hits, and swallowing that under a new label would have hidden the existing UNINSTRUMENTED path.

**Two pre-existing red cases failed on this change.** I checked before touching them whether the fixture described a case that SHOULD fail: per issue #1344's own definition a testless package being charged is a false positive, so the fixture was under-specified rather than the change wrong. That is the discriminator between fixing a fixture and bending a test to fit.

**And the limit is measured, not assumed.** The natural hope — that this also discharges the React-UI tax — is false: three of four GUI packages own some tests and are still charged, correctly. Blocker 2 stays a decision, and the description says so rather than letting the fix imply more than it does.

This does **not** close INFRA-046.

ACTIONABLE FINDINGS: 0

PR #2136, opened 2026-08-22
feat(harness): INFRA-046 and INFRA-054 are both decided and closed

woojubb:
Two owner decisions, implemented and recorded. Both items close.

## INFRA-054 — DECIDED: the direct-write paths stay

Fast-forward promotion is **not** adopted. The three paths it would have to remove are deliberate
safety devices:

| path | why it stays |
| ---- | ------------ |
| `protect-main.bypass_actors` | the recovery route when `main` is wedged |
| `main-pr-source-guard` admitting `hotfix/*` | the incident path that does not wait for `develop` to be releasable |
| Dependabot disabled by POLICY | a policy is reversible by a decision; a mechanism would be rebuilt |

The cost accepted instead is INFRA-051's: `main` accumulates promotion merges and each cycle
re-records ancestry. **Two promotions today exercised that path and asked nothing of a human.**

One precondition genuinely moved and does not change the decision — non-merge commits on `main`
unseen by `develop` went **10 → 0**, absorbed by ordinary promotion. Recorded so the zero is not
later read as evidence the paths were closed: the measurement is about what *has* landed, the
condition was about what *can*.

`unearned-done-claims` refused the archive over three unticked criteria and was right to — they
describe the implementation that was not adopted. Each now carries `allow-unmet-criterion` naming
that reason, **on the line the criterion starts on**; a continuation line does not count, which the
scan caught on my first attempt.

## INFRA-046 — DECIDED: render surfaces leave the patch denominator

`.tsx`/`.jsx` are excluded, over the alternative of standing up component-test infrastructure.

**An exclusion from the DENOMINATOR, not a pass.** Their lines are neither measured nor counted as
covered, so the percentage describes only the code this gate can speak about. A render-surface-only
diff reports `SKIPPED` — *"nothing to measure here"* — rather than `OK`, which would claim it
measured and was satisfied.

Four cases pin it, and all four fail if the exclusion is reverted:

- `.tsx`/`.jsx` out, `.ts`/`.js` in;
- a render-only diff is SKIPPED with `measured = 0`;
- a `.ts` file in the same diff still measured and still failing — `measured = 2`, not 4;
- the **pre-existing** classification case, which asserted `.tsx` *was* coverable. The mutation run
  is what showed that case had to move to the decided behaviour rather than be left contradicting it.

### All three of this item's blockers, and what each turned out to be

| # | blocker | nature | outcome |
| - | ------- | ------ | ------- |
| 1 | a package with no test suite was charged (issue #1344) | **code defect** | fixed today |
| 2 | React-UI denominator policy (issue #1348 class) | **decision** | made today |
| 3 | `regression-red-proof` had no real verdicts | **evidence** | arrived today |

I had described blocker 1 as a decision. It was not — the record named the file. Correcting that
classification is what turned it into work instead of a wait.

Blocker 3's own words were *"one observed firing is what promotes it"*: a condition written
precisely enough to be discharged by an event rather than by an argument, and that is exactly how it
ended.

## What is NOT claimed

**`patch-coverage` remains advisory.** Both defects that held it there are cleared, so enforcing it
is now an unobstructed decision rather than a blocked one — and nobody has taken it. The record says
so instead of letting "INFRA-046 is done" imply otherwise. `regression-red-proof` is the gate this
item promoted; `patch-coverage` is the gate it made promotable.

## Verification

22 tests in `scripts/harness/__tests__/check-patch-coverage.test.mjs`, 136 scans, 4671 tests across
the pre-push tiers.

ACTIONABLE FINDINGS: 0

Refs INFRA-046, INFRA-054

### DIFF
--- a/.agents/tasks/completed/INFRA-046-promote-advisory-gates.md
+++ b/.agents/tasks/completed/INFRA-046-promote-advisory-gates.md
@@ -1,7 +1,8 @@
 ---
 title: 'INFRA-046: promote advisory CI gates (regression-red-proof, patch-coverage) to blocking'
-status: in-progress
+status: done
 created: 2026-07-25
+completed: 2026-08-22
 priority: high
 urgency: now
 area: .github/workflows/ci.yml, repo rulesets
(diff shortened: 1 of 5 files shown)


woojubb:
## Reviewer verdict

**Base** `a7ea988b875df55e5d78ce876e751d3b630aa153` · **Head** `34060545040f12899bfd1d286d9a67c9a6fc6b93`

CI: **20 passed, 0 failed**, `mergeStateStatus=CLEAN`. Threads: 0.

Two items closing on owner decisions, so the review question is whether each record earns its `done` rather than whether the code is clever.

**INFRA-054** closes as a DECISION and says so. `unearned-done-claims` refused the archive over three unticked criteria and the refusal was correct — those criteria describe the implementation that was not adopted. Each now names that reason on the line the criterion starts on. I got that wrong first (a continuation line does not count) and the scan caught it, which is the check working.

**INFRA-046** closes with all three blockers cleared by three different mechanisms, and the record distinguishes them: a code defect, a decision, and an evidence condition discharged by an event. That distinction is the useful part — I had misclassified the code defect as a decision, and correcting the classification is what turned it into work.

What I checked on the code:

- **The exclusion is from the denominator, not a pass.** A render-only diff reports `SKIPPED`, not `OK`. The difference is a claim about what was measured, and the case asserts the constant rather than a string literal so a rename cannot silently weaken it.
- **A `.ts` file in the same diff still fails, with `measured = 2` and not 4** — the number is the assertion, not the verdict alone.
- **A pre-existing case asserted `.tsx` WAS coverable** and now asserts the opposite. That is the one worth a second reader's attention: it was changed because the behaviour was decided, and the mutation run is what proved it had to move rather than my say-so.

**Not claimed:** `patch-coverage` stays advisory. Enforcing it is now an unobstructed decision nobody has taken, and the record says that instead of letting "INFRA-046 is done" imply enforcement.

ACTIONABLE FINDINGS: 0

Issue #2113, opened 2026-08-22
P1 topology leaf: narrow transport contracts and delete the omnibus package facade


woojubb:
Parent tracker: #2068
Blocked by: #2108, #2109, #2110, #2111, #2112

## Outcome

Leave only adapter/config/channel/admission contracts in the transport owner and remove the old omnibus root facade.

## In scope

- remaining transport contract package
- all old-root import cleanup
- ownership/dependency mechanical guards

## Acceptance criteria

- Executor/session/command/analytics/mobility consumers no longer import the omnibus package.
- The root barrel/facade is removed without compatibility forwarding.
- Owner-family and dependency-direction guards pass.

## Scope boundary

This is one independently mergeable leaf. New concerns become sibling issues under #2068; they do not expand this issue. The audited API is prerelease, so do not add compatibility shims or forwarding aliases.

Issue #2110, opened 2026-08-22
P1 topology leaf: move session, interaction, event, and persistence contracts to runtime owners


woojubb:
Parent tracker: #2068
Blocked by: #2080

## Outcome

Move the session/runtime contract families to the owners selected by the topology map.

## In scope

- interactive session/capability/event/driver contracts
- session persistence/goal/plan contracts
- framework/session/transport consumer imports

## Acceptance criteria

- Each family lands in exactly the owner named by #2080.
- Transport adapters consume the runtime contracts inward.
- Mobility and analytics families are out of scope.

## Scope boundary

This is one independently mergeable leaf. New concerns become sibling issues under #2068; they do not expand this issue. The audited API is prerelease, so do not add compatibility shims or forwarding aliases.

Issue #2109, opened 2026-08-22
P1 topology leaf: move background, subagent, and workspace contracts to the execution owner


woojubb:
Parent tracker: #2068
Blocked by: #2080

## Outcome

Move the execution-bounded contract families out of the transport-named hub.

## In scope

- background tasks/job groups
- subagent and execution workspace data contracts
- executor/framework direct imports

## Acceptance criteria

- Execution owners no longer import their own contracts from transport.
- No session/command/mobility contracts move in this issue.
- No compatibility re-export remains.

## Scope boundary

This is one independently mergeable leaf. New concerns become sibling issues under #2068; they do not expand this issue. The audited API is prerelease, so do not add compatibility shims or forwarding aliases.

Issue #2108, opened 2026-08-22
P1 topology leaf: move command contracts to their domain owner


woojubb:
Parent tracker: #2068
Blocked by: #2080

## Outcome

Move only the command descriptor/result/host-action contract family out of `agent-interface-transport` to the owner selected by #2080.

## In scope

- command contract source/export/package metadata
- direct consumer import migration
- focused dependency/build tests

## Acceptance criteria

- The command family has one owner and no forwarding export from the old package.
- No background/session/transport family is moved here.
- The package graph remains within #2080's target edges.

## Scope boundary

This is one independently mergeable leaf. New concerns become sibling issues under #2068; they do not expand this issue. The audited API is prerelease, so do not add compatibility shims or forwarding aliases.

PR #2199, opened 2026-08-23
feat(harness): interface packages compose downward across declared layers (ARCH-101)

### LINKED ISSUE #2180: P1 blocker: INTERFACE-DEPS forbids the contract-package decomposition it is supposed to permit
Blocks #2109, and by the same mechanism every other leaf under #2068 (#2108, #2110, #2111, #2112, #2113).

Found while starting #2109. Reporting rather than working around it, and **not** amending the rule
inside a topology leaf: this is repository policy.

## The contradiction

`.agents/project-structure.md` § Interface Package Rule:

> An `agent-interface-*` package's internal dependencies are a subset of `{agent-core}` — contracts
> never depend on implementation packages (INFRA-025; mechanized as the `INTERFACE-DEPS` rule in the
> `deps` scan).

`scripts/harness/check-dependency-direction.mjs` → `checkInterfacePackageDeps` enforces the literal
reading: any `@robota-sdk/*` production dependency other than `agent-core` is a violation.

The owner map merged by #2080 (`.agents/specs/contract-family-owner-map.md`) requires
interface→interface edges. They are not optional and no migration order avoids them.

## Why no ordering avoids it

`session-contracts` and `session-capability-contracts` both import all four families #2109 moves:

```
session-contracts             -> background-group-contracts, background-task-contracts, workspace-contracts
session-capability-contracts  -> background-group-contracts, background-task-contracts,
                                 subagent-contracts, workspace-contracts
```

- Move execution first (#2109, current plan): those two modules stay behind in
  `agent-interface-transport`, so `agent-interface-transport → agent-interface-execution` is created.
  **INTERFACE-DEPS violation.**
- Move session first: `agent-interface-session → agent-interface-execution`. Same violation, and it
  also creates the temporary cycle #2080 forbids.
- Move everything at once: the six owners still depend on each other. Same violation.

The target graph is acyclic — `interface-family-owner` proves it — but acyclic is not the same as
edgeless, and INTERFACE-DEPS as written permits only an edgeless interface layer.

## The rule's intent versus its text

The stated intent is the second half of its own sentence: *"contracts never depend on implementation
packages."* An `agent-interface-*` → `agent-interface-*` edge does not violate that intent — both
sides are contracts, and the direction it guards (implementation → contract) is untouched.

The text achieves the intent by a proxy that was sufficient while ONE interface package held every
family. `agent-interface-tui` is the only other one and it needs no peer. Decomposing into six owners
makes the proxy wrong.

## A defect in #2080 that this exposes, stated plainly

The merged ARCH-100 spec-doc asserts, under New-surface placement:

> Every new package's internal dependency set stays a subset of `{agent-core}` plus peer
> `agent-interface-*` packages, **exactly as INFRA-025 already requires**.

That is false. INFRA-025 requires `⊆ {agent-core}`, with no peer clause. The spec asserted the rule
already permitted what it forbids, and the claim passed review because it reads like a restatement of
the rule rather than an amendment to it. The owner map itself is unaffected — the placements and the
acyclicity proof stand — but the compatibility claim was wrong, and this issue is the correction.

## What needs deciding

Amend INTERFACE-DEPS so a contract package may depend on a peer contract package, while implementation
dependencies stay forbidden. Concretely, `checkInterfacePackageDeps` would permit a dep that is
`agent-core` **or** matches the `agent-interface-` prefix, and reject everything else.

Two properties the amendment should keep, since both are load-bearing today:

- the implementation→contract direction stays enforced (nothing about that changes);
- the interface layer stays **acyclic**, which is not currently checked by `deps` at all. The
  `interface-family-owner` scan (#2080) proves acyclicity for the modules it maps; a peer-dependency
  allowance should be paired with a cycle check over real `agent-interface-*` manifests, or the
  amendment trades one guarantee for none.

## Why this is not decided by an agent

`.agents/rules/` is repository policy, and the standing delegation for this issue programme excludes
repository-wide policy files. Filing for the owner rather than amending it inside #2109.

## Impact

#2109 cannot land while this stands. #2108/#2112 hit it the moment they create their owner package.
The `deps` scan will go red on the first leaf that adds the manifest edge, so this fails closed rather
than silently — the decomposition stops, it does not corrupt.

woojubb:
Closes #2180. Unblocks #2108 through #2113 under tracker #2068.

## What was blocking

The Interface Package Rule required an **edgeless** interface layer — `deps ⊆ {agent-core}`. That was
an exact proxy for the property while one package held every contract family. ARCH-100 decomposed it
into six owners and the proxy became an over-approximation that forbids the target graph.

The owner ruled that the general layer rule governs this prefix: composition across **differing**
layers, one-directional, is permitted; only **same-layer** is forbidden, and upward is forbidden by
the same sentence.

**Nothing implemented the ruling.** `checkInterfacePackageDeps` still failed every
interface→interface edge, so #2109 would have gone red on its first push — `agent-interface-transport`
retains `session-contracts` until #2110 and must depend on `agent-interface-execution`.

## Why this is one PR and not two

Relaxing the manifest prohibition alone would be **worse than the prohibition it replaces**.

Today a same-layer edge is refused only as a side effect of refusing *every* edge, and
`interface-family-owner` proves the module graph is **acyclic** — which no longer implies legal.
`command → execution` is perfectly acyclic and forbidden.

That is fixed permanently by a test that asserts both halves on the same fixture:

> `the same-layer graph is ACYCLIC — findCycles reports nothing`
> `and it is still ILLEGAL — the layer check refuses what acyclicity permits`

Relax without the direction check and the case the ruling exists to forbid becomes reachable and
unguarded at the module level.

## One fact, one parser, two altitudes

The layer declaration is machine-readable in `.agents/specs/contract-family-owner-map.md`, beside the
owner map it is the second axis of. `scripts/harness/interface-layers.mjs` is the **only** thing that
parses it:

- `check-dependency-direction.mjs` judges **manifest** edges (`package.json` dependencies)
- `interface-family-owner` judges **module** edges (imports between contract files)

Both call the same `judgeEdge`. Two guards parsing one markdown table would be two parsers that can
disagree about one fact — the duplication class ARCH-100's map was built to avoid. The document
coupling is accepted once, in one place.

**An undeclared package is not legal by default.** An edge touching one is refused, naming what is
missing, rather than passing because nothing was found.

## Falsified before shipping

Each case demonstrated red against the guard, not assumed:

| case | result |
| --- | --- |
| downward (`session → execution`) | accepted |
| downward across two layers (`mobility → execution`) | accepted |
| **same-layer** (`command → execution`) | refused, both layers named |
| **upward** (`execution → session`) | refused, direction named |
| **undeclared** peer | refused, missing package named |
| implementation dep (`session → agent-framework`) | refused |
| `agent-core` | accepted |

And at module level, by editing the real declaration to put session at layer 0: three `LAYER:`
findings appeared while **`ACYCLICITY:` reported nothing** — the proof that the new condition adds a
property the old one did not have.

## The rule amendment

Six lines out, six lines in. `.agents/project-structure.md` stays at its frozen **385** —
`routing-document-size` is satisfied by construction rather than by an exemption. `rule-statement-floor`
(HARNESS-117) confirms the amended rule still states `INTERFACE-DEPS`.

## A third constraint, checked and deliberately not implemented

The owner also ruled that a family root may not depend on its own members, only the reverse, and that
an aggregator over a family carries a completely different prefix.

I checked whether it falls out of the layer declaration for free. **It does not in general** — a root
and its members can sit at different layers and the edge is still forbidden in one direction, so it is
a directional rule about a pair rather than a layer rule.

For this family it is **vacuous by construction**: there is no `agent-interface` root and no
aggregator, and the naming half of the rule forecloses ever adding one inside the prefix. Recorded in
the map so a later reader does not add a bundle under either name. No guard checks it, deliberately —
a check that can never fire is its own defect.

## Scope

No production TypeScript is moved. `.agents/project-structure.md` and the auditor definitions are not
in conflict with ARCH-102, which merged as `fe3ed577f` before this branch was cut.

## Verification at `e4161906b`

- `pnpm harness:verify-like-ci` — PASS, all 12 stages
- `pnpm harness:scan` — 140 passed, 2 skipped, 0 failed
- 43 unit tests across the two suites

`typecheck` failed on the first two runs and was **not this branch** — it touches zero `.ts` files.
`#2193` (SEC-015) landed on `agent-session` while my local `dist/` predated it. `pnpm build`, then
green. Second time today; the `dist` scan's advisory says exactly this, and it is worth reading before
treating a cross-package type error as your own.

### DIFF
--- a/.agents/project-structure.md
+++ b/.agents/project-structure.md
@@ -316,12 +316,12 @@ doubles→owner /testing`). Pre-existing mechanisms are frozen per package in
   `scripts/harness/interface-entry-baseline.json` and the count may only shrink. The entry edge
   exists because the source edge alone measured something narrower than this rule's words, so a
   100-line prototype-walking forwarder sat outside the rule and inside the green.
-- An `agent-interface-*` package's internal dependencies are a subset of `{agent-core}` —
-  contracts never depend on implementation packages (INFRA-025; mechanized as the
-  `INTERFACE-DEPS` rule in the `deps` scan). `agent-interface-transport` owns the
-  background-task/subagent/compaction data contracts and, post-DATA-001, the
-  session/workspace/command/event/usage contract families; `agent-executor`/`agent-session` import
-  them and keep only runtime SPI.
+- An `agent-interface-*` package depends on `{agent-core}`, and on a PEER `agent-interface-*` package
+  only DOWNWARD across the declared layers, one-directionally. Same-layer and upward are refused, as
+  is any implementation package — contracts never depend on implementations (INFRA-025; ARCH-101;
+  mechanized as the `INTERFACE-DEPS` rule in the `deps` scan). The layer that authorizes an edge is
+  declared once in [`specs/contract-family-owner-map.md`](specs/contract-family-owner-map.md) and read
+  by both guards through `scripts/harness/interface-layers.mjs`. Do not restate a layer elsewhere.
 - Implementation packages (`agent-transport` with subpath `/headless`; the per-concern `agent-transport-tui` / `-ws` / `-http` / `-mcp` packages; `agent-provider` with subpaths `/anthropic`, `/openai`, etc.; `agent-command`) depend on the corresponding `agent-interface-*` package, not on `agent-framework`, for interface types. The transport-facing contract types (command, interaction, event, workspace, session, and transport contracts) live in `agent-interface-transport` as their SSOT (per INFRA-010). This is **mechanically enforced** by `scripts/harness/check-interface-imports.mjs` (wired into `pnpm harness:scan` as the `interface-imports` scan): any implementation package that imports an `agent-interface-transport`-exported symbol from `@robota-sdk/agent-framework` fails the gate. Runtime values and framework-owned types (e.g. `TInteractiveSessionOptions`, `ICommandHostContext`, `ICommandModule`, `TSettingsData`) still come from `agent-framework`.
 - `agent-framework` depends on the `agent-interface-transport` package to consume the contracts it needs (it does not depend on `agent-interface-tui`, which only `agent-transport-tui` consumes).
 - Do not place interface packages in `agent-core` — `agent-core` is zero-deps and owns foundational primitives only.
--- a/.agents/spec-docs/todo/ARCH-101-interface-packages-compose-downward-across-declared-layers.md
+++ b/.agents/spec-docs/todo/ARCH-101-interface-packages-compose-downward-across-declared-layers.md
@@ -1,0 +1,215 @@
+---
+status: approved
+type: RULE
+tags: [typescript]
+---
+
+# ARCH-101: interface packages compose downward across declared layers
+
+Registered as GitHub issue https://github.com/woojubb/robota/issues/2180.
+Unblocks issues #2108–#2113 under tracker issue #2068.
+
+## Problem
+
+`.agents/project-structure.md` § Interface Package Rule restricts an `agent-interface-*` package's
+internal dependencies to a subset of `{agent-core}`. `checkInterfacePackageDeps`
+(`scripts/harness/check-dependency-direction.mjs:232-246`) enforces the literal reading: any
+`@robota-sdk/*` production dependency other than `agent-core` is an `[INTERFACE-DEPS]` violation.
+
+The contract-family owner map merged by ARCH-100 requires interface→interface edges. The owner has
+since ruled that the general layer rule governs this prefix too — composition across **differing**
+layers, one-directional, is permitted; only **same-layer** dependencies are forbidden. So the target
+graph is legal.
+
+**Concrete symptom, and it is what blocks work today.** Nothing implements the ruling. Issue #2109
+creates `agent-interface-execution`; `agent-interface-transport` retains `session-contracts` and
+`session-capability-contracts` until issue #2110, and those name 20 execution symbols. So transport's
+manifest must depend on `agent-interface-execution`, and `pnpm harness:scan` fails with
+
+```
+[INTERFACE-DEPS] Interface-package violation: @robota-sdk/agent-interface-transport must not
+depend on @robota-sdk/agent-interface-execution … (deps ⊆ {@robota-sdk/agent-core}).
+```
+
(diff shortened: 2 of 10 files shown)


woojubb:
## `regression-red-proof` caught an accidental green in this change — fixed in `93ec4e2b4`

Recording it here rather than only in the code, because the second half is the alarming one.

### What the floor found

`accidental-green-fail (all-pass)` on `scan-interface-family-owner.mjs`: **reversing the fix left every
test green.**

The LAYER condition lived inline in `main()`. The tests exercised `judgeEdge` — the shared predicate —
directly and thoroughly, and never tested that the scan *consults* it. Predicate covered, the scan's
use of it not. Deleting the whole condition broke nothing.

> **A guard reachable only through `main()` is a guard no test can falsify.**

That is a property of the code's shape, not a lapse in diligence — no amount of additional testing at
that altitude would have found it. Fixed by extracting an exported pure `findLayerViolations(edges,
layers)` and testing it directly. Proved rather than asserted: neutering it turns **4 tests red**;
restoring turns them green.

### The more alarming half

The same change amended a second guard, `checkInterfacePackageDeps`, to judge layer edges by the
declaration. **Its test file had no layer cases at all.**

The falsification table in this PR's description — downward accepted, same-layer refused, upward
refused, undeclared refused, implementation-dep refused, `agent-core` allowed — was produced by hand
with `node -e` during development. Every case was really run and really observed. **None of it was
ever encoded**, so none of it could fail later.

And the floor could not have told me: that guard's own `regression-red-proof` entry came back
`inconclusive`, because its hunks had no clean prior state to reverse into. The one mechanism that
would have flagged the first gap was structurally unable to flag the second.

Hand-falsification that never becomes a test is the same class as a verification command whose failure
looks like success: the work was done, the evidence was real, and it left no artifact that can fail
later. **Seven cases are now permanent** in `check-interface-package-deps.test.mjs`.

### Verification at `93ec4e2b4`

- `pnpm harness:verify-like-ci` — PASS, all 12 stages
- `pnpm harness:scan` — 141 passed, 1 skipped, 0 failed
- 45 tests across the two affected suites

Recorded as a comment rather than a description edit deliberately: editing the body re-fires the
workflows and cancels the in-flight run, and five checks were still running when this was written.

Issue #2068, opened 2026-08-22
[TRACKING] P1 architecture: decompose agent-interface-transport into domain-owned contract packages


woojubb:
> [!IMPORTANT]
> This is a tracking issue only. Implementation PRs must target one leaf below. Do not expand a leaf; file a sibling under this tracker.

## Execution leaves

- [ ] #2080 — P1 architecture leaf: define the contract-family owner map and acyclic target graph
- [ ] #2108 — P1 topology leaf: move command contracts to their domain owner
- [ ] #2109 — P1 topology leaf: move background, subagent, and workspace contracts to the execution owner
- [ ] #2110 — P1 topology leaf: move session, interaction, event, and persistence contracts to runtime owners
- [ ] #2111 — P1 topology leaf: move peer and handoff contracts to the session-mobility owner
- [ ] #2112 — P2 topology leaf: move analytics and usage contracts to their reporting owner
- [ ] #2113 — P1 topology leaf: narrow transport contracts and delete the omnibus package facade

GitHub sub-issue and blocked-by relationships are authoritative for order. The final removal/guard leaf must close last; no compatibility shim or forwarding facade is part of the target state.

## Closure rule

Close this tracker only when every sub-issue is closed and the final leaf proves the obsolete contract/package surface is gone. A partially migrated state does not satisfy this tracker.

## Summary

`agent-interface-transport` is no longer a transport contract package. It is an omnibus public-contract hub for commands, sessions, persistence, workspace, background execution, subagents, analytics/usage, admission, peers, and handoff. Package ownership follows a dependency-rule convenience instead of bounded contexts and change axes.

The current repository rule explicitly blesses this shape. This issue therefore proposes amending that rule and migrating directly; an unpublished interface does not justify an umbrella compatibility facade.

## Evidence

- [The package description still identifies transport adapters/configuration](https://github.com/woojubb/robota/blob/15800bf0f0023406cf06981a7afcc2cfdcd17f63/packages/agent-interface-transport/package.json#L2-L5), while its root exports roughly 250 declarations across unrelated families.
- [The repository rule explicitly assigns session/workspace/command/event/usage/background/subagent families to this package](https://github.com/woojubb/robota/blob/15800bf0f0023406cf06981a7afcc2cfdcd17f63/.agents/project-structure.md#L297-L325). That is the architectural decision to amend, not a baseline to preserve.
[...]

PR #2214, opened 2026-08-23
feat(interface): extract analytics contracts to their reporting owner (ARCH-105)

### LINKED ISSUE #2112: P2 topology leaf: move analytics and usage contracts to their reporting owner
Parent tracker: #2068
Blocked by: #2080

## Outcome

Move analytics/report/usage types that are not transport contracts to their actual reporting owner.

## In scope

- analytics report and usage contract families
- session analytics and direct consumers
- exports/SPEC ownership

## Acceptance criteria

- Analytics changes no longer version the transport contract package.
- Only analytics/usage symbols move.
- No forwarding export remains.

## Scope boundary

This is one independently mergeable leaf. New concerns become sibling issues under #2068; they do not expand this issue. The audited API is prerelease, so do not add compatibility shims or forwarding aliases.

woojubb:
Closes #2112. Parent tracker: #2068. Owner map: ARCH-100. Layer rule: ARCH-101.

**Last wave-1 leaf**, after #2203 (execution) and #2209 (command). All three domain families are out
of the omnibus.

## A split, not a move

The seven analytics declarations lived **inside** `session-contracts.ts`. There was no file to
`git mv` — which is why the owner map records this family as `symbols@session-contracts`, and why a
file-level plan for this leaf would have done **nothing at all**.

`@robota-sdk/agent-interface-analytics` — 7 symbols, **empty dependency set**. Every field is a
primitive or another member of the set, so it needs nothing, not even `agent-core`. The build graph
confirms it: the package lands at **tier 0 beside `agent-core`**. That placement is the graph
asserting the empty set rather than the manifest claiming it.

9 files rewired across 4 consumer packages, 3 of 10 statements split.

## The extraction was verified, and the naive reading was wrong

A first pass over the block reported six outside types — `IExecutionOrigin`, `IHistoryEntry`,
`IInteractiveSessionRecord`, `ISpanCompletionEventData`, `TServerMessage`, `TUI`.

**All six are comment mentions**, explaining how the record-side projection relates to the framework's
event. Filtering to real field positions is what settled it. A scan that does not separate a field
from a doc comment reports this extraction as entangled when it is not — and the answer would have
been to abandon a correct plan.

## A regression in my own scan, found by my own test

`projectGraph` parsed only **relative** imports. A cross-family import is relative only while both
families share a package; once a leaf moves one out, the same dependency is written as
`@robota-sdk/agent-interface-<owner>` and a relative-only parser stops seeing it.

After three leaves, `session-contracts`' edges into execution, command and analytics had **all** become
package specifiers. The projection lost them, and `session` moved into wave 1 as though it depended on
nothing.

**ACYCLICITY was still green — on a graph that was quietly emptying.** Fewer edges make acyclicity
*easier* to satisfy, so the verdict was getting cheaper exactly as the work progressed.

Same class as the shrinking module set #2203 fixed, one level over: **there it was which FILES are
read; here it is which EDGES are seen.** Five regression tests, including a cycle whose only closing
edge is a package import.

The scope in that pattern is built from `HARNESS.internalPackagePrefix`, not hardcoded.
`harness-scope-literal` caught the literal and its reasoning is exact: *a hardcoded scope does not fail
when the scope changes — it matches nothing, and matching nothing reads as a pass.*

## Four of seven types had no assertion anywhere

Only `IUsageSnapshot` was covered, incidentally, in the transport package's contract test.
`IUsageSource`, `IUsageSourceTotals`, `IUsageBySourceReport` and `IRunTraceTurn` were asserted nowhere.

Extracting the family is the moment that became visible — a contract has no natural home for its own
assertions while it is a paragraph inside someone else's file. All seven are covered now.

## Baseline: a pure shrink

`file-size` absorbed **one change and no growth**: `session-contracts.ts` **471 → 360 (−111)**. Unlike
#2203 (+2) and #2209 (+1), this leaf pays no split cost — nothing was added, 110 lines left.

`build-types-ordered`: 78 → 79 packages. The comment already said it churns once per remaining leaf.

## Verification at `c9ea36f50`

- `pnpm harness:verify-like-ci` — PASS, all 12 stages
- `pnpm harness:scan` — 141 passed, 2 skipped, 0 failed
- `pnpm harness:test` — 4865 tests, 247 files, 0 failed
- workspace `pnpm typecheck` — exit 0 across all 79 packages

For a type-level extraction the compiler is the assertion: a missed or over-eager rewrite fails to
resolve.


woojubb:
## A false negative about feasibility, which is the direction nobody checks for

Worth stating separately from the PR body because of what it nearly cost.

Before extracting, I checked whether the seven analytics declarations are self-contained. A pattern
match over the block reported **six types from outside the moving set**:

`IExecutionOrigin`, `IHistoryEntry`, `IInteractiveSessionRecord`, `ISpanCompletionEventData`,
`TServerMessage`, `TUI`

Read at face value, that says the family is entangled with the session and framework layers and cannot
be extracted cleanly — which would have been grounds to abandon the plan, or to widen it into
something much larger.

**All six are comment mentions.** They appear in doc comments explaining how the record-side
projection relates to `agent-core`'s span event and how the report crosses the sidecar boundary. Not
one is a field type. Filtering to real field positions gives the true answer: every field of the seven
is a primitive or another member of the set, and the extraction is clean.

### Why this direction is the dangerous one

A pattern match over source that cannot distinguish a **field position** from a **comment mention**
produces two kinds of error, and they are not treated the same way:

- a **false positive** about a defect gets investigated, because someone has to fix it;
- a **false negative about feasibility** gets *believed*, because it argues for not doing something,
  and not doing something needs no further justification.

Nobody audits a plan that was abandoned. The measurement said "entangled", entangled means "harder
than we thought", and harder-than-we-thought is a conclusion that survives without evidence.

The only reason it was caught here is that ARCH-100's owner map had already recorded — from a
different measurement, months of leaves earlier — that this family is self-contained. **Two
independent measurements disagreeing is what forced the third one that settled it.** Had the map said
nothing, the naive reading would have stood.

No action requested; the extraction in this PR is the correct one and is verified. Recording it
because the general form is worth having: **a measurement that argues against work is not audited by
the work not happening.**

Issue #2184, opened 2026-08-23
allocate-work-item-id returns a plausible ID from a stale clone instead of refusing


woojubb:
## What happened

Measured 2026-08-23, during parallel multi-session work. A central allocator session ran:

```
$ node scripts/harness/allocate-work-item-id.mjs TRANS --dry-run
TRANS-006
```

…but only after a `git pull`. The run **immediately before** the pull, in the same clone, returned:

```
TRANS-005
```

`TRANS-005` had already been allocated, used, delivered and archived to `.agents/tasks/completed/` — merged as PR #2182 minutes earlier. The clone had fetched but not fast-forwarded, so its working tree predated the merge.

## Why nothing inside the script could catch it

The allocator reads three sources and says so: record files, every citation in the tracked tree, and open issue titles. Its header is explicit that reporting the ABSENCE of a source matters, because "an allocator that quietly skips a source allocates from a smaller set than it claims to."

That reasoning covers a source being *unreachable*. It does not cover every source being *stale together*. In a clone behind its upstream, the records are stale, the citations are stale, and the issue-title read is fine — so the two tree-derived sources agree with each other perfectly and the third does not contradict them, because a merged-but-unpromoted issue is still open by design here. The script is internally consistent and externally wrong.

**The failure mode is the dangerous one: it returns a plausible ID rather than an error.** A caller cannot tell a correct answer from a stale one, and the collision it produces is exactly the class the script was written to close (issue #1916, three collisions in one session).

## The specific irony worth recording

The allocator exists because reading the highest number and adding one has "a shelf life measured in minutes when more than one session is working." A stale clone gives that read a shelf life measured in *hours*, and hands it back through the tool that was supposed to have removed it. Running allocation centrally through one session — the mitigation for clone-local invisibility — concentrates the risk into whichever clone that session happens to be sitting in.

## Suggested fix

Refuse, rather than answer, when the clone cannot see the current tree:

- `git rev-list --count @{u}..` / `..@{u}` (or a `git fetch --dry-run` equivalent) to detect the branch being behind its upstream, and exit non-zero naming the gap.
- Say which check failed, in the same voice the script already uses for an unreachable issue list — the point is that a caller learns the answer is unavailable, never that it silently narrows.
- Offline is not the same as stale: a clone with no network should still allocate, with the existing "allocated from a smaller set" notice. Behind-upstream-and-knowable is what should refuse.

[...]

Issue #2175, opened 2026-08-22
develop is 1 warning over its own lint ceiling, and only the main-promotion path will find out


woojubb:
## Measured on `develop` @ `ce5266b71`

```
$ npx eslint packages apps --ext .ts,.tsx --no-cache -f json   → warnings=2093  errors=0
$ grep '"lint"' package.json
  "lint": "eslint packages apps --ext .ts,.tsx --cache --max-warnings 2092"
```

`develop` is **one warning over its own ceiling**, with no work in progress. Measured with `--no-cache` to rule out a stale-cache artefact, and independently reproduced in a second clone.

## Where it detonates

`pnpm lint` is not run by any develop-targeting pull request. CI's `quality` job runs `pnpm harness:verify -- --base-ref origin/<base>`, which lints only the affected package via that package's own `eslint src/ --ext .ts` — no `--max-warnings` involved. Nothing in `.github/workflows/` invokes `pnpm lint` at all.

The root script has exactly one caller:

```
"harness:verify:release": "… && pnpm typecheck && pnpm lint"
```

So the failure is latent on every feature PR and surfaces at the **`main` promotion**, on whoever runs the release verification — furthest from the change that caused it, and at the least convenient moment. Related: issue #1984 already records that this ratchet's only CI execution path is the promotion job and it has therefore never run on a pull request. This issue is the consequence that observation predicted, now measured as fact.

## Why a one-warning overshoot is worth filing

A ratchet exists to make a number unable to rise. This one is already above its own mark, so it currently constrains nothing — every state of the tree fails it equally, which means the signal carries no information about whether a given change made things worse. That is strictly worse than no ceiling, because the red is *expected* and therefore read as background noise, which is how a gate stops being read at all.

## What makes the fix a decision rather than a bump

The dominant warning class is `@typescript-eslint/ban-types` at **852 of 2093**, and the banned type is `unknown`:

```
"unknown": "❌ Avoid `unknown` in shipped source. Prefer specific unions/interfaces, SSOT types
            (e.g. TUniversalValue), and type guards at boundaries.
            Note: disabling this rule in src/ is prohibited; tests are exempt."
```

[...]

PR #2239, opened 2026-08-23

### LINKED ISSUE #1916: A work-item ID has no allocator, and three collisions landed in one session

### LINKED ISSUE #2019: P1 security: plugin marketplace inputs are interpolated into execSync shell commands
## Summary

Plugin marketplace URLs, installation paths and registry paths are interpolated into shell command strings. The production adapter passes those strings to execSync. A crafted marketplace source or a path containing shell metacharacters can therefore execute additional host commands during add, update, revision lookup, or plugin installation.

## Evidence

- [The framework defines the execution port as a single command string](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/plugins/marketplace-types.ts#L38-L49).
- [The production adapter sends that string to execSync](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-command/src/plugins/default-plugin-command-adapter.ts#L34-L38).
- [Marketplace add constructs git clone by interpolation](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/plugins/marketplace-client.ts#L78-L80).
- [Marketplace update and revision lookup interpolate stored paths](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/plugins/marketplace-client.ts#L170-L172) and [here](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/plugins/marketplace-client.ts#L228-L230).
- [Plugin installation constructs another interpolated git clone](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/plugins/bundle-plugin-installer.ts#L235-L239).

A capture-only reproduction supplied a URL containing a semicolon and a harmless marker command. The final command delivered to the execution port contained the marker unchanged. The probe did not invoke a shell or create the marker file.

## Security impact

An attacker who controls a marketplace URL, plugin repository URL, marketplace name, or persisted installation path can gain arbitrary command execution as the user running agent-cli. This can be reached through a copied command, a malicious shared configuration, or compromised plugin metadata.

## Required improvement

Replace the string-shell execution port with an argv-based process API, for example executable plus readonly argument array, and always invoke Git with shell disabled. Validate each URL and filesystem argument separately. Do not treat quoting helpers as the primary boundary: remove shell parsing from this workflow. Scrub inherited Git-related environment variables and define an explicit environment allowlist for plugin subprocesses.

## Acceptance tests

- URLs and paths containing spaces, quotes, semicolons, newlines, command substitutions and option-like prefixes are passed as literal arguments.
- Tests assert executable and argv separately rather than comparing a command string.
- Marketplace add, update, revision lookup and bundle installation have no execSync or shell-string path.
- Git subprocesses do not inherit unsafe GIT_* variables.

### LINKED ISSUE #2020: P1 security: plugin manifests and registries can escape plugin roots and drive recursive deletion
## Summary

Remote plugin manifests and local registry JSON are cast to TypeScript shapes after only shallow object checks. Names, versions, source paths, install locations and installPath values are then joined into paths or passed to recursive removal without canonical containment checks. A malicious marketplace or tampered registry can escape the plugin cache root and write, rename, copy, or delete arbitrary user-accessible paths.

## Evidence

- [Manifest parsing checks only that data is an object and name is a string before casting](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/plugins/marketplace-client.ts#L275-L289).
- [A remote marketplace name selects the final rename destination](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/plugins/marketplace-client.ts#L97-L119).
- [Plugin name and version form installation paths without segment validation](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/plugins/bundle-plugin-installer.ts#L73-L106).
- [Relative marketplace sources are joined without proving containment](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/plugins/bundle-plugin-installer.ts#L183-L203).
- [Install cleanup recursively removes the computed target](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/plugins/bundle-plugin-installer.ts#L220-L239).
- [Registry JSON is cast and its installPath is recursively removed](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/plugins/marketplace-registry.ts#L62-L78).

A safe temporary-directory reproduction used a manifest name of ../../escaped-market. The client accepted that name and created the marketplace outside the intended marketplace root; no real user path was targeted.

## Security impact

- Arbitrary directory placement and overwrite during marketplace installation.
- Arbitrary recursive deletion when uninstalling or cleaning a plugin/marketplace.
- Loading plugin content from outside the expected trust root.

## Required improvement

Define strict runtime schemas for every persisted and remote plugin structure. Restrict identifiers and versions to single safe path segments. Canonicalize the root and candidate path, then enforce descendant containment before every copy, rename, load or recursive delete. Resolve relative sources against the marketplace root and reject absolute paths and escapes. Treat registry paths as untrusted hints; derive owned paths from validated identifiers wherever possible. Refuse symlink traversal at mutation boundaries.

## Acceptance tests

- Reject dot segments, absolute paths, separators, drive/UNC paths, NULs and encoded traversal in every identifier/path field.
- Tampered registry installPath values cannot delete outside the plugin root.
- Symlinks inside plugin roots cannot redirect copy, load, rename or deletion outside the root.
- Failed validation performs no filesystem mutation and reports a typed error.

### LINKED ISSUE #2021: P1 security: resumable session records are world-readable by default and are not excluded from Git
## Summary

agent-cli persists complete resumable session records under the project .robota directory. SessionStore creates the directory and JSON files without owner-only modes, so under the common umask 022 they are 0755 and 0644. robota init also does not add session records to .gitignore. These records contain user prompts, assistant output, tool calls/results and session metadata and can expose source code, secrets or personal data to other local users or Git history.

## Evidence

- [SessionStore creates its base directory without a restrictive mode](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-session/src/session-store.ts#L48-L51).
- [SessionStore writes the temporary record without a restrictive mode](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-session/src/session-store.ts#L78-L85).
- [The CLI constructs the project-local session store by default](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-cli/src/cli.ts#L327-L334).
- [robota init writes settings and AGENTS.md but no targeted ignore rule](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-cli/src/init/init-command.ts#L183-L188).

A safe probe under umask 022 measured a newly created sessions directory as 0755 and its JSON record as 0644.

## Security impact

- Other accounts on a shared workstation or CI host can read conversation transcripts and tool output.
- Session files can be accidentally committed or uploaded with a repository.
- Secrets returned by tools or pasted into prompts persist beyond the terminal session.

## Required improvement

Create session directories with mode 0700 and files with mode 0600, including temporary files used for atomic replacement. Verify or tighten existing permissions rather than assuming mkdir mode applies to an existing path. Add targeted project ignore entries for session records and other runtime-secret artifacts during initialization without blanket-ignoring reviewable project settings. Document retention, deletion and sensitivity. Consider a user-global owner-only default store with an explicit opt-in for project-local persistence.

## Acceptance tests

- Under permissive umasks, newly created and replaced session files remain 0600 and directories 0700.
- Existing overly broad directories are rejected or tightened safely.
- robota init adds idempotent, targeted ignore rules for runtime session data.
- Windows ACL behavior has an equivalent platform-specific test or documented secure fallback.
- Resume and session-analysis features still work after the permission change.

### LINKED ISSUE #2022: P1 security: model and tool output can inject terminal OSC/CSI control sequences
## Summary

Untrusted model output and tool output reach Ink Text components without a terminal-control sanitization boundary. Markdown rendering preserves existing escape bytes, and both completed and streaming responses are rendered to the terminal. A malicious repository, fetched document, plugin or model response can emit OSC/CSI sequences that manipulate the terminal independently of the visible transcript.

## Evidence

- [Markdown rendering returns input text while applying presentation escapes](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-transport-tui/src/render-markdown.ts#L110-L118).
- [Completed message content is rendered through MessageList](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-transport-tui/src/MessageList.tsx#L262-L318).
- [Streaming text is rendered by StreamingIndicator](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-transport-tui/src/StreamingIndicator.tsx#L67-L116).
- Tool summaries and error messages share the same terminal transport and need the same trust boundary.

A capture-only probe constructed an OSC 52 sentinel and passed it through renderMarkdown; the returned string still contained the OSC sequence. The sentinel was inspected as encoded data and was not printed to a terminal.

## Security impact

- Clipboard overwrite through OSC 52 on supporting terminals.
- Deceptive hyperlinks through OSC 8.
- Window-title, cursor, erase and screen-spoofing attacks through other OSC/CSI sequences.
- Fragmented streaming escape sequences can evade a sanitizer that processes chunks independently.

## Required improvement

Add one centralized, allowlist-based terminal text encoder at the transport boundary for every untrusted field. Strip C0/C1 controls and OSC/DCS/APC/PM/SOS/CSI sequences while preserving ordinary Unicode, tabs/newlines as intended, and ANSI styling generated internally after sanitization. Streaming must use a stateful sanitizer so an escape split across chunks is still removed. Plain, JSON and file outputs should define their own non-terminal policy rather than sharing terminal escapes.

## Acceptance tests

- OSC 52, OSC 8, title changes, cursor movement and erase sequences are neutralized in assistant, tool, error and plugin output.
- Every split point of a control sequence across streaming chunks is tested.
- Internally generated colors and formatting continue to render.
- Sanitization cannot be bypassed by Markdown code blocks, links, diffs or carriage returns.

### LINKED ISSUE #2080: P1 architecture leaf: define the contract-family owner map and acyclic target graph
Parent tracker: #2068

## Outcome

Amend the architecture/rule documents with the final owner package for every contract family before any package move begins.

## In scope

- `.agents/project-structure.md` interface-package rule
- architecture map and dependency-direction documentation
- a migration order that prevents temporary cycles

## Acceptance criteria

- Every exported family in `agent-interface-transport` has exactly one target owner.
- The proposed package graph is acyclic and checked against current consumers.
- The rule change names the mechanical owner/dependency guard to add.
- No production TypeScript is moved in this issue.

## Scope boundary

This is one independently mergeable leaf. New concerns become sibling issues under #2068; they do not expand this issue. The audited API is prerelease, so do not add compatibility shims or forwarding aliases.

### LINKED ISSUE #2081: P1 session leaf: implement the versioned InteractiveSessionRecord decoder
Parent tracker: #2067

## Outcome

Add the single owner-owned runtime decoder for a complete persisted interactive-session record.

## In scope

- record schema/version and nested field decoders
- field-path diagnostics
- malformed-record corpus and property tests

## Acceptance criteria

- The decoder accepts `unknown` and returns only a fully validated record.
- All nested messages, events, tasks, goal/plan, dates, IDs, and branch data are covered.
- No store, artifact, handoff, or log consumer is migrated in this issue.

## Scope boundary

This is one independently mergeable leaf. New concerns become sibling issues under #2067; they do not expand this issue. The audited API is prerelease, so do not add compatibility shims or forwarding aliases.

### LINKED ISSUE #2083: P1 hook leaf: define decoded allow, deny, and error outcomes
Parent tracker: #2075

## Outcome

Replace exit-code/truthiness interpretation with one decoded hook-execution result contract.

## In scope

- `allow | deny | error` result union
- HTTP/prompt/agent response decoder
- command-executor exit mapping

## Acceptance criteria

- Non-boolean or missing `ok`, malformed JSON, timeout, spawn failure, and HTTP failure become `error`.
- Reason/source metadata is preserved for diagnostics.
- Session enforcement policy is not changed in this issue.

## Scope boundary

This is one independently mergeable leaf. New concerns become sibling issues under #2075; they do not expand this issue. The audited API is prerelease, so do not add compatibility shims or forwarding aliases.

### LINKED ISSUE #2097: P1 session leaf: route artifact and handoff ingestion through the record decoder
Parent tracker: #2067
Blocked by: #2081

## Outcome

Make portable artifacts and live handoff validate the complete session record with the owner codec before persistence.

## In scope

- session artifact deserialize path
- handoff destination integrity-then-decode flow
- artifact/handoff malformed corpus tests

## Acceptance criteria

- Envelope integrity/version checks are followed by the same nested record decoder.
- Handoff never stages or persists an undecoded record.
- Store and JSONL behavior are out of scope.

## Scope boundary

This is one independently mergeable leaf. New concerns become sibling issues under #2067; they do not expand this issue. The audited API is prerelease, so do not add compatibility shims or forwarding aliases.

### LINKED ISSUE #2108: P1 topology leaf: move command contracts to their domain owner

### LINKED ISSUE #2109: P1 topology leaf: move background, subagent, and workspace contracts to the execution owner

### LINKED ISSUE #2110: P1 topology leaf: move session, interaction, event, and persistence contracts to runtime owners

### LINKED ISSUE #2111: P1 topology leaf: move peer and handoff contracts to the session-mobility owner
Parent tracker: #2068
Blocked by: #2080, #2090

## Outcome

Move peer messaging, handoff, and transfer semantic contracts out of the transport contract hub.

## In scope

- peer message/session handoff contract families
- mobility/protocol/framework imports
- package exports and SPEC registry

## Acceptance criteria

- Mobility semantics have one owner.
- Wire codecs import semantic DTOs without owning them.
- No compatibility re-export remains.

## Scope boundary

This is one independently mergeable leaf. New concerns become sibling issues under #2068; they do not expand this issue. The audited API is prerelease, so do not add compatibility shims or forwarding aliases.

### LINKED ISSUE #2112: P2 topology leaf: move analytics and usage contracts to their reporting owner

### LINKED ISSUE #2135: Delete Merged Branches: the ancestry check forbids deleting any squash-merged branch, and 57 have accumulated
## The rule's verification forbids the action the same rule mandates

`.agents/rules/git-branch.md` § "Delete Merged Branches (mandatory)" says a merged branch **must not be left standing**, then requires this before a remote delete:

> **Verify before remote deletion:** `git merge-base --is-ancestor origin/<branch> origin/main` (or `origin/develop` for non-release merges) must succeed. If it does not, the branch carries commits the target does not have — **do not delete it; surface it**.

**This repository squash-merges.** A squash merge creates a new single commit on the target; the branch's own commits become ancestors of nothing. So the check fails for every squash-merged branch, and an agent following the rule honestly reaches *"do not delete it; surface it"* every time.

## Measured 2026-08-22

| | |
| --- | --- |
| merged branches still on `origin` | **57** |
| of those, `--is-ancestor` FAILS | **49** |
| of those, it succeeds | 8 (the ones merged with a merge commit rather than squashed) |

Among the 49 is `feat/task-records-name-their-issue`, merged as pull request #2011 earlier today and watched through by two sessions.

**None of those 57 is a mistake.** Each is the correct output of the rule as written. This is not fifty-seven people forgetting; it is one rule that cannot be satisfied.

## Why nothing ever reported it

The rule **fails toward inaction**. A guard that produces a false verdict eventually contradicts something. A guard that refuses a correct action produces only a growing pile, and the pile is the sole evidence — you have to go looking for it, and nothing prompts you to.

Both halves of the rule are correct in isolation:

- "do not leave a merged branch standing" — right;
- "do not delete a branch whose commits the target lacks" — right.

They became incompatible when the merge method changed, and nothing re-reads a rule against a change in how the repository operates. The ancestry check encodes a contingent fact — *we merge with merge commits* — as a property of merged branches.

## The sound check for a squash merge

Ancestry of the **branch** is the wrong subject. The right one is the **pull request's merge commit**:

```bash
gh pr list --state merged --head "<branch>" --json number,mergeCommit
git merge-base --is-ancestor "<mergeCommit>" origin/develop   # or origin/main
```

That answers "was this branch's work actually merged into the target" under both merge methods, which is the property the current check was reaching for. It also keeps the guard's real value: a branch with **no** merged pull request, or whose merge commit is not on the target, still refuses deletion.

## Suggested change

1. Replace the ancestry-of-branch step with ancestry-of-merge-commit, stated for both merge methods.
2. Keep "surface it, do not delete" for the genuine case — no merged pull request, or a merge commit absent from the target.
3. Clean up the 57 on that basis, after each holder confirms none must stay. The rule already allows a branch to stay with a recorded reason, so the sweep asks rather than assumes.

## Attribution note, since it came up while measuring

Branch authorship is recoverable only from the `Claude-Session` trailer, which survives squash. 28 of the 57 carry no trailer at all and cannot be attributed to a session by any field — every session authenticates as the same account, so `actor` carries no information. That is a separate observation and not a blocker for this cleanup, but it bounds what "ask the author" can achieve.

### LINKED ISSUE #2178: Harness: nothing asserts a routing document still states the rules it owns
## The gap

`.agents/project-structure.md` § Interface Package Rule owns five mandatory rules:

- an `agent-interface-*` package contains no classes or runtime logic (HARNESS-103)
- its internal dependencies are a subset of `{agent-core}` (INFRA-025 / the `INTERFACE-DEPS` rule)
- implementation packages import contracts from the interface package, not `agent-framework` (INFRA-010)
- `agent-framework` depends on `agent-interface-transport`
- interface packages are not placed in `agent-core`

**If all five were deleted, no check in `pnpm harness:scan` would go red for the deletion.**

The scans that enforce those rules (`interface-runtime`, `deps`, `interface-imports`) read the
CODE. They keep working perfectly with the prose gone — they do not verify that the document
claiming to own the rule still states it. `routing-document-size` reads the file, but it is a
**ratchet on line count**: it fires when a document GROWS. A deletion moves it in the safe direction.

## Why this is not hypothetical

It happened during #2080 (ARCH-100). Extracting a new section out of that file sliced from the new
heading to the next `##`, and the pre-existing `Rules:` block sat inside that span. Five mandatory
rules were relocated into a document that does not own them. Nothing was lost from the repository,
but `.agents/project-structure.md` stopped stating its own rules and every scan stayed green.

**What caught it was luck, not coverage, and that is the point of this issue.** The change happened
to also ADD lines, so the net went over the frozen ceiling and `routing-document-size` fired — a
check about SIZE, reporting a CONTENT loss it was not watching for. Follow-up on that red is what
surfaced the deletion. A change that deleted the same five rules while removing 27 lines elsewhere
would have left the ratchet green and nothing would have spoken.

The general defect is independent of how the rules came to be deleted: a slice-based extraction takes
whatever lies between two anchors, and owning the anchors does not mean owning the span. But an
accidental slice is only one route in. A deliberate "tidy up" edit reaches the same end state.

## Why the existing checks do not cover it

| check | what it reads | why it stays green |
| --- | --- | --- |
| `interface-runtime`, `deps`, `interface-imports` | package source | enforce the rules' CONTENT against code; indifferent to whether any document states them |
| `routing-document-size` | line count of routing docs | a ratchet against growth — a deletion is movement in the permitted direction |
| `conflict-markers` | cross-document contradictions | needs two statements to compare; one deleted statement contradicts nothing |
| `doc-authority` / `document-standards` | document structure and type | judge shape, not whether a named rule survives |

Every one is a check on something real. None of them is a check that a rule-owning document still
contains its rules.

## What would close it

A mechanical assertion that each rule-owning document still states the rules attributed to it —
for example, a declared inventory of `(document, rule anchor)` pairs that a scan verifies, so
deleting a rule's statement fails the gate the way deleting its enforcement would. The specific
mechanism is open; the requirement is that the assertion fails on deletion rather than only on
growth.

Related but not duplicates: this is not `conflict-markers` (two statements disagreeing), and not
`routing-document-size` (a document growing past its budget). It is a rule's statement disappearing
with its enforcement intact and every gate green.

## Evidence

- The five rules: `.agents/project-structure.md` § Interface Package Rule.
- The near miss and its restoration: PR #2176, commits `fe58700b9` and `43a05114b`.
- The ratchet that fired sideways: `routing-document-size`, frozen at 385 lines for that file.

### LINKED ISSUE #2179: Harness: interface-family-owner cannot see braceless relative imports (export */import * as)
Follow-up to #2080 / PR #2176. **Hardening, not a live defect** — see "Impact today" below before treating this as urgent.

## What the parser still cannot see

`scripts/harness/scan-interface-family-owner.mjs` projects the real source of
`packages/agent-interface-transport/src` onto the contract-family owner map and fails when the
resulting package graph has a cycle. It parses relative-import edges with a regex.

PR #2176 corrected that regex after a MUST review finding: it had matched only
`import … from './x.js'`, so extension-less imports and every `export … from` re-export edge were
dropped silently. The corrected pattern handles both keywords and `.js` / `.ts` / `.mjs` / `.mts` /
no extension.

It still **requires braces**, so two braceless forms remain invisible:

```ts
export * from './x';              // re-exports everything the target declares
import * as ns from './x';        // binds the target under a namespace
```

Neither names a symbol, so an edge from either should resolve to the **target module's owner** rather
than through the symbol-level owner map.

## Impact today

**Zero live instances outside the barrel.** The only star re-export in the package is:

```
packages/agent-interface-transport/src/index.ts:298   export * from './session-mobility-contracts.js';
```

`index.ts` is excluded from the graph by design — the barrel is not a contract module and has no owner
in the map. There are no namespace imports in the package at all.

So no edge is being dropped right now and the current ACYCLICITY verdict is not affected. The reason
to fix it is forward-looking: the moment someone adds `export * from './x'` between two owner
packages, a real edge disappears from the graph and the gate passes a graph that may genuinely have a
cycle — with nothing reporting it.

## Why this is the same defect class, stated once

A checker that reads source with a regex is only as good as the spellings it enumerated, and the
spellings it missed are invisible by construction: **the gate cannot tell "no edge" from "an edge I
cannot parse."** This scan has already been caught by that once (the MUST on #2176, where it was
blind to a third of the real edges by its own unit of measure). Closing the reported instances while
leaving a known form unhandled is how the class survives a fix.

## Work already done, held out of PR #2176 deliberately

The fix and its tests are written and verified locally, and were **not** pushed:

- pattern extended to `(?:export\s+\*|import\s+\*\s+as\s+<ident>)\s+from\s*'\./<module>(\.m?[jt]s)?'`
- two regression tests, **both confirmed RED against the corrected parser before the fix** — each
  written so the braceless edge is the ONLY link between two owners, so the verdict flips rather than
  the fixture merely containing the form
- full suite 25 passing, scan green on real source

It was held back because `.agents/rules/git-branch.md` § "An open PR's diff is frozen except to
resolve a finding" permits a push only to resolve a **reported** finding, and this is a hardening
suggestion for a form with no live instance — the "improvement the reviewer did not ask for" that rule
names. Pushing it would also have invalidated a clean verdict quoting the exact base/head and cost
another review round.

That tension is the subject of #2145 ("PR review loops cannot persist a clean round without violating
the frozen-diff rule"); this issue is the artefact that carries the work forward instead.

## Acceptance criteria

- `export * from './x'` and `import * as ns from './x'` produce graph edges to the target module's
  owner.
- A regression test exists for each in which the braceless edge is the **only** link between two
  owners, and each is demonstrated to fail against the pre-fix parser.
- `pnpm harness:scan` exit 0.

## Deliberately not in scope

Extending the graph to `index.ts`. The barrel re-exports everything and has no owner; including it
would add edges from a non-module to every family. If that changes, it is a separate decision about
what the graph models.

### LINKED ISSUE #2180: P1 blocker: INTERFACE-DEPS forbids the contract-package decomposition it is supposed to permit

### LINKED ISSUE #2194: The architecture map states its relationship to SPEC.md; SPEC.md does not state its relationship to the map
## What exists — one direction

`.agents/specs/ARCHITECTURE-MAP.md`, third paragraph:

> Package `docs/SPEC.md` files remain the owner contracts for each package. Architecture-map subdocuments explain cross-package structure, dependency direction, deployment topology, and durable architecture lessons **without duplicating package-level contract truth**.

That is the map telling itself what not to duplicate. It is correct, and it is the whole of the stated relationship.

## What does not exist — the other direction

Measured on `develop`:

| Question | Answer |
| --- | --- |
| Does `spec-writing-standard` mention the architecture map? | **No** — zero references. Its "Architecture Overview: Layer structure" means layers *inside* the package. |
| Does `spec-template.md` have a section for the package's family, layer, or permitted dependencies? | **No** — Scope, Boundaries, Architecture Overview, Type Ownership, Public API Surface, Extension Points, Error Taxonomy, Test Strategy, User-Facing Contract, Class Contract Registry. None of them names the package's position in the repository. |
| How many package SPECs reference the map? | **1 of 58.** |
| Does any scan compare a SPEC's declared surface or dependencies against the map's `dependency-direction.md`? | **No.** `check-architecture-map-completeness.mjs` checks the map's own structural spine; `check-architecture-map-paths.mjs` checks that paths cited *in the map* resolve. Both are about the map's internal integrity. |

So a SPEC author is never told the map exists, never told their package sits in a declared family or layer, and never asked to make the package's dependencies agree with `dependency-direction.md`. A SPEC can declare a surface and a dependency set that the map forbids, and no document or check connects the two.

## Why it is not a documentation nicety

A package's **layer** currently has no per-package home. The map owns `architecture-map/dependency-direction.md` — "layer ownership and target dependency rules" — and SPEC.md, which is the package's own contract document, says nothing about it. The one fact that determines whether a dependency edge is legal lives only in a document the package does not reference.

That gap is what made a live contradiction invisible:

- ARCH-100 (#2080, merged as #2176) added `.agents/specs/contract-family-owner-map.md` declaring a target layer graph — `session → command, execution, analytics`.
- `.agents/project-structure.md` § Interface Package Rule and `check-dependency-direction.mjs:235` forbid exactly those edges.
- Not one package SPEC references either document, so nothing brought them into contact. The contradiction surfaced only when a scan was written against the map by hand (#2180), after the map had merged.

The repository has now ruled (owner, 2026-08-23) that same-layer packages must not depend on each other while different-layer one-directional composition is fine. **That rule is not mechanically checkable until layer is declared somewhere a check can read it** — and the natural candidates are the map, each SPEC, or both with a stated owner. Which one owns it is precisely the relationship this issue says is unwritten.

## What would close it

State the relationship from the SPEC side, in the documents a SPEC author actually reads:

1. `spec-writing-standard` and `spec-template.md` name the map, and say which facts belong to which document — the map owns cross-package structure and layer; the SPEC owns the package's contracts — so neither restates the other.
2. A package's family and layer get one owner, named explicitly rather than left to convention.
3. A check that the two agree, at whatever granularity the owner decision above makes checkable. Without it this is prose that can drift the same way, and the repository's own record says prose alone did not hold before (`check-contract-disposition.mjs`: "Prose did not hold, so this is the mechanical half").

Raised by the repository owner, 2026-08-23.

### LINKED ISSUE #2198: agent-provider-defaults bundles four providers while carrying their prefix
## The shape

`agent-provider-defaults` depends on four packages that share its own prefix:

```
agent-provider-defaults   out:[anthropic, gemini, openai, openai-compatible]   in:[]
```

Nothing depends on it in return. It is the family's **aggregator** — the package that bundles the providers and offers them as one — and it sits inside the family it bundles.

Owner ruling, 2026-08-23:

> 하나의 패키지가 여러 동급 패키지를 가지고 있는 특수한 패키지가 있다면 그 패키지는 묶어서 하나로 제공하려는 패키지 같은거 같은데 그건 접두어를 agent-interface가 아닌 다른걸로 바꿔야 할수도 있습니다.

A package that holds several same-level packages in order to offer them as one should not carry the prefix of the things it bundles.

## Why the prefix is the problem and not the dependency

The dependency is legal. The owner's layer rule permits one-directional composition across layers, and an aggregator is a layer above what it aggregates — the same shape as `dag-framework` over `dag-core`, which nobody reads as a defect.

What the shared prefix does is make that invisible. `agent-provider-*` reads as "one provider", so a package under it that depends on four other providers looks like a peer reaching sideways. Measured across the workspace, this is one of only two same-family edge groups that required a judgement call rather than resolving on sight, and both turned out to be naming rather than structure (the other is #2197).

There is also a reader cost independent of the rule: someone choosing a provider sees five `agent-provider-*` packages of which one is not a provider.

## What would close it

Re-prefix `agent-provider-defaults` out of the provider family, into a name that says it is an assembled default set rather than a provider. `agent-provider-replay` and `agent-provider-bytedance` are genuine providers and stay.

Pre-release, so the rename is available: `.agents/rules/code-quality.md:59` — *"Pre-release, existing files, rules, packages, or names are not preserved for their own sake — relocating, renaming, deleting, or recreating them is permitted (and expected) when it reaches the correct end-state."*

## Measured scope of the same shape elsewhere

At the two-segment prefix level this is the **only** true instance in the workspace. The one other package depending on two same-prefix siblings, `agent-transport-webrtc-web → [gui, protocol]`, is not an aggregator — it is a browser surface over a presentation core and a wire base, two different layers, and its naming problem is #2197's.

At a one-segment reading, `dag-framework` (10 siblings) and `dag-cli` (7) have the same aggregator shape inside the `dag-` prefix. Recorded as observation rather than folded in: whether a one-segment prefix constitutes a "family" for this rule is a separate question, and answering it here would decide it silently.

`agent-interface-*` has **no** aggregator today — neither of its two members depends on the other. The concern there is forward-looking: the ARCH-100 target adds six interface packages, and if a bundle over them is ever wanted it must not be `agent-interface-*`.

woojubb:
Promotes `develop` (`80364022d`) to `main` (`68fc0490e`) — 42 commits.

Approved by the owner. `promote.mjs` reports the merge is clean and promotes develop's tree unchanged, and the required-status-check declarations reconcile against the live rulesets.

## No release side effect

`@robota-sdk/agent-cli` is `3.0.0-beta.79` on both branches. `release-tag-on-version-bump.yml` pushes a tag only when that package's version bump lands on `main`, so this promotion pushes no tag, cuts no binaries, and publishes nothing.

## Why this is the work, not bookkeeping

GitHub interprets a closing keyword only on a pull request that targets the default branch. The default branch here is `main` and every change routes feature → develop → main, so **the `Closes #N` that work PRs write is ignored on all of them** — the promotion is the only PR that can fire them, which is what `promotion-closes.mjs` exists to bridge.

Twenty issues below were finished and stayed open for that reason alone. Four are P1 security items whose fixes merged today: terminal control-sequence injection (#2022), world-readable session records (#2021), plugin path escape (#2020), and shell interpolation in the plugin marketplace (#2019).

## One closing line that is closing an unfinished issue, recorded rather than quietly carried

`Closes #1916` comes from PR #2133's own declaration and the derivation is faithful to it. **The issue's stated bar is not met**, and the repository's own documentation is what says so — `.agents/tasks/README.md` states that a collision became DETECTABLE at push time and is not PREVENTED, while #1916 asked for a mechanism that makes one *impossible to commit*. Issue #2184 is a live instance from today: the allocator returned already-spent IDs from a clone that had not pulled.

The body is not edited to drop the line — the derivation is the SSOT and `promotion closes` verifies the two match. The reasoning is on the issue as a comment and it will be reopened immediately after this merges.

Closes #2021
Closes #2022
Closes #2111
Closes #2110
Closes #2112
Closes #2020
Closes #2108
Closes #2019
Closes #2109
Closes #2194
Closes #2198
Closes #2180
Closes #2097
Closes #2083
Closes #2178
Closes #2179
Closes #2081
Closes #2080
Closes #2135
Closes #1916

Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012fmnzs5W63fCNLiG4xEJLY


woojubb:
Closing rather than holding this open, and re-cutting after the blocker lands.

**`release-grade verification` failed on the lint-warning ceiling**: 2198 warnings against a `--max-warnings` ceiling of 2092, exit 1, zero errors. That is issue #1984 and issue #2175 landing exactly as they predicted — the ratchet's only CI execution path is this promotion job, so 105 warnings accumulated across 42 commits with every develop PR green, and the first thing to notice was a promotion that could not merge.

Two reasons to close rather than wait:

- **The fix has to land on `develop` first.** A promotion promotes develop's tree unchanged; it cannot carry its own blocker's repair.
- **This branch is already stale.** It promotes `80364022d`; develop is `f8f344eb8` (TRANS-007 merged), with ARCH-108 and SEC-016 close behind. Re-cutting after those land makes one promotion instead of two, and a promotion held open while develop moves under it is the stale-base problem in its own most expensive form.

Nothing about the closing set changes — the twenty issues stay open and the next promotion carries the same derivation, plus whatever lands in between.

The blocker is being fixed now: re-freeze the ceiling at the measured count AND make it run on every pull request, so the raise is paid for by the mechanism that makes it the last one. Raising a ceiling at the moment it first catches something, without closing the hole that let it grow unseen, is how a ratchet stops being one.

PR #2246, opened 2026-08-23
fix(harness): run the lint-warning ceiling on every pull request (issue #1984)

woojubb:
Closes the blocker that stopped today's promotion. Approved by the owner before the work started.

## What happened

`release-grade verification` failed on the promotion to `main`: **2203 warnings against a `--max-warnings` ceiling of 2092**, exit 1, zero errors.

The ceiling is `--max-warnings` on the root `lint` script, and the only CI path that executed it was `release-grade verification` — required on pull requests to `main`, therefore run **once per promotion**. So 111 warnings accumulated across 42 commits with every develop pull request green, and the first thing to notice was a promotion that could not merge.

That is issue #1984 (*"the ratchet's only CI execution path is the promotion job, so it has never run on a pull request"*) and issue #2175 (*"develop is 1 warning over its own ceiling, and only the main-promotion path will find out"*) arriving exactly as written. #2175 saw the first single warning of the 111.

> A gate that runs once per release reports a number nobody can act on. By the time it speaks, the additions are spread across dozens of merged commits and no author is left to attribute them to — the count arrives as a fact about the repository rather than as a finding about a change.

## Two halves, and the second is what pays for the first

**The ceiling is re-frozen at the measured 2203.** On its own that is the thing a ratchet must not do: the governing rule is that the baseline may fall and must never rise without a deliberate re-freeze, and a re-freeze with nothing else in it is just the number moving. Raising a ceiling at the moment it first catches something is how a ratchet stops being one.

**So the same change makes it run on every pull request to develop.** From here, a warming count is a finding on the pull request that added it, while its author still holds the change.

## Why the `quality` job

`pnpm harness:scan` runs on every pre-push and a full workspace lint is minutes — the ratchet scan rejects putting it there for exactly that reason, and that reasoning still holds. CI has the time the pre-push path does not. `quality` is already a required context on develop, so this gates **without a ruleset change**.

## Two things the harness refused, both correctly

**`ci-mirror-map`'s anti-drift test blocked the first push.** The new required step was claimed by no `verify-like-ci` stage and declared as no plumbing. Its message is the reason it matters — *leaving a required step unanswered means "I ran the CI-equivalent check" covers less than it claims.* A mirror that omits a gate is the same defect as a gate that does not run, one layer over: both produce a green about a smaller question than the reader believes.

Declared as a **stage**, not plumbing, because it is a gate. And it is not redundant with `affected-verify`, which is the objection a reader reaches for: that stage lints the **affected scopes**, so the warnings it reports are the warnings in its packages. The ceiling governs the sum over all of them, and only a whole-workspace run produces that number — a warning added in one package is invisible to a scoped lint and counted by the ceiling.

**The ratchet scan's own docstring said the ceiling is enforced "on the promotion path"**, and its passing message said "on the release path, not by this scan". Both were true this morning and this change makes them false, so both are corrected here — the same rule this fleet applied to five documents on another leaf today.

## Cost, stated rather than discovered

`verify-like-ci` now runs a whole-workspace lint, so the local mirror is minutes slower. That is the cost of the gate being real, and the three lanes were told before this was pushed rather than left to find it as a slowdown.

## Not in this change

Reducing the 2203. `943` are `@typescript-eslint/ban-types` and `624` are `no-magic-numbers`, which split by rule into mechanical passes rather than one sweep across every lane's in-flight work. Filed separately.

## Verification

```
pnpm lint                                          -> exit 0 at the new ceiling
node scripts/harness/scan-lint-warning-ratchet.mjs -> passes, ceiling 2203 at baseline
pnpm harness:scan                                  -> 141 passed, 2 skipped, 0 failed
pnpm exec vitest run scripts/harness/__tests__/ci-mirror-map.test.mjs -> 51 passed
```

The count was measured on `f8f344eb8` with `eslint ... -f json`, summing `warningCount` across 1949 files, rather than read off a summary line.

Refs issue #1984, issue #2175

Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012fmnzs5W63fCNLiG4xEJLY

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -427,6 +427,24 @@ jobs:
         if: env.PRODUCT_APPLICABLE == 'true'
         run: pnpm harness:verify -- --base-ref origin/${{ github.base_ref }} --skip-build --skip-record-check --skip-repository-check harness-tests
 
+      # issue #1984 — the lint-warning CEILING runs on EVERY pull request, not only the promotion.
+      #
+      # The ceiling is `--max-warnings` on the root `lint` script, and until this step existed the
+      # only CI path that executed it was `release-grade verification`, which is required on pull
+      # requests to `main` and therefore runs once per promotion. Measured 2026-08-23: 111 warnings
+      # had accumulated across 42 commits with every develop pull request green, and the first thing
+      # to notice was a promotion that could not merge. A ratchet nothing turns is a number.
+      #
+      # It lives HERE and not in `harness:scan`: that scan runs on every pre-push and a full
+      # workspace lint is minutes, which `scan-lint-warning-ratchet.mjs` rejects for that reason. CI
+      # has the time and the pre-push path does not. `quality` is already a required context on
+      # develop, so this gates without a ruleset change.
+      #
+      # `--cache` is deliberate and load-bearing for the runtime; the count it reports is unaffected.
+      - name: Lint-warning ceiling
+        if: env.PRODUCT_APPLICABLE == 'true'
+        run: pnpm lint
+
       # INFRA-037: build-contracts is the ONE scan that reads packages/*/dist (validates that declared
       # main/types/exports paths resolve to built files) and silently no-ops without dist. It stays in
       # `quality` (which restores dist above) so its coverage is preserved; the other 47 scans + the
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "test:coverage:packages": "pnpm --filter \"./packages/**\" run --if-present test:coverage",
     "test:coverage:apps": "pnpm --filter \"./apps/**\" run --if-present test:coverage",
     "typecheck": "pnpm run -r --workspace-concurrency=-1 --if-present typecheck",
-    "lint": "eslint packages apps --ext .ts,.tsx --cache --max-warnings 2092",
+    "lint": "eslint packages apps --ext .ts,.tsx --cache --max-warnings 2203",
     "lint:fix": "eslint packages apps --ext .ts,.tsx --cache --fix && prettier --write .",
     "lint:fix:staged": "scripts/harness/with-repo-lock.sh pnpm exec lint-staged",
     "dev": "pnpm run -r --if-present dev",
--- a/scripts/harness/ci-mirror-map.mjs
+++ b/scripts/harness/ci-mirror-map.mjs
@@ -161,6 +161,12 @@ export const CI_STAGES = [
     why: 'THE package test suites, lint and scoped typecheck — the gates verify-like-ci omitted entirely (INFRA-056)',
   },
   {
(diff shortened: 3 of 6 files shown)


woojubb:
### Rebased onto `c1dd93768`, and the ceiling was RE-MEASURED rather than carried forward

```
measured on 437a1dff2  ->  2203 warnings, 0 errors, 1950 files
```

Still exact. ARCH-108 (#2244) moved 85 symbols across 13 workspaces and added a file without adding a warning — 1949 files became 1950, the count did not move. The number this change freezes is true of the tree it will land in, not of the tree it was written against. That distinction is the whole subject of this pull request, so it would have been a poor place to assume.

```
pnpm lint                                            exit 0
node scripts/harness/scan-lint-warning-ratchet.mjs   passes, ceiling 2203 at baseline
ci-mirror-map.test.mjs                               51 passed
```

### `workflow provenance` is red by design and stays red

This change edits `.github/workflows/ci.yml`, which provides required contexts and is triggered by `pull_request` — so the edited definition is what judges this pull request. The gate refuses that, correctly, and its own documentation says it makes a self-edit *"impossible to merge rather than merely visible"*.

**What the gate asks for is visibility, and here it is, in the form the scan's own message names:**

- **What changes in the control plane**: one step added to the `quality` job — `run: pnpm lint`. Nothing else in `ci.yml` is touched. No job is added, removed, renamed or re-gated; no `if:` condition, `needs:`, trigger, permission or concurrency group changes.
- **Which required contexts this edit can move**: the gate lists them — `build`, `quality`, `scans`, `dependency audit`, `commitlint`, `tui-e2e`, `examples-typecheck`, `windows-shell`, `regression-red-proof` on develop; `promotion ancestry`, `main PR source guard`, `promotion closes`, `release-grade verification` on main.
- **Which of them this edit actually moves**: `quality`, and only by making it stricter. The added step can fail a pull request that would previously have passed. It cannot pass one that would previously have failed.
- **The direction is the part worth checking.** A control-plane self-edit is dangerous when it *weakens* the gate judging it. This one tightens it, and it tightens it against the change itself: this pull request's own `quality` run executes the new step.

**Approved by the owner** after the alternatives were measured. The one that avoids the control-plane edit — having the ratchet scan run eslint itself — was rejected because that scan deliberately does not run eslint (a full workspace lint is minutes and `harness:scan` is on the pre-push path), and because it would have split the scan's behaviour by environment while invalidating its own documented reasoning about declared sizes.

**This is the first pull request editing `ci.yml` since the gate landed on 2026-08-22 (PR #1977).** There was no precedent to follow and no written procedure, which is filed as issue #2256 — the gate closes the door and names no key, and whatever the first person improvises becomes the precedent. That issue asks for the procedure to live in `git-branch.md` beside the merge preconditions, and for the scan's message to point at it rather than be it.

Issue #2260, opened 2026-08-23
The next release removes four published exports and the package that now owns them is not on the registry


woojubb:
## The shape

Four runtime functions are exported from the **published** `@robota-sdk/agent-interface-transport@3.0.0-beta.79`, the current latest on npm. On `develop` they live in `@robota-sdk/agent-interface-session`, which **is not published to npm at all** and which nothing in the release configuration will publish.

```
$ npm view @robota-sdk/agent-interface-transport versions
[ ... "3.0.0-beta.77", "3.0.0-beta.78", "3.0.0-beta.79" ]

$ npm view @robota-sdk/agent-interface-session
npm error code E404 — 'is not in this registry'
```

The published tarball's own entry declaration, `package/dist/node/index.d.ts:1034`, ends its export list with:

```ts
export { …, readAssistantReplies, readErrors, readLastAssistantText, readToolCalls };
```

and on `develop` those four are here instead:

```
packages/agent-interface-session/src/interaction-contracts.ts:99  export function readAssistantReplies(...)
```

## Why the next release breaks a consumer with no migration target

- **`agent-interface-transport` will be republished.** 20 of the 82 pending changesets name it, and it is in the `fixed` group in `.changeset/config.json`.
- **`agent-interface-session` will not be.** Zero pending changesets name it, it is absent from the `fixed` group, and it has no published version to bump from.
- **Transport cannot re-export them.** Its only dependency is `@robota-sdk/agent-core`; it does not depend on the session package.

So the next publish ships a version of `agent-interface-transport` whose main entry no longer exports these four, while the package that now owns them does not exist on the registry. This:

```js
import { readAssistantReplies } from '@robota-sdk/agent-interface-transport';
```

works against `3.0.0-beta.79` today and fails after the next release, and **the consumer cannot repair it by adding a dependency**, because there is nothing to add. That last part is what separates this from an ordinary breaking rename.

Nothing is broken right now: `3.0.0-beta.79` is both the version in `package.json` and the latest on npm, so no publish has happened since the move. The window is open until the next release.

## What this is NOT

**Not the `/testing` subpath.** Both interface packages declare `"./testing"` in `exports` in the repository, but the published `beta.79` tarball declares only `"."` — the subpath has never been in a published artifact, so no registry consumer can have imported from it. A review finding on ARCH-108 about a removed `/testing` subpath was about repository-internal consumers, and it is a different and smaller problem than this one.

[...]

Issue #2250, opened 2026-08-23
Editing a PR body cancels the in-flight review of the head you just pushed


woojubb:
`claude-code-review.yml` reviews on `[opened, synchronize, reopened, edited]`, deliberately **skips**
the `edited` event, and shares one concurrency group across all four:

```yaml
concurrency:
  group: claude-review-${{ github.event.pull_request.number }}
  cancel-in-progress: true

if: ${{ … && (github.event.action != 'edited' || github.event.changes.base != null) }}
```

The group key contains no event. So an `edited` event — which the job then declines to act on —
**cancels the in-flight review of the current head and puts nothing in its place.**

## Measured on PR #2244

```
14:46:42  synchronize  cancelled   ← review of the newly pushed head 440d886f1
14:46:59  edited       skipped     ← `gh pr edit --body-file`, 17 seconds later
```

The push and the body update are the same routine sequence: push the branch, then update the PR
description to match. Doing them in that order destroys the review of what you just pushed.

## Why this is worse than a lost run

Nothing re-triggers a review afterwards. The PR is left carrying the verdict for the **previous**
head, and it looks like a normal verdict — the staleness is visible only by comparing its
`REVIEWED HEAD` against the tip.

On #2244 that stale verdict also asserted *"the MUST remains at the current HEAD"* while its current
HEAD was the pre-fix commit, so the surviving artifact was not merely old but actively wrong about
being current (issue #2249). **The cancellation is what left it as the newest thing on the PR.**

And the merge gate correctly refuses to merge on a verdict that does not name the current head — so
the observable symptom is a PR that cannot merge, with no failing check and no obvious cause. The
skipped job reports success.

## Repair

Either of:

1. **Put the event in the concurrency key** — `claude-review-${{ pr.number }}-${{ github.event.action }}` — so a skipping event cannot cancel a reviewing one. Simplest, and it keeps `cancel-in-progress` doing its real job: superseding an older review of an older push.
2. **Do not subscribe to `edited` at all**, and drop the `if` clause that exists only to skip it. The clause admits one case (`changes.base != null`, a retarget), so this means subscribing to that case specifically rather than to all edits.

(1) preserves the retarget case and is the smaller change.

## Falsification

Push a branch, wait for the review to start, edit the PR body, and assert a review for the new head
still completes. Today it is cancelled. Without this test the fix is unverified, because the failure
is a timing window and does not reproduce when the review has already finished.

## Note on the workaround

[...]

PR #2275, opened 2026-08-23
fix(tools): refuse a manifest whose security controls the applicator cannot apply (TOOL-005)

### LINKED ISSUE #2027: P2 security: workspace manifest permissions and environment are accepted but silently ignored
## Summary

The public workspace manifest contract declares environment variables and permission settings, but the manifest applicator processes only filesystem entries. Callers can provide security controls that validate at the type boundary yet have no effect. This creates false sandbox assurances: a consumer may believe a task has a restricted network, command or environment policy when the runtime applies none of it.

This corresponds to the already recorded internal task TOOL-005.

## Evidence

- [IWorkspaceManifest advertises environment and permissions](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-tools/src/sandbox/types.ts#L77-L86).
- [The applicator iterates and applies only entries](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-tools/src/sandbox/workspace-manifest.ts#L17-L38).
- [The apply result reports entry status only](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-tools/src/sandbox/types.ts#L95-L104).

There is no application, rejection or warning path for the two advertised fields.

## Security impact

- Applications can present a sandbox policy in configuration while executing with broader real capabilities.
- Secrets expected to be absent from the task environment may remain available.
- Network or command restrictions can be assumed by reviewers but never enforced.

Severity is P2 because exploitation depends on a caller relying on this misleading contract, but the result is a broken security boundary rather than a documentation-only mismatch.

## Required improvement

Either implement these fields end to end in the sandbox backend and report their effective state, or remove them from the accepted contract and reject manifests that contain them. Do not silently ignore future security-bearing fields. Capability support should be explicit per backend, fail closed when requested controls are unavailable, and be observable in the apply result.

## Acceptance tests

- A requested environment allowlist is actually reflected in the executed sandbox process, or is rejected before execution.
- Network/command permission requests are enforced by capable backends and fail closed on incapable ones.
- Unknown security-bearing manifest keys fail validation.
- The apply result reports requested versus effective policy, not only copied entries.

woojubb:
Closes #2027.

`IWorkspaceManifest` declares `environment` and `permissions`. The built-in applicator iterates `manifest.entries` and reads neither — it applied the entries and returned success, so a caller that asked for an environment allowlist or a read/write policy got a sandbox with **neither, and no way to find out**. That is why the issue is filed as security rather than cleanup: the contract advertises a control it does not deliver.

## The measurement chose between the issue's two options

The issue offers "implement these fields end to end" **or** "reject manifests that contain them". Measured before choosing:

```
production sites assigning manifest.environment or .permissions ....... 0
implementations of ISandboxClient.applyManifest ........................ 0
production callers of applyWorkspaceManifest ........................... 1
```

`applyManifest` is the optional escape hatch the applicator delegates to first, and it has **zero implementations** — so there is no backend today that could honour the two fields, and the branch that would hand them to one is dead in this repository.

Implementing them end to end therefore means implementing them for a backend that does not exist, and the acceptance test would be written against a double authored to satisfy it. The other option is available and smaller and true.

## What the change does

**Refuses before applying anything.** The refusal runs at the top of the built-in path, before the entry loop, so a refused manifest leaves nothing half-built. Previously the entries were applied and `{ entries: [...] }` came back looking like success.

**Checks emptiness, not presence.** `environment: {}` and `permissions: { read: [] }` request nothing; refusing them would fail a caller who asked for no controls at all. `permissions` counts as empty when neither list has an entry.

**Leaves the delegating branch alone.** A client implementing `applyManifest` is claiming ownership of the whole manifest, and this function cannot know what it honoured. Second-guessing it here would break a backend that does the right thing.

## What is deliberately NOT in this change

Making a delegating backend's claim **observable** — the third thing the issue asks for — needs a wider `IWorkspaceManifestApplyResult`, and that is a public type of a published package (`@robota-sdk/agent-tools@3.0.0-beta.79` is live on npm). Left on #2027, because #2260 is the open, live example of a published-surface change going wrong in this repository: four functions exported from a published package now live in one that is not on the registry. A change to that type should be checked against the published entry declaration, not against the repository's copy — which is a different piece of work than this one.

## Four mutants, each killed by the case that names its own subject

Not "the suite went red" — each mutant was checked against *which* case caught it, because a count survives a mutant that merely reshuffles which case lands in which bucket.

| Mutant | Killed by |
| --- | --- |
| the refusal is never called | 4 failed |
| refuses **after** the entry loop | the "leaves nothing half-built" case |
| an empty `permissions` treated as a request | the "accepts declared-but-empty controls" case |
| the refusal moved **before** the delegation | the "does not second-guess a backend" case |

The source was restored from a byte-identical copy after each mutant and verified with `diff -q`.

## Verification

```
tsc --noEmit -p packages/agent-tools     exit 0
eslint (both files)                      exit 0
workspace-manifest.test.ts               14 passed
pnpm harness:scan                        141 passed, 2 skipped, exit 0
harness:review:record                    0 findings @ a23163a62
```

One fixture correction on the way: the first draft used a `local-directory` entry type that does not exist, and `assertUnreachable` caught it — the invalid fixture failed for a reason unrelated to the property under test, which is the shape that has changed a conclusion four times in this repository today.

### DIFF
--- a/packages/agent-tools/src/__tests__/workspace-manifest.test.ts
+++ b/packages/agent-tools/src/__tests__/workspace-manifest.test.ts
@@ -198,8 +198,16 @@ describe('unenforceable manifest security controls (TOOL-005 / issue #2027)', ()
 
     const result = await applyWorkspaceManifest(client, manifest);
     expect(result.entries).toHaveLength(1);
+    // Pins the premise the refusal-ordering case depends on: an APPLIED entry really is readable at
+    // this path. Without it, `rejects.toThrow()` over the same path proves only that nothing is
+    // there — which a wrong path satisfies just as well. The negative control would become a
+    // tautology the day the target root changed, and silently.
+    await expect(client.readFile('/workspace/task.md')).resolves.toContain('Solve this task');
   });
 
+  // This records an ACCEPTED HOLE, not coverage: a delegating client still receives `environment`,
+  // and this function still cannot tell whether it honoured it. Making that observable needs a wider
+  // apply result, which is a published type — tracked on issue #2027.
   it('leaves the delegating path alone — a client owning applyManifest is not second-guessed', async () => {
     const manifest = { ...manifestWithOneEntry(), environment: { TOKEN: 'secret' } };
     let received: IWorkspaceManifest | undefined;
--- a/packages/agent-tools/src/__tests__/workspace-manifest.test.ts
+++ b/packages/agent-tools/src/__tests__/workspace-manifest.test.ts
@@ -165,6 +165,20 @@ describe('unenforceable manifest security controls (TOOL-005 / issue #2027)', ()
     await expect(applyWorkspaceManifest(client, manifest)).rejects.toThrow(/permissions/);
   });
 
+  it('refuses a non-array permissions member a JS caller can supply but the type does not declare', async () => {
+    const client = new InMemorySandboxClient();
+    // The type says `{ read?: string[]; write?: string[] }`. This package is published, so a
+    // JavaScript consumer reaches this function without that check. An array-only emptiness rule
+    // would read `.length` off a boolean, get `undefined`, and accept the request unenforced —
+    // which is the defect this function exists to refuse.
+    const manifest = {
+      ...manifestWithOneEntry(),
+      permissions: { network: true } as unknown as IWorkspaceManifest['permissions'],
+    };
+
+    await expect(applyWorkspaceManifest(client, manifest)).rejects.toThrow(/permissions/);
+  });
+
   it('names both fields when both are requested', async () => {
     const client = new InMemorySandboxClient();
     const manifest = {
--- a/packages/agent-tools/src/__tests__/workspace-manifest.test.ts
+++ b/packages/agent-tools/src/__tests__/workspace-manifest.test.ts
@@ -7,6 +7,7 @@ import {
   InMemorySandboxClient,
   validateWorkspaceManifestPath,
 } from '../index.js';
+import type { IWorkspaceManifest } from '../sandbox/types.js';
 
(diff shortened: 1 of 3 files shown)


woojubb:
### A known nuance in the refusal, recorded rather than fixed

`requestsSomething` treats a present-but-falsy non-array member as a request, so `permissions: { network: false }` is refused.

**That is deliberate and it is fail-closed, but `false` is genuinely ambiguous at this boundary.** It can mean *"apply a network restriction of: none"* or *"I am explicitly not asking for network control"*, and the guard reads it as the first. Refusing the ambiguous case is the correct default for a security surface — the cost of refusing a request that was not one is a failed call, and the cost of accepting one that was is a sandbox that reports a policy it never applied, which is the defect this change exists to remove.

What it does mean is that such a caller gets a message saying they *requested* a control when they may have meant the opposite. **That is a message problem, not a logic one**, and if it ever surfaces as a complaint this comment is the cause.

It stays as-is here because narrowing it would require this function to know which members are opt-outs and which are opt-ins, and it cannot: the member is not in the declared type at all — a JavaScript consumer of a published package is how it arrives. Distinguishing those needs the wider apply result that lets a backend say what it honoured, which is tracked on #2027 with the rest of the observability half.

Raised by an independent read-only review of this branch, which also produced the two findings in `34c86d804` and the residual in `1a73b299d`. Recorded as a comment rather than a body edit, since editing the body cancels the in-flight review of the head just pushed (#2250).

Issue #1734, opened 2026-08-15
ARCH-030: outbound protocol replies bypass the carrier delivery boundary


woojubb:
## Problem

`createWsHandler` currently has two outbound delivery semantics: subscribed session events use a carrier-safe delivery boundary, while replies produced by inbound request continuations receive the raw carrier `send`. After WS/WebRTC carriers began failing closed on a dead connection, delayed command, submit-error, background-log, job-group, or background-control replies can throw from an ignored Promise continuation and become an unhandled rejection. The delivery-error callback is bypassed.

## Root direction

Make one connection-scoped outbound delivery boundary own every `TServerMessage`, including parse/query/control replies and all asynchronous continuations. Route failures through the carrier lifecycle exactly once, isolate observer failures, and retain existing resume-buffer semantics. Add delayed-reply-after-disconnect regressions for protocol, WebSocket, and WebRTC carriers, including an unhandled-rejection assertion.

## Provenance

Found by the local PR review for ARCH-016/017/018/020/028. Depth guardian verdict: FOUNDATIONAL. The same failure mode was previously noted by WS-001, establishing recurrence.

## Reopened scope from ACLI-R1-F035 (2026-08-23)

The single outbound delivery boundary now also owns slow-consumer resources: queue/byte/time budgets, `accepted | queued | refused` results, overflow/drain/terminal policy, and shared enforcement for WS text/binary, WebRTC, and resume replay. Add non-reading peer and replay-burst contract cases.

woojubb:
Landed on `develop` via #1745 (squash `235da814f`).

One connection-scoped outbound delivery boundary now carries every outbound `TServerMessage` — the session-event fan-out and all eleven reply families alike. The carrier builds it from its own sink and its own failure policy and passes it down; `TOutboundDeliver` is branded and `createOutboundDelivery` is its only producer, so a raw `send` is refused by the compiler wherever a boundary is required, and `WsSessionDelivery`'s raw sink is private with `deliver` as its only public exit.

Two corrections to this issue's own text, both confirmed by independent review: the WebRTC path named here was already guarded through `SessionResumeBridge` — the unguarded exposure was the bare `createWsHandler` the pairing gate and transport build when no resume bridge is supplied; and six *synchronous* reply families escaped too, throwing into the carrier's inbound listener rather than as rejections, so guarding only the five Promise continuations named in the title would have left them.

[...]

PR #2303, opened 2026-08-24
fix(transport): the replay budget reaches production (ARCH-030)

woojubb:
Part of #1734 (ARCH-030). **Not `Closes`** — the comparison is on the issue and items remain.

## The defect is mine, from PR #2294

`IAttachOptions.pendingBytes` was added so a bridge attachment could carry its carrier's backpressure
reading. Measured after it landed:

```
production callers passing `pendingBytes` to a bridge attachment ....... 0
```

Only a test passed one. So the **replay path — the burst this budget exists for** — ran unbudgeted
while every other outbound path on the same connection was guarded.

**An option added and not wired is a capability that exists and is never reached**, and counting its
existence as delivery is what erases the difference. My own PR body counted it as delivered
enforcement.

It surfaced from comparing #1734's acceptance items against `git show origin/develop:<file>` rather
than against that PR body — which is the reason the post-merge step reads the tree.

## The fix, and the check for the shape it could have repeated

`session-attachment.ts` now passes the reading. `IPairingChannel` carries `bufferedAmount` as an
optional structural member, for the reason `IDataChannelSink` already documents: the slice is
structural and a double may omit it, while a real data channel always has one.

Checked before calling it done — `session-attachment.ts:51` is the **only** bridge attach in
production, so this is not one site fixed with others left behind.

## The test asserts behaviour, not wiring, and that distinction is the point

**A case checking that the option reaches `bridge.attach` would have passed on the day this defect
existed**, because the option was reaching nothing on the way in. Asserting the argument arrives
proves the argument arrives.

What can fail is a peer over budget going unclosed. The mutant that removes the wiring produces
exactly that:

```
AssertionError: expected [ …(3) ] to have a length of +0 but got 3
```

Three frames replayed to a peer over its limit, where none should have been. Re-run after the rebase
rather than reused.

A companion case asserts a reading peer still receives its whole tail — without it, a budget that
refused **everything** would also pass.

## Verification

`pnpm harness:verify-like-ci` exit 0, all 12 stages, on base `d39c9e129`; 143 scans; 322 tests across
the two packages.

## What remains on #1734 — judged, and posted on the issue

- **Queue budget: declined.** The boundary holds no queue — measured, zero array operations in
  `outbound-delivery.ts`. The *carrier* queues, and `bufferedAmount` is that queue's reading. A queue
  budget would mean adding a second place that holds frames in front of the one that already does.
  Reverses if a reason to hold frames is ever proposed (retry, reorder, coalesce); none is.
- **Time budget: real, and filed separately as #2306.** A peer reading arbitrarily slowly stays under
  a byte budget forever. `bufferedAmount` is a quantity, not a history, so no threshold over it can
  tell 1 MiB held for 20ms from 1 MiB held for 20 minutes. Different mechanism, so it is not folded
  into this issue's scope.
- **Drain policy: dependent on the queue, therefore also declined.** With no queue there is nothing to
  drain; the carrier drains itself and overflow is terminal by design.

The issue is **not closed on my judgement** — a decline is a proposal back to the scope's author, not
a disposition to apply to someone else's issue.

### DIFF
--- a/packages/agent-transport-webrtc/src/__tests__/session-attachment-budget.test.ts
+++ b/packages/agent-transport-webrtc/src/__tests__/session-attachment-budget.test.ts
@@ -1,0 +1,95 @@
+/**
+ * ARCH-030 / issue #1734 — the replay path's backpressure budget REACHES production.
+ *
+ * `IAttachOptions.pendingBytes` existed with zero production callers: the bridge could take a
+ * carrier's backpressure reading and nothing gave it one, so the replay burst — the case the
+ * reopened scope names — ran unbudgeted while every other outbound path on the same connection was
+ * guarded.
+ *
+ * This case asserts the BEHAVIOUR, not the wiring. A test that checked the option reaches
+ * `bridge.attach` would have passed on the day the defect existed, because the option was reaching
+ * nothing on the way in. What can fail is a peer over budget going unclosed.
+ */
+import {
+  DEFAULT_MAX_PENDING_BYTES,
+  SessionResumeBridge,
+} from '@robota-sdk/agent-transport-protocol';
+import { createTestInteractiveSession } from '@robota-sdk/agent-interface-session/testing';
+import { describe, expect, it } from 'vitest';
+
+import { attachSession } from '../session-attachment.js';
+
+import type { IPairingChannel } from '../pairing-gate-options.js';
+import type { IInteractiveSession } from '@robota-sdk/agent-interface-session';
+
+function fakeSession(): {
+  session: IInteractiveSession;
+  fire: (event: string, arg: unknown) => void;
+} {
+  const handlers = new Map<string, (arg: unknown) => void>();
+  const session = createTestInteractiveSession({
+    on: ((event: string, handler: (arg: unknown) => void) =>
+      handlers.set(event, handler)) as IInteractiveSession['on'],
+    off: ((event: string) => {
+      handlers.delete(event);
+    }) as IInteractiveSession['off'],
+  });
+  return { session, fire: (event, arg) => handlers.get(event)?.(arg) };
+}
+
+/** A channel whose backpressure the case controls, matching the optional structural slice. */
+function channelHolding(bytes: () => number): IPairingChannel & { sent: string[] } {
+  const sent: string[] = [];
+  return {
+    sent,
+    send: (data: string) => void sent.push(data),
+    close: () => undefined,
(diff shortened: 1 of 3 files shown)


woojubb:
## Merging on a verdict whose base has moved — what was verified by hand, and what would have changed it

The verdict names this pull request's current head and `ACTIONABLE FINDINGS: 0`, but its base has
moved twice since it was written. `merge-gate` cannot reach past that, so this is an override, and an
override is only honest if it says what was checked in the gate's place.

### What I verified

**1. Non-overlap, at package level rather than file level.**

```
landed under me:   7ee0ca6e4  HARNESS-118 record
                   a0b7891ea  ARCH-044 subagent wire payload
                   77acaf306  CLI-078 record
their packages:    agent-framework, agent-subagent-runner
my package:        agent-transport-webrtc
shared files:      none
shared packages:   none
```

File-level non-overlap alone would not settle it — two changes can share a package and interact
through it without sharing a file. These share neither.

**2. The content this override covers is the content the verdict read.** Rebased locally onto the
current `develop` and compared `git patch-id --stable` before and after:

```
1c9610a28cfd645ded61283b5eaa1976b422444d   before
1c9610a28cfd645ded61283b5eaa1976b422444d   after
```

Identical. So the local re-verification below is of this branch's own three files, on the base it will
actually merge onto — not of a different tree that happens to be nearby.

**3. Re-verified at the true base, locally, without pushing.** A rebase with zero content change is
not a finding to resolve, and pushing it would spend a CI cycle to re-read a byte-identical diff —
and would start another race with a base that has now moved three times while this was open.

**4. Required checks by name, from the branch ruleset rather than the aggregate.** All 11 pass. The 5
`SKIPPED` entries are promotion-only contexts, not in the required set for a `develop` pull request.

### What would have changed the decision

- **Any shared file or shared package** with the three intervening commits. There is none; if there
  were, the local re-verification would not be enough, because a passing suite does not establish that
  two independent changes compose.
- **A patch-id that moved across the rebase.** That would mean the verdict read different content
  from what merges, and no amount of local green would cover it.
- **A failing required check**, or a non-zero finding count. Neither.
- **A verdict naming a head other than this one.** The staleness here is entirely in the base; the
  head matches.

### What this does not claim

It does not claim the gate was satisfied. It was not — the base is stale by two commits and the gate
is right to say so. It claims the specific thing the gate could not check was checked by hand, and
names it, so a reader can disagree with the reasoning rather than having to trust it.

Issue #2045, opened 2026-08-22
P1 contract: agent-transport-protocol has no total runtime message decoder


woojubb:
## Summary

`agent-transport-protocol` owns the TypeScript message unions but does not own total runtime decoding. Server and clients trust `JSON.parse(...) as TClientMessage/TServerMessage`, and routing checks mostly inspect only the discriminator.

Valid JSON with an invalid shape can therefore cross the wire boundary as a typed protocol value.

## Evidence

- [Protocol unions are owned in `ws-protocol.ts`](https://github.com/woojubb/robota/blob/15800bf0f0023406cf06981a7afcc2cfdcd17f63/packages/agent-transport-protocol/src/ws-protocol.ts#L32-L124).
- [Server routing accepts parsed values without total field validation](https://github.com/woojubb/robota/blob/15800bf0f0023406cf06981a7afcc2cfdcd17f63/packages/agent-transport-protocol/src/ws-handler.ts#L77-L144).
- [Resume bridge repeats the assertion boundary](https://github.com/woojubb/robota/blob/15800bf0f0023406cf06981a7afcc2cfdcd17f63/packages/agent-transport-protocol/src/session-resume-bridge.ts#L136-L151).
- [GUI client](https://github.com/woojubb/robota/blob/15800bf0f0023406cf06981a7afcc2cfdcd17f63/packages/agent-transport-gui/src/client/ws-session-client.ts#L61-L76) and [WebRTC web client](https://github.com/woojubb/robota/blob/15800bf0f0023406cf06981a7afcc2cfdcd17f63/packages/agent-transport-webrtc-web/src/client/rtc-responder-gate.ts#L108-L119) perform their own unchecked casts.

Fields such as `prompt`, `taskId`, `seq`, and nested background payloads are not exhaustively validated. Arrays, empty objects, wrong primitive types, and malformed nested values can reach typed APIs or corrupt ACK/replay behavior.

## Proposed direction

Add owner-side `decodeClientMessage` and `decodeServerMessage` codecs that validate every variant and nested field. Carriers should only implement `raw bytes -> owner decoder -> typed message` and use one consistent `protocol_error`/client error result.

## Acceptance

- Every union variant has encode/decode round-trip coverage.
- Null, arrays, missing fields, wrong primitives, malformed nested values, and oversized payloads fail consistently.
- WS and WebRTC carriers share the same contract corpus.
- No `JSON.parse(...) as TClientMessage/TServerMessage` remains outside the owner codec.
[...]

PR #2307, opened 2026-08-24
fix(protocol): refuse a submit prompt that is not a string (TRANS-008)

woojubb:
Refs #2045. **Deliberately `Refs`, not `Closes` — see "What this does not decide" below.**

## The one measured crossing

The submit handler's guard was `if (!msg.prompt)` — falsy, not typed — and `parseClientMessage` does `JSON.parse(data) as TClientMessage` without decoding. So `{}`, `[]`, `42` and `true` all passed it and reached `session.submit(input: string)`.

**Why this field and not the other thirteen carrying payloads: the value does not stop at the session.**

```
execution-controller:249   emit('user_message', displayInput ?? input)
ws-session-events.ts:47    user_message: 'forwarded'
ws-session-events.ts:104   deliver({ type: 'user_message', content, … })
ws-protocol.ts:70          { type: 'user_message'; content: string; … }
```

It is delivered to **every attached client, not only the sender**, inside a frame whose `content` is declared `string` — and both client cast sites (`ws-session-client.ts:66`, `rtc-responder-gate.ts:118`) accept it with `as TServerMessage`. One unguarded field let a client put a shape-invalid value into a typed frame delivered to its peers.

## Measured at `b9081fe7c` before designing

`TClientMessage` has **24 variants, 14 with payload fields, and exactly 2 ad-hoc guards** — both falsy checks. So an unvalidated payload reaches every handler. **What the consumers do with it is a separate question, and asking it is what narrowed this:**

| path | outcome |
| --- | --- |
| `permission-response` | **No harm.** `interpretApproval` compares by strict equality against `'allow-session'`/`'allow-project'` and otherwise returns `allowed: result === true`. Every malformed value denies — **a decoder here would prevent nothing.** |
| `ack` / `resume` | **Self-inflicted, bounded.** A non-numeric seq drops nothing; a huge one forces the sender's own full refresh. The buffer is separately bounded by frames and bytes. |
| `submit.prompt` | **Crosses to peers.** Fixed here. |

## Verification

Red proof is its own commit: 4 cases fail, 2 pass as controls.

| mutant | result |
| --- | --- |
| the falsy guard restored | red |
| the type half alone, emptiness dropped | red — caught by the empty-prompt control |
| refuse everything | red — caught by the accepts-a-string control |

The controls are load-bearing in both directions: without the empty-prompt case a type check would silently admit `''`; without the accepts-a-string case a handler that refused everything would satisfy all four positive cases.

**One correction from inside this work.** My first mutant run reported all three surviving. That was my harness, not the code — a command substitution inside the `echo` string clobbered `$?` before it was read, so the exit code belonged to a different command. Re-run reading the status directly, all three die. Recorded because it is the same defect this repository has been finding all day: **a number read from the wrong place looks like a measurement.**

The error message changed from `prompt is required` to `prompt must be a non-empty string`, and one existing assertion was updated. The old wording was false for a prompt that *was* provided and had the wrong type.

144 scans pass, `agent-transport-protocol` 187 / `agent-transport-gui` 21 / `agent-cli` 441, typecheck 0, `pnpm lint` exit 0.

## What this does not decide

**Issue #2045 stays open.** It proposes a total runtime decoder for the package; this closes the one crossing I could measure. Whether the package should decode totally is a data-correctness disposition and belongs to the owner, not to this merge. **A merge that closed the issue would read as that decision having been made.**

Three things stated so a later reader is not misled by the green:

- **`rtc-responder-gate.ts` is untouched and still casts.** It is a second client transport with the same hole, in a package #2045 never names — a decoder scoped to the issue's file list would leave it defective while reading as complete.
- **Thirteen of the fourteen payload-carrying fields were not traced to their use sites.** The ones the issue names were. If another crosses the way `prompt` does, that is a finding for #2045, not a widening of this diff.
- The record carries the same three statements, not only this description.

### DIFF
--- a/.agents/tasks/TRANS-008-a-truthy-non-string-submit-prompt-is-re-broadcast-to-every-attached-client-as-a-.md
+++ b/.agents/tasks/TRANS-008-a-truthy-non-string-submit-prompt-is-re-broadcast-to-every-attached-client-as-a-.md
@@ -1,0 +1,103 @@
+---
+title: 'TRANS-008: a truthy non-string submit prompt is re-broadcast to every attached client as a typed frame'
+issue: https://github.com/woojubb/robota/issues/2045
+status: todo
+created: 2026-08-24
+priority: medium
+urgency: soon
+area: packages/agent-transport-protocol
+depends_on: []
+---
+
+# TRANS-008: a truthy non-string submit prompt is re-broadcast to every attached client as a typed frame
+
+## Objective
+
+A `submit` frame whose `prompt` is a truthy non-string reaches the session, is re-emitted as a
+`user_message`, and is delivered to every attached client inside a frame whose `content` is declared
+`string`. One unguarded field lets a client put a shape-invalid value into a typed frame delivered to
+its peers.
+
+## Measured at `b9081fe7c`, before designing
+
+Issue #2045 proposes a total runtime decoder for `agent-transport-protocol`. Its premise holds — the
+package owns the unions and decodes nothing, and there are **three** cast sites, one more than the
+issue lists:
+
+```
+ws-handler.ts:92                     JSON.parse(data) as TClientMessage
+agent-transport-gui/…:66             JSON.parse(data) as TServerMessage
+agent-transport-webrtc-web/…:118     parsed as TServerMessage        (not named in the issue)
+```
+
+`TClientMessage` has 24 variants, 14 carrying payload fields, and exactly **2** ad-hoc guards —
+`!msg.prompt` and `!msg.name`, both falsy checks. So an unvalidated payload reaches every handler.
+
+**What the consumers do with it is a separate question, and asking it changed the scope.**
+
+- **`permission-response` — no harm.** `msg.result` reaches `resolvePermission` unvalidated, but
+  `interpretApproval` (`agent-session/src/abortable-approval.ts:61`) compares by strict equality
+  against `'allow-session'` / `'allow-project'` and otherwise returns `allowed: result === true`.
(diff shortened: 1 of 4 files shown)


woojubb:
**Merging on a verdict that names an older base, with the reasoning recorded rather than implied.**

The verdict names `REVIEWED BASE: b9081fe7c`; live `origin/develop` is now `48b9f6f92` (#2296 landed while CI ran). By the rule I have been applying all day — *the verdict ages, not the field* — that verdict is stale, and the honest default is to rebase and re-request.

I did not, for two measured reasons:

**The intervening commit cannot interact with this one.** #2296 touched `agent-command/docs/SPEC.md`, a scenario record, a scenario helper, an example and one new test — all in `agent-command`. This change touches `agent-transport-protocol/src/ws-handler.ts` and its tests. **No shared file and no shared package.**

**I re-ran the full verification at the new base rather than arguing from non-overlap.** Rebased locally onto `48b9f6f92`: build and typecheck clean, `agent-transport-protocol` 187 tests pass, 144 scans pass. That is direct evidence the reviewed diff and the shipping diff behave identically, which is what the freshness rule is protecting.

I also did not push the rebase. `pre-push-check` correctly refuses a push to a PR carrying `ACTIONABLE FINDINGS: 0`, and a rebase with zero content change is not a finding to resolve. Overriding the hook to obtain a re-read of an identical diff would spend a CI cycle to reach the same answer.

**What would have changed this:** any shared file, any shared package, or a failure in the re-run at the new base. Recorded so the choice is checkable rather than taken on trust.

Issue #2228, opened 2026-08-23
A SPEC's advertised export is proved by a comment that says the export is not there


woojubb:
`check-spec-public-surface.mjs` decides whether a SPEC's advertised export is real by asking whether the
name appears anywhere in the package's `src` — with comments included in the corpus.

So a comment that says the symbol is **not** here counts as proof that it **is**.

Concretely, before ARCH-108 (issue #2113), `packages/agent-interface-transport/src/index.ts:71` read:

```ts
// HARNESS-103: `createSessionCapabilityHost` / `readSessionCapability` are NOT here. They are the
```

Both names were rows in that package's Public API table. Both had moved to
`agent-interface-session` in ARCH-106. The scan passed on both, and the sentence it accepted as
evidence is a sentence written specifically to tell a reader the opposite.

## How it was found

Not by the scan. By comparing the SPEC's table rows against the built `dist/**/*.d.ts` — what the
package actually declares. That found **6** phantom rows; the scan had reported **2**. The four it
missed were each mentioned in a comment somewhere in `src`.

Two measurements of the same quantity disagreed, and the gap was the defect.

## Why this is the wide-corpus class, not the narrow one

Same family as the `examples/` finding: a guard whose corpus is too wide passes QUIETLY, while one
whose corpus is too narrow fails loudly and gets investigated. Here the excess corpus is comments —
a region of the file that can assert anything, including the negation of what the check is testing.
Prose is the worst possible evidence for a mechanical claim, because it is the one part of a source
file that is free to be about the symbol rather than to be the symbol.

Related: issue #2215 (a guard whose green strengthens as its subject disappears), issue #2221
(`export *` makes a symbol invisible to the interface guards).

## Repair

Require the export positively rather than searching text: resolve each advertised row against the
package's real declared surface — the built `.d.ts` for every entry declared in `package.json`
`exports`, or a parse of the barrel's export statements. A row then has to be a genuine export to
pass, instead of merely being mentioned.

Note the second half: the corpus must cover **every** declared entry, not just the root. Transport
publishes a `/testing` subpath, and a check reading only `dist/node/index.d.ts` would trade this
defect for the narrow-corpus one.

## Falsification

The current scan cannot be tested for this: deleting a real export while leaving a comment naming it
keeps the scan green. That test is the deliverable — it must go red before the repair and green
after, or the repair is unverified.

woojubb:
## The same defect, from the opposite direction — it is a class, not one odd case

[...]

PR #2324, opened 2026-08-24
fix(config): an unreadable org policy no longer reads as no policy (issue #2023)

### LINKED ISSUE #2023: P1 security: malformed layered settings fail open and silently drop permission deny rules
## Summary

The layered settings loader treats unreadable JSON, schema-invalid content and an empty file the same as a missing file. Startup then continues with defaults or other layers. If a user relies on a project, environment, policy or user layer to deny dangerous tools, corruption or a partial write silently removes that protection and may allow an operation that should have been blocked.

This corresponds to the already recorded internal task CONFIG-002 and needs a public defect/security record.

## Evidence

- [readSettingsFile catches parse/read failures and returns an empty object](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/config/config-loader.ts#L49-L63).
- [All parsed layers are later merged and defaults are applied](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/config/config-loader.ts#L267-L285).
- [Default permission lists are empty](https://github.com/woojubb/robota/blob/12e146e53d9d1b54b492b064c2b00b6b23e5be3e/packages/agent-framework/src/config/config-loader.ts#L32-L42).

A safe reproduction wrote malformed JSON to a project settings file. Loading did not throw and the resolved deny list was empty.

## Security impact

- Fail-open loss of deny rules after truncation, merge conflict, concurrent write or disk corruption.
- Operators receive no actionable indication that their intended security policy was ignored.
- Different configuration owners may behave inconsistently if they parse files separately.

## Required improvement

Make missing, unreadable, invalid JSON and schema-invalid settings distinct outcomes. A present but invalid security-bearing layer must stop initialization with a typed error naming the source path and validation problem. Centralize parsing for all configuration layers so every consumer gets identical behavior. Writers must use owner-safe atomic replacement and preserve restrictive modes. If recovery is offered, it must require an explicit user action and must not silently continue with weaker permissions.

## Acceptance tests

- Invalid JSON, empty files, wrong field types and unreadable files fail startup for every layer.
- A missing optional layer remains supported and is distinguishable from an invalid one.
- Deny rules cannot disappear without a visible fatal error.
- Concurrent/partial write simulations never expose an empty policy.
- Interactive, print, serve and headless modes return consistent non-zero failures.

woojubb:
Closes #2023.

## The measurement that decided everything downstream

**What does `null` mean to the consumers?** Measured before designing anything — all four enforcement
sites:

```
provider-command-profile-operations.ts:41    orgPolicy?.allowedProviders && !…includes(profileName)
provider-command-profile-operations.ts:123   orgPolicy?.requireApiKeyFromEnv && isApiKeyPlaintext(…)
interactive-session.ts:970                   this.orgPolicy?.blockedCommands?.includes(name)
interactive-session-host-actions.ts:114      orgPolicy?.allowedProviders && !…includes(…)
```

**`null` means "no restrictions" at every one.** And `loadOrgPolicy()` returned `null` for a missing
file *and* for a corrupt one.

So a policy file truncated by a partial write, or mangled by a sync tool, produced behaviour
**identical to having deployed no policy at all** — provider allowlisting, blocked commands and the
plaintext-key requirement removed together, silently, with an administrator unable to tell the two
apart. The call site says it in the repository's own words: consumers *"read absence as no policy
configured"*.

The loader had been unreachable until CLI-083 wired it three hours before this work. It is reachable
now.

## The comment being replaced was an argument, not a mistake

> `// allow-fallback: malformed org-policy.json must not crash CLI startup`

Whoever wrote that was thinking about availability, and the fix answers it rather than deleting it:

- **missing** file → `null`, no throw. That is the common case and the one the comment protected.
- **present but unreadable** → throws. That is a file an administrator deployed and cannot be read,
  which is their error to see.

`packages/agent-framework/src/config/settings-io.ts` (CLI-069) and the trusted-device store answer the
same question the same way, **one and two files over**. This loader was the one that did not.

## Found while measuring, and worse than the filed defect

`JSON.parse(raw) as IOrgPolicy` asserted a shape it never checked. `{"allowedProviders": "anthropic"}`
— a string where an array belongs — is truthy, so the guard **fires**, and then:

```
"anthropic".includes("ant")  →  true
```

**A profile named `ant` passes an allowlist naming only `anthropic`.** A policy that mis-enforces is
worse than one that does not load, because it looks like it is working. Valid JSON of the wrong shape
is now refused.

Checked structurally rather than with a schema: `IOrgPolicy` is a hand-written interface and a schema
in the loader would make the loader the type's source of truth as a side effect of validating it.

## Red-first, and what the two passing cases are for

Four of six cases fail against the previous loader. The two that pass — a missing file returns `null`,
a well-formed policy is accepted unchanged — describe behaviour that did not change, and **without
them a loader that threw on everything would also pass.**

## What this does NOT change, stated because it is a real cost

The throw propagates from `buildCommandSetup`, called outside any try in
`packages/agent-cli/src/cli.ts`, so it reaches the user through the top-level handler **with a stack
trace around it**. That is exactly the exposure `SettingsParseError` already has from the same
function, so this is consistent rather than novel — but an administrator with a typo now sees a stack
trace, and presenting these two readably is one improvement covering both.

## Three gates caught this change, and one of them was reading me wrong

- **`spec-public-surface`** — the new runtime export was undocumented. Documented.
- **`no-fallback`** — read my prose **quotation** of the `allow-fallback` token as a declaration of
  one. The same text-matching-cannot-see-mood class as #2228, this time against me. Reworded.
- **`lint-ceiling`** — four uses of `unknown`. The rule's own message names `TUniversalValue` as the
  SSOT type and says type guards belong at boundaries, which is what this is.

## Verification

`pnpm harness:verify-like-ci` exit 0, all 12 stages, on base `2572b4b62`; 143 scans; `pnpm lint` exit
0.


woojubb:
## MUST accepted and fixed in `5e4901f2e` — and it was right about something I had not measured

The finding is correct, and the part worth recording is *why* my counter-argument failed rather than
that it did.

I wrote in the pull request body that the throw's exposure matched `SettingsParseError`'s and was
therefore consistent rather than novel. **That is true and it is not the measurement that mattered.**
The finding measured something I did not: `packages/agent-cli/src/cli.ts` has an established pattern
for terminal startup errors, and it sits **three lines above** the call I was changing —
`resolveShellPreset` wrapped in `try { … } catch { …is terminal — exit is the correct response }`.

I had read line 75, where `parseCliArgs` does the same thing, and filed it as argument validation. **A
pattern already seen, under the wrong heading, is invisible in a way an absent one is not** — I would
have found it by searching for one I believed missing.

## The fix

`buildCommandSetupOrExit` catches **only** `OrgPolicyParseError`, writes the message, exits 1.

Narrow deliberately: a broad catch turns every unrelated startup defect into a clean exit wearing
someone else's message, which is a worse failure than the one being fixed.

Placed beside `buildCommandSetup` rather than in the shell entry point, for two reasons where the
second survives the first being removed:

- `file-size` freezes that file at 441 lines and refused the inline version at **458**;
- the presentation of a setup failure belongs next to the setup rather than in the shell that calls
  it.

## The companion test's first version asserted nothing, and that is the useful part

Two cases: a corrupt policy is presented and exits 1; an **unrelated** failure still propagates.
Without the second, a broad catch passes the first.

Its first version used a nonexistent cwd to produce the unrelated failure — **and a nonexistent cwd
does not fail at all.** So the case asserted nothing, and the broad-catch mutant survived it: the
exact mutant that case exists to kill.

Probed what `buildCommandSetup` actually throws for an unrelated reason — `SettingsParseError`, from a
corrupt user settings file beside a good policy — and rewrote around it. Both mutants now die:

```
guard re-throws instead of presenting  → case 1:  expected 'exited' but got 'Organization policy file …'
guard catches broadly                  → case 2:  expected /invalid JSON/ but got 'exited'
```

Before this commit, four existing cases were green **with no guard at all**. Nothing asserted the
presentation.

## One thing deliberately not fixed here

`packages/agent-cli/src/cli.ts:158` calls `readSettings(getUserSettingsPath())` outside the same
`try`, so `SettingsParseError` carries the identical exposure this finding is about. **It is not what
this finding named**, and an improvement noticed while reading is not a ground to push. Filed
separately.

Verified on base `72c69f371`: `verify-like-ci` exit 0 across all 12 stages, 144 scans.

woojubb:
## `PRE_PUSH_ALLOW_FROZEN_DIFF=1` used — ground: **rebase**, and why the alternative was unavailable

`pre-push-check` refused this push because the latest verdict reads `ACTIONABLE FINDINGS: 0`. That
refusal is correct about what it can see: a diff changing after a clean verdict. It cannot see whether
the change is a rebase, which is one of the three named grounds.

### It is a rebase, verified rather than asserted

```
pushed head's merge-base with develop .... 72c69f371
this head's merge-base .................. c1c3ac079
pushed patch-id ......................... 93f843c87c56c3bdb008
this patch-id ........................... 93f843c87c56c3bdb008   identical
```

**My first attempt at this check got it wrong**, and it is worth saying how: I used the verdict's
stated `REVIEWED BASE` as the pushed head's base. It was not — the pushed head's own merge-base was
`72c69f371`. Wrong base, wrong answer, and it reported a content change that had not happened. The
same class of mistake as reading three different bases as three answers to one question.

### Why not simply merge on the existing verdict

That is the cheaper route and it was measured first. It requires establishing that the intervening
commits do not interact with this branch — a passing local suite does not establish that two
independent changes compose. **The measurement refused it:**

```
shared files:    packages/agent-framework/src/index.ts
                 packages/agent-framework/docs/SPEC.md
shared package:  packages/agent-framework
```

`818f0c82c` (`createSession is internal again`) touched the barrel and the SPEC this branch also
touches. So the override route was closed, a rebase was required, and a fresh verdict at the current
base is the only thing that can carry this to merge.

### And the rebase was checked for the failure a clean rebase hides

A clean rebase means no conflict, not that both sides survived. Either taken whole would leave a green
tree with the other silently gone:

```
this branch's OrgPolicyParseError export ... alive, index.ts:110
their createSession removal ................ held, absent from the barrel
```

### What would have changed the decision

- **A patch-id that moved** — then it is not a rebase, the ground evaporates, and this waits for
  #2324 to land and goes into a second pull request.
- **Non-overlap holding** — then no push at all: merge on the existing verdict with the reasoning
  recorded, which is the cheaper path and the one taken when it was available.
- **A finding on this pull request** — then the ground is the finding, and the rebase is incidental.

Verified at `c1c3ac079`: `verify-like-ci` exit 0 across all 12 stages, 144 scans, 12 tests.

PR #2341, opened 2026-08-25
fix(harness): a parenthesised M/D date is not a completion ratio (HARNESS-122)

### LINKED ISSUE #2339: progress-report-quantification reads an M/D date as a partial completion ratio
## Summary

`scan-progress-report-quantification.mjs` fires on a **date** written `M/D` when a completion word sits next to it. Measured at `e5551e9b6`:

```
ratio 8/14 reported without a percentage:
  "ARCH-011은 완료(8/14)입니다."
```

`8/14` is 14 August — the `completed:` date in `ARCH-011`'s frontmatter. It is not a progress ratio, and there is no countable work set anywhere in the sentence.

## Why the scan is right to fire and still wrong

The scan is deliberately mechanical and says so in its header — it does not attempt to decide whether a sentence *is* a mid-work update, only that a partial ratio (`N/M`, `N < M`) in a completion context without a percentage is a violation. `8 < 14` holds, and the Korean word `완료` ("completed") is immediately before the parenthesis, so the completion context is real by any lexical test.

The header also enumerates the false-positive classes it suppresses explicitly, each covered by fixtures:

> identifier lists (`ARL-04/05/06/07`, `TC-01/04`), step/round references (`Step 2/3`), decimal scores (`7.7/10`), line references (`lines 54/146`) and completed results (`45/45 scans pass`)

**`M/D` dates are not among them.** This is an unlisted class, not a misfire of a listed one.

## Why Korean makes it reachable where English mostly does not

English writes "completed on 8/14", putting a preposition between the completion word and the date. Korean writes `완료(8/14)` — the completion word is adjacent to the parenthesis with nothing in between. **The suppression classes were derived from English-language transcripts**, and the repository's own language policy has agents narrating to the owner in Korean, so this class is reachable in normal use and was not reachable in the sample the classes came from.

## What was deliberately not done

**No acknowledgment was written**, and the scan is left red on the affected host.

`progress-report-acknowledgments.json` states its own contract: each entry *"names a real violation of the quantified-progress rule that has already happened."* This is not one. Writing an entry to clear a red scan would put a false statement in a ledger whose only value is that its entries are true — the ledger would still read green, and the next reader could no longer trust that an entry means what it says.

A scan being wrong visibly is a better state than a ledger being wrong invisibly.

## Direction

A suppression for a date-shaped `N/M` — bounded by plausible month/day ranges, or by an adjacent date context — plus a fixture carrying the literal line above, in the same form as the five existing suppression classes.

Worth deciding at the same time whether the suppression vocabulary under `.agents/harness.config.json` → `progressReportQuantification` needs any non-English cases beyond this one, since the same derivation gap would produce them.

Found while running `pnpm harness:scan` before a push at `e5551e9b6`.

woojubb:
## What

`scan-progress-report-quantification` reported a **date** as a progress ratio:

```
ratio 8/14 reported without a percentage:
  "ARCH-011은 완료(8/14)입니다."
```

`8/14` is 14 August — `ARCH-011`'s `completed:` date. No countable work set appears in the sentence.

Closes #2339.

## Why it was urgent

`pre-push.mjs` refuses on any red scan, and the finding lives in an **append-only transcript** — so it was red for every push from the affected host, on every branch, indefinitely. It blocked a lane rather than annoying one.

**The fix self-unblocks:** this PR's own push is the first one that could pass.

## Why the scan was right to fire

It deliberately does not judge whether a sentence *is* a mid-work update — only that `N/M` with `N < M` in a completion context and no percentage is a violation. `8 < 14` holds, and `완료` sits immediately against the parenthesis.

**Korean is what makes the class reachable.** English writes "completed on 8/14", with a preposition between the completion word and the date. Korean writes `완료(8/14)`, adjacent. The five suppression classes already in the engine — identifier lists, step references, decimal scores, line references, completed results — were derived from English transcripts, so this class was **not reachable in the sample they came from.**

## The suppression, and why all three clauses are needed

```js
const parenthesised = /\(\s*$/.test(before) && /^\s*\)/.test(after);
if (parenthesised && completed >= 1 && completed <= 12 && total > 12 && total <= 31) continue;
```

- **the parenthesis holds the ratio and nothing else** — `(3/14 done)` is a progress note and still fires;
- **first component is a possible month**;
- **second is above 12 and at most 31** — this is the clause that keeps the class narrow. `완료(3/7)` could be three of seven, so it still fires; `(8/14)` cannot be read as anything but a date.

**Range alone would be useless:** `3/7 done` is the scan's commonest true positive and both of its components are in date range.

**Stated residual:** a date whose day is also ≤ 12 (`완료(8/9)`) still reports. Left firing rather than guessed at — that errs toward reporting, which is the safe side for a guard.

## Why not configuration

`.agents/harness.config.json` → `progressReportQuantification` carries `transcriptRoot`, `enforceSinceIso`, `completionKeywordPattern`, `identifierNounPattern`, `identifierNounSuffixPattern`. No date class. The near-miss fails: `완료` is a **completion keyword**, not an identifier noun, so listing it there would suppress genuine `완료 8/14` violations — trading a false positive for a false negative in the class the scan exists to catch.

## Verification — by mutation, not by the tests passing

Five cases added: one suppression, four positive controls. Disabling the guard (`if (false && …)`) kills **exactly one**:

```
mutant:   1 failed | 53 passed   ← the failure is the suppression case
baseline: 54 passed
harness:scan: 144 passed, 1 skipped, 0 failures
```

So the suppression case measures the guard, and the four controls measure that it does not over-reach — `완료 8/14`, `완료(3/7)`, `(3/14 done)` and `완료(13/28)` all still fire with the guard on.

## What was deliberately not done

**No acknowledgment entry.** `progress-report-acknowledgments.json` states its contract: each entry *"names a real violation of the quantified-progress rule that has already happened."* This is not one, and a ledger for real violations stops meaning anything the first time it absorbs a false positive.

**No `pre-push` hatch.** `pre-push.mjs` has no form of "this scan is wrong and I have verified why" — recorded on #2339 as the reason the fix had to be the fix. A general override would become the answer to every red scan the moment it existed; if one is ever right it needs its own decision, not a slot in a PR about a date regex.

## Gate

`.agents/spec-docs/draft/HARNESS-122-…` carries the Evidence Log including the owner's GATE-APPROVAL sign-off, quoted. A peer session's instruction to take this path set the order of the work and was explicitly not treated as that approval.

https://claude.ai/code/session_01C4KcLfy15wxYjKpaUduAMH

### DIFF
--- a/.agents/spec-docs/draft/HARNESS-122-progress-report-quantification-reads-an-m-d-date-as-a-partial-completion-ratio.md
+++ b/.agents/spec-docs/draft/HARNESS-122-progress-report-quantification-reads-an-m-d-date-as-a-partial-completion-ratio.md
@@ -1,0 +1,89 @@
+---
+status: draft
+type: INFRA
+tags: [infra, harness, scan]
+---
+
+# HARNESS-122: an M/D date is read as a partial completion ratio
+
+## Problem
+
+`scan-progress-report-quantification` reports a violation on a **date**:
+
+```
+ratio 8/14 reported without a percentage:
+  "ARCH-011은 완료(8/14)입니다."
+```
+
+`8/14` is 14 August, `ARCH-011`'s `completed:` date. No countable work set appears in the sentence.
+
+The scan fires correctly by its own rule — it deliberately does not judge whether a sentence _is_ a
+mid-work update, only that `N/M` with `N < M` in a completion context and no percentage is a
+violation. `8 < 14` holds and `완료` is adjacent to the parenthesis.
+
+**The consequence is a total block on one lane.** `pre-push.mjs` refuses on any red scan; the finding
+is in an append-only transcript, so every push from the affected host fails on every branch until the
+class is handled. `pre-push.mjs` has no override for a scan the author believes is wrong.
+
+## Prior Art Research
+
+Waived: the subject is a false-positive class in a scan this repository wrote, over transcripts
+this repository produces, against a rule this repository states (`agent-conduct.md`, HARNESS-026).
+No external product documents a Korean-adjacent completion-word/date-parenthesis interaction with a
+progress-quantification rule, because no external product has that rule. The evidence that decides
+the design is in this repository — the scan's five existing suppression classes and their fixtures —
+and it is read directly in the Solution below rather than approximated from an external source.
+
+## Solution (draft direction)
+
+Add a **date class** to the engine's suppressions, beside the five already there (identifier lists,
+step/round references, decimal scores, line references, completed results).
+
+The decidable form: an `N/M` whose components fall in date ranges (`N ≤ 12`, `M ≤ 31`) and which sits
+in a date-shaped context — parenthesised, or adjacent to a date token. **Range alone is not enough**:
+`3/7 done` is a genuine violation with both components in range, and a suppression keyed only on
+magnitude would retire the scan's most common true positive.
+
+This cannot be configuration. `.agents/harness.config.json` → `progressReportQuantification` carries
+`transcriptRoot`, `enforceSinceIso`, `completionKeywordPattern`, `identifierNounPattern`,
+`identifierNounSuffixPattern` — no date class. The near-miss fails: `완료` is a completion keyword,
+not an identifier noun, so listing it there would suppress genuine `완료 8/14` violations.
+
+**No `pre-push` hatch is proposed.** A general "this scan is wrong" override becomes the answer to
+every red scan the moment it exists. If one is ever right it needs its own decision and its own
+evidence, not a slot in this change.
+
+## Completion Criteria (draft)
+
+- A fixture with `완료(8/14)` produces no finding.
+- **Positive control**: `완료 8/14` — same numbers, no parenthesis — still produces one.
+- **Second control**: `3/7 done` still fires; `3/7 done = 43%` does not.
+- No acknowledgment entry is added. The ledger's contract is real violations; clearing a false
+  positive through it would make an entry mean nothing.
+- `pnpm harness:scan` green on the affected host — which is this item's own unblock condition, so the
+  push carrying the fix is the first push that can pass.
+
+## Evidence Log
+
+- 2026-08-25 — Measured at `e5551e9b6`. `pnpm harness:scan`: `1 of 143 scans failed`, sole finding the
+  `8/14` line above. `pre-push` refused the push carrying `TRANS-009`/`TRANS-010`.
+- 2026-08-25 — Config inspected: five fields under `progressReportQuantification`, no date class; the
+  `identifierNounPattern` alternative rejected because it would suppress genuine `완료 8/14`.
+- 2026-08-25 — `pre-push.mjs` environment reads enumerated: remote name, remote URL, base ref, mode.
+  No override for a disputed scan. Recorded on issue #2339 as the reason the fix must be the fix.
+- 2026-08-25 — Filed as issue #2339; task record `HARNESS-122`.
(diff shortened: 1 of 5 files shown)


woojubb:
**MUST accepted — the finding is correct, and I reproduced it.** Running the exported predicate against the current branch:

```
0 findings  ←  ARCH-011은 완료(8/14)입니다.      (the intended suppression)
0 findings  ←  감사 완료(3/20)입니다.             ←  a genuine 3-of-20 progress statement, dropped
1 finding   ←  감사 완료(3/7)입니다.
1 finding   ←  감사 3/20 완료입니다.
```

The second line is the defect. **And the review's sharpest point is that my own PR text names the rule the implementation breaks:** *"range alone is not enough… a suppression keyed only on magnitude would retire the scan's most common true positive."* I wrote that about `3/7 done`, then shipped a guard that is magnitude-plus-parenthesis and believed the `total > 12` clause made it something else. It does not — it only moves which denominators are affected, and 13–31 is an ordinary range for a real work count.

**The missing test is the finding too.** All four of my positive controls kept a denominator ≤ 12 or broke the parenthesis. None exercised a genuine ratio with a denominator inside the suppressed band, which is exactly the region the guard changed. The mutation check I ran proved the suppression case measures the guard; it could not tell me the controls did not cover the guard's actual reach.

**Not pushing a fix yet, deliberately.** `완료(8/14)` and `완료(3/20)` are the same shape, and I do not think they are separable by pattern at all — the disambiguating information is the author's intent, which is not in the text. So the next revision is a design decision rather than a tightening, and narrowing the numeric band would repeat this bug with a smaller footprint.

I am taking the design question back to the owner rather than guessing at a fourth condition. Leaving this thread open until there is a fix to point at.

woojubb:
**Fixed at `fd9371f2e`, and the fix is a different approach rather than a tightening of the old one.**

The engine guard is withdrawn entirely. As said above, `완료(8/14)` and `완료(3/20)` are the same shape and the disambiguating information is the author's intent, which is not in the text — so any numeric narrowing repeats this bug with a smaller footprint.

**An acknowledgment entry now says which of two true things it asserts:**

- `violation` — the default, and what every entry written before this meant: the rule was broken, history is append-only, it is recorded rather than fixed.
- `false-positive` — the finding is not a violation, with a reason saying how the scan read it wrong.

The old ledger had one meaning, so a finding that was not a violation had no honest entry: clearing it asserted something that never happened. Both entries are now true statements, and the advisory reports the two counts separately — a violation says the rule was broken, a false positive says the *scan* is wrong, and one total reads as the first while hiding the second.

```
⚑ progress-report quantification: 5 finding(s) acknowledged — 4 real violation(s) recorded,
  not cleared by editing history; 1 finding(s) the scan read wrong, each with its reason.
```

**This is not a general suppression.** It is per-finding, it carries a reason, and the existing anti-rot is unchanged — an entry whose finding stops appearing still fails the scan.

**Verified by mutation, twice.** Ignoring an entry's `kind` kills two cases. Disabling the `kind` validation kills one — the typo case (`false-postive`), which would otherwise fall through the `?? 'violation'` default and clear a finding while counted as a violation. Baseline 55 passed; `pnpm harness:scan` 144 passed, 0 failures.

**One thing from this worth stating, because it bears on how the first attempt got through.** That attempt *was* mutation-verified — one of five cases died with the guard disabled — and the check could not see the defect, because none of its four positive controls used a denominator inside the band the guard changed. **A mutation test proves a case measures the guard; it says nothing about whether the cases cover the guard's reach.** That is recorded in the Evidence Log rather than left as a lesson in a session.

Thanks for the catch — it was a real false-negative regression in the scan's core class, and the review named the exact sentence in my own PR text that the implementation contradicted.

Issue #2433, opened 2026-08-28
gate.mjs writes a GATE-IMPLEMENT entry the plan-order scan cannot bind: the spec path token is missing, so the checkpoint commit is refused


woojubb:
## What is broken

`scripts/harness/gate.mjs judge --gate GATE-IMPLEMENT` writes the Evidence Log entry the planning checkpoint is recognised by, but the entry it writes omits the token the plan-order scan binds on: `scan-user-execution-plan-order.mjs` `completeGateImplementEntry` requires, on one line, the exact `.agents/spec-docs/todo/<basename>` (or `active/`) token beside the exact `.agents/tasks/<basename>` token, the exact `SCENARIO DRAFTED` signal, and the phrase "whole worktree". The judge's entry carries the Task path, the signal and "The whole worktree …" — never the spec path — so the checkpoint commit is refused by the pre-commit hook:

```
✗ user-execution-plan-order: checkpoint is neither the first GATE-IMPLEMENT PASS transitioning the exact Task/spec pair into in-progress nor one continuation PASS (…) on a pair already in-progress.
```

Measured on CORE-049 (issue #2350), 2026-08-29, branch `fix/2350-permission-patterns-match-by-argument-kind` at `91dc2961c`: `gate.mjs judge --gate GATE-IMPLEMENT` → 7 PASS, entry appended; `git commit` of the pair + ledger → the refusal above; adding one line naming both exact paths to the entry → accepted. HARNESS-127 (issue #2378) and HARNESS-129 hit the same shape with the guardian's prose; the writer has moved into a script and the gap moved with it.

## Why it is its own item

The entry form is declared in the scan (HARNESS-128, issue #2394 — the catalogue does not own it), and now a second writer produces it. Either the catalogue declares the tokens and both writers follow it, or `gate.mjs` reuses the scan's own matcher to verify what it wrote. Not this PR's scope.

## Acceptance

`gate.mjs judge --gate GATE-IMPLEMENT` writes an entry that `completeGateImplementEntry(body, { basename, signal })` accepts, pinned by a test that runs the scan's matcher over the judge's output; the same for the continuation form (issue #2422).

woojubb:
Fresh reproduction on 2026-08-29 while checkpointing INFRA-138 confirms the issue unchanged on `origin/develop` `9c43eee8c`:

1. `gate.mjs judge --gate GATE-IMPLEMENT` wrote a 7/7 PASS entry.
[...]

Issue #2430, opened 2026-08-28
A hooks.PermissionDecision command hook never fires in -p print mode while permissions from the same settings file take effect


woojubb:
## What was observed

While writing the user-execution scenarios for CORE-049 (issue #2350), `user-execution-scenario-author` measured on `develop` `dd46c9183`: with a user-level `settings.json` in a scratch `HOME` carrying both a `permissions` block and a `hooks.PermissionDecision` command hook, `robota -p … --session-log <replay>` applied the `permissions` rules (a deny was recorded as `Permission denied` in the session record) but the command hook never fired — nothing was written to the file the hook was told to write.

## Why it matters

Either print mode does not run permission-decision hooks (then the docs must say so and the setting should be refused, not ignored), or the hook is silently dropped by the settings-source / workspace-trust path that the same file's `permissions` block passes through. An operator relying on the hook for audit or for an external decision would see nothing and assume it ran.

## Not chased

The author did not chase it; the scenario used the server log and the session record instead. Provenance: CORE-049 PLAN run `r20260828152917`, finding 3.

woojubb:
**처분: 수정됨 (FIXED)** — 2026-09-04 일괄 이슈 정리

이 이슈의 수정이 `develop`에 반영되었습니다.

- 커밋: `00bc2a811`
- 내용: Root cause: settings HooksSchema omitted PermissionDecision/PreModelCall/PostModelCall and z.object strips unknown keys, so the hook was silently dropped at load in every mode. Added the events + Record<THookEvent,true> parity test (16 cases).

이 이슈는 227건 전체를 한 번에 분류·처리한 일괄 작업의 일부입니다. 분류 근거와 검수 결과는 로컬 원장에 기록되어 있습니다.

Issue #2429, opened 2026-08-28
A relative Read/Write/Edit filePath reaches the permission gate unresolved, so an absolute pattern cannot judge it


woojubb:
## What is missing

`Read`, `Write` and `Edit` declare `filePath` as an absolute path in their schemas, but nothing before the permission gate guarantees it: a relative argument reaches `matchesPattern` as written, and an absolute pattern such as `Read(/w/**)` cannot be compared with `src/x` without knowing the tool's `cwd`. CORE-049 (issue #2350) makes that case **unevaluable** at the gate (a deny prompts; an allow does not auto-approve) rather than guessing.

## What it should do

Decide where a relative `filePath` is resolved against the tool's `cwd` — in the tool's argument normalisation before the gate sees it, so the gate compares canonical absolute paths — and state it in `packages/agent-tools`'s SPEC. `path-guard.ts` (`isPathInside`, `:51-54`) already confines execution to `cwd`; the permission decision needs the same canonical input.

## Provenance

Named by the `proposal-reviewer` of CORE-049 (2026-08-29) as a separate root: the gate should not guess the base of a relative path.

woojubb:
**처분: 수정됨 (FIXED)** — 2026-09-04 일괄 이슈 정리

이 이슈의 수정이 `develop`에 반영되었습니다.

- 커밋: `17d9a4973`
- 내용: Read/Write/Edit resolve relative filePath against cwd; session wrapper canonicalises path-kind args before gate/hooks/tool. 8 tests green. branch-guard refuses hooks bypass. patch-M0-2429.diff.

이 이슈는 227건 전체를 한 번에 분류·처리한 일괄 작업의 일부입니다. 분류 근거와 검수 결과는 로컬 원장에 기록되어 있습니다.

Issue #2428, opened 2026-08-28
A malformed permission pattern is refused only at the gate, never where the operator wrote it


woojubb:
## What is missing

A permission pattern the gate cannot interpret is learned about only when an invocation reaches the gate. CORE-049 (issue #2350) makes such a pattern **unevaluable** at evaluation time — a deny that cannot be evaluated prompts rather than silently passing — but nothing refuses the pattern where the operator wrote it: `permissions.allow`/`deny` in the CLI config, the preset `allowedTools`/`deniedTools` lists that `toolNamesToPatterns` turns into `Tool(*)`, and `PermissionEnforcer` in `agent-session` (`permission-enforcer.ts:150`) all accept any string.

## What it should do

At config/preset load, each argument-scoped pattern is checked against the grammar of the tool's declared `argumentKind` (CORE-049): a URL pattern that the URL grammar rejects, a path pattern for a tool with no path kind, a pattern naming a tool no profile declares an argument key for — refused with the pattern and the reason, before any agent turn runs. The gate-level unevaluable route stays the floor; this is the loud surface above it.

## Provenance

Named by the `proposal-reviewer` of CORE-049 (2026-08-29) as a separate root: load-time validation is a different surface (agent-session / agent-cli) from the gate.

woojubb:
**처분: 수정됨 (FIXED)** — 2026-09-04 일괄 이슈 정리

이 이슈의 수정이 `develop`에 반영되었습니다.

- 커밋: `868447535`
- 내용: validatePermissionPattern/findInvalidPermissionPatterns in agent-core (URL grammar via matchUrl probe, declared tool without argument key, syntax); PermissionEnforcer constructor throws naming pattern+reason before any turn; unprofiled tools stay on the gate floor. 5+25 tests green.

이 이슈는 227건 전체를 한 번에 분류·처리한 일괄 작업의 일부입니다. 분류 근거와 검수 결과는 로컬 원장에 기록되어 있습니다.

PR #2439, opened 2026-08-28
fix(agent-core): permission patterns match by the kind of argument they scope

### LINKED ISSUE #2350: P1 security: a host-scoped permission pattern matches any host, including the cloud metadata endpoint
## A host-scoped permission pattern auto-approves every host on the internet

`globToRegex` in `packages/agent-core/src/permissions/permission-gate.ts:41-47` translates `*` to `.*` and `**` to `.+`, with no exclusion for `/`, `?` or `#`. The pattern is then anchored against the **raw argument string**.

For a URL argument that is not a wildcard over hostnames. It is a wildcard over the whole URL, including the query and fragment. Reproduced against the shipped function:

```
pattern: https://*.example.com/**

MATCH   https://sub.example.com/ok                      ← intended
MATCH   https://evil.tld/?a=.example.com/x              ← query carries the match
MATCH   https://evil.tld/#.example.com/x                ← fragment carries the match
MATCH   https://169.254.169.254/?x=.example.com/y       ← cloud metadata endpoint
```

The last line is the one that matters. An operator who writes `WebFetch(https://*.corp.example.com/**)` — the natural way to stop being prompted for an internal documentation host — has granted silent, auto-approved access to **the link-local cloud metadata service**, and to every other host an attacker can name, because the matched text can live entirely in a part of the URL the server never sees as a path.

## Why this is filed separately from the SSRF work

Found while designing the egress boundary for issue #2026, where the allowlist is the escape hatch that makes reclassifying `WebFetch` tolerable. But the defect is **not** in the egress design and is not confined to URLs:

- The same matcher backs `Read`, `Write`, `Edit` and `Bash` argument patterns. `Bash(git *)` matches any command containing `git ` at the right offset; `Read(/home/me/**)` is a wildcard over a string that happens to be a path.
- It is in `agent-core`'s permission gate — the foundation every consumer reaches — not in a tool.

Per `.agents/rules/finding-depth.md` this is FOUNDATIONAL relative to the egress Task: that Task is reachable only because this exists, and folding the matcher rewrite into a security-boundary change would turn one reviewable unit into a permission-system rewrite. Filed, and the egress Task will declare `depends_on` it.

## What a fix has to settle

1. **`*` must not cross a delimiter.** Which delimiter depends on the argument kind — `/` for paths, and for URLs the answer is that a glob over a raw URL string is the wrong shape entirely.
2. **URL patterns need parsing, not globbing.** Scheme + host (+ optional path prefix), compared against `new URL(arg).hostname` after canonicalisation. `new URL` already normalises `2130706433`, `0x7f.1`, `127.1` and `①②⑦.0.0.1` to `127.0.0.1`, so a parsed comparison gets alternate-encoding handling for free; a string glob gets none of it.
3. **Whether one matcher can serve paths and URLs at all.** They have different delimiters, different canonical forms and different threat models. One function serving both is how this happened.

## Acceptance

- `https://*.example.com/**` does not match a URL whose host is not under `example.com`, for query, fragment, userinfo and port placements of the matched text.
- A path pattern's `*` does not cross `/`.
- The tests state the anti-goal explicitly — the refusal must not come from matching the URL as an opaque string — so a future implementation cannot pass them by escaping more characters while keeping the same shape.

## Provenance

Found during design review of the egress boundary for issue #2026 (SECURITY-003). Reproduced by running the shipped `globToRegex` verbatim; no request to any endpoint was made.

woojubb:
## Background

An operator narrows what an agent may do with permission patterns such as `WebFetch(https://*.corp.example.com/**)` or `Read(/src/**)`. The gate turned every such pattern into one regular expression — `*` → `.*`, `**` → `.+`, nothing excluded — and matched it against the tool's argument as a raw string, whatever kind of thing the argument was. A host pattern was therefore a wildcard over the whole URL: `https://169.254.169.254/?x=.example.com/y` matched `https://*.example.com/**` because the text lived in the query, and so did a fragment, a path segment or userinfo carrying it. A path pattern's `*` crossed `/`. Issue #2350 (P1) reproduced all five URLs against the shipped function; the same shape was measured live through the CLI's print mode (a deny on `127.0.0.1` bypassed by `127.1`; `Write(<dir>/*)` covering a file one directory down).

Lane: L2

The change touches `packages/agent-core/docs/SPEC.md`'s contract sections and `src/` in four packages.

## Purpose

Make a pattern mean what the operator wrote: a host pattern matches hosts, a path pattern matches path segments, and a pattern or argument the gate cannot interpret is never read as "not denied".

## What changes

- `packages/agent-core/src/permissions/permission-gate.ts` — a tool profile declares its argument as one object, `argument: { key, kind }` (`path` | `url` | `command` | `text`; `IToolPermissionArgument`), and the gate matches by kind. `url`: the pattern is split by a grammar (scheme or `*`; host labels with `*`/`**` or an IPv6 literal; port or `*`; optional path) and the argument is parsed with `new URL`; scheme, canonicalised host (label-wise for wildcards, IDN through `domainToASCII`, trailing dot dropped), port (the scheme default when the pattern names none) and percent-decoded path segments are compared; query and fragment never participate. `path`: separators and `.`/`..` normalised, `*` inside a segment, `**` across, drive prefixes absolute. `command`/`text`: the glob as before.
- A match is one of three answers. What cannot be interpreted — a URL that does not parse, carries userinfo, has no host or a non-special scheme, or a segment that does not decode; a pattern the grammar rejects or whose literal host does not parse; a relative path under an absolute pattern — is **unevaluable** and reaches `hasUnevaluableArgumentPattern`, which the gate already routes to a prompt (deny in `plan`, deny in the policy resolver). A bare `Tool(*)` matches any invocation, key or no key (the preset `deniedTools` contract; a keyless tool's `(*)` used to prompt).
- `packages/agent-tools` and `packages/agent-framework` profiles declare their kinds; every test registration moves to the object form; the SPEC's contract sections carry the grammar and every unevaluable condition.
- CORE-049 Task + spec through the gate pipeline, archived; the first scenario-bearing item under the lane process (three agent-executable CLI scenarios).

## Why this way

`proposal-reviewer` (three rounds, ENDORSE): the kind belongs to the profile that already names the key (CORE-030's principle), not to the pattern's spelling (A2); parsing, not escaping, is what makes canonical forms right in both directions (A3); a third state, not a boolean, because on the deny list a non-match falls through to the mode policy and `inspect` is `auto` in every mode (A5); one object, not a union alias, because object shapes are interfaces here (A6); reclassifying `WebFetch` alone (A4) leaves paths and the deny side untouched and removes the escape hatch the egress design (issue #2026) needs. Depth: root — one function serving three argument kinds was the cause.

## How it was verified

Recorded in the spec's GATE-VERIFY/GATE-COMPLETE entries: the six test files (71 cases) in four packages; typecheck of all four (the `@ts-expect-error` key-without-kind registration would be an unused directive if the type accepted it); the regression red-proof checker on all three source files; four mutations (URL kind to the string glob, path kind to the glob, and each unevaluable side to `no-match`) each re-redding exactly the named cases; the three CLI scenarios run against the built binary; `pnpm harness:scan`.

The local review ran four rounds before the push: it found the new host regex joining labels with an optional dot (`*.example.com` matched `evilexample.com` — the defect class inside the fix; fixed with literal-dot joins and six suffix vectors), gitignore-style `**` semantics, the barrel exports, the allow-direction statement for a keyless `Tool(*)`, and the decode rules — each folded in with vectors. The pre-push suite then refused the branch for importing `node:path`/`node:url` into agent-core's browser bundle (CORE-028); the matcher is builtin-free now (a lexical normaliser; IDN through the platform `URL`), pinned by the bundle test and IDN vectors. After the first push the `quality` check went red on the lint-warning ceiling (2204 against 2203): the base sits exactly at the ceiling and the new `matchUrl` carried a `complexity` warning (23 against 15). `matchUrl` is now four clause functions — parse, host, port, path — with no verdict change (59/59), and the workspace lint is back at 2203. The pre-push suite does not run the workspace lint by design; the lesson (run `pnpm lint` before the first push of product code) is mirrored into `.agents/memory/`.

## Not in this PR

- Issue #2427 — `Bash(git *)` still matches `git status; rm -rf /` (the command kind keeps the glob; separators are a shell-parsing decision).
- Issue #2428 — a malformed pattern is refused only at the gate, never where the operator wrote it.
- Issue #2429 — a relative `filePath` reaches the gate unresolved.
- Issue #2430 — a `hooks.PermissionDecision` command hook never fires in print mode.
- Issue #2433 — `gate.mjs` writes a GATE-IMPLEMENT entry the plan-order scan cannot bind (hit on this branch's checkpoint).
- Issue #2394 (HARNESS-128) — the checkpoint entry forms are still declared in the scan.

Closes #2350 — closed by hand with the delivering commit after landing, since `develop` is not the default branch.


woojubb:
## Review — CORE-049 (issue #2350), four rounds, final confirmation on `9c676927f`

**Scope of the change.** `packages/agent-core/src/permissions/permission-gate.ts` now dispatches on the declared argument kind (`url` / `path` / `command` / `text`) with a third matcher state, `unevaluable`, that prompts on the deny side instead of passing; the per-kind matchers live in the new builtin-free `argument-matchers.ts` (agent-core ships a browser bundle — CORE-028). Tool profiles in `agent-tools` and `agent-framework` declare `argument: { key, kind }`; `SPEC.md` names the per-kind rules.

**Rounds 1–3 (all resolved, each with vectors in `permission-gate.test.ts`):**
- MUST — the host regex joined labels with an optional dot, so `WebFetch(https://*.example.com/**)` matched `https://evilexample.com/` — the defect class inside the fix. Fixed: literal-dot joins in `hostPatternToRegex`; six suffix vectors; mutation-proved (restoring the optional dot turns the vectors red).
- SHOULD — `**` in a path pattern was not gitignore-style. Fixed: `segmentGlobToRegex` (`**` spans zero-or-more segments, adjacent `**` collapsed).
- SHOULD — barrel exports (`IToolPermissionArgument`, `TArgumentKind`) and the both-directions statement for a keyless `Tool(*)` in SPEC.md. Fixed.
- SHOULD — a `%2F` in a pattern segment and a percent-decode failure in the argument must be unevaluable, not compared as text. Fixed: `decodedSegments` returns null on either; `%2A` in a pattern is a literal `*` (marked, not a wildcard).
- MUST (pre-push suite) — `node:path` / `node:url` in agent-core's browser bundle. Fixed: lexical `normalisePathText`; IDN via the platform `URL` (`labelToASCII`); `browser-bundle-node-builtins.test.mjs` 2/2 on a bundle rebuilt from the head.

**Round 4 — confirmation of `9c676927f` (diff vs `3c2635419`: three files).**
- SHOULD (`packages/agent-core/docs/SPEC.md:336`) — the `path` row still named `path.posix.normalize`. Resolved: the row now states the lexical rule (`\` → `/`; empty and `.` segments dropped; `..` collapsed, discarded at an absolute root, kept on a relative path; trailing `/` dropped; `/…` and `X:/…` absolute with the drive lower-cased); each clause checked against `normalisePathText` on the vector set.
- SHOULD (`argument-matchers.ts:145-154`) — the IDN branch was untested. Resolved: TC-01 pins `WebFetch(https://*.bücher.example/**)` vs `https://a.xn--bcher-kva.example/` → match, and deny `WebFetch(https://*.exa mple.com/**)` → `approve` (the prompt verdict), reaching the null branch of `labelToASCII`.
- CONSIDERs applied and red-proofed against the pre-fix source in an isolated worktree (test file only): the label-boundary vector fails on the old grammar (`exam\ple.com` truncated to host `exam`, deny matched `https://exam/`; new host class `[^/\\:?#@[\]]+` makes the pattern unevaluable), and the `C:\` vs `Read(c:/**)` vector fails under `^[A-Za-z]:\/` (with `^[A-Za-z]:(?=\/|$)` the drive root `c:` is absolute). The `hostPatternToRegex` comment no longer names `domainToASCII`.
- Suite at head: `permissions/__tests__` 59/59; `tsc --noEmit` on agent-core exit 0; `pnpm harness:scan` receipt written at `9c676927f` (147 passed, 1 skipped).

| Severity | Location | Status |
| --- | --- | --- |
| SHOULD | `packages/agent-core/docs/SPEC.md:336` | Resolved in `9c676927f` |
| SHOULD | `packages/agent-core/src/permissions/argument-matchers.ts:145-154` | Resolved in `9c676927f` (`permission-gate.test.ts:166-183`) |
| CONSIDER | `packages/agent-core/src/permissions/argument-matchers.ts:148` | Open, non-gating: IDNA-mapped separators / numeric labels in a wildcard host label are unevaluable rather than mapped — every such divergence prompts, none passes |

Residuals filed as their own issues, not fixed here: issue #2427 (command separators), issue #2428 (load-time refusal of a malformed pattern), issue #2429 (relative filePath reaches the gate unresolved), issue #2430 (print-mode hook), issue #2433 (gate.mjs GATE-IMPLEMENT entry).

ACTIONABLE FINDINGS: 0
REVIEWED BASE: d15c9ee8dc1836c38eb08922826b81cda1568da3
REVIEWED HEAD: 9c676927f069a4c115fde5fab74bfc1bbc9064d4

woojubb:
## Review — follow-up on `6ad461510` (two commits after `9c676927f`)

**Why the commits exist.** The `quality` check on the first push went red on the lint-warning ceiling: 2204 warnings against `--max-warnings 2203`. Measured: a clean base at `d15c9ee8d` lints to exactly 2203; the branch at `9c676927f` to 2204; the one new warning was `matchUrl` in `packages/agent-core/src/permissions/argument-matchers.ts` — `complexity` 23 against 15.

**`e4314cba6` — `matchUrl` split into four clause functions (`parseComparableUrl`, `matchHost`, `matchPort`, `matchPathname`): verdict-preserving.** Static mapping: every early return in the old body has a counterpart with the same verdict — grammar null → unevaluable; parse throw / non-special scheme / empty host / userinfo → unevaluable; scheme mismatch → no-match after all four as before; host: regex null → unevaluable, miss → no-match, literal parse throw → unevaluable, mismatch → no-match, both propagated unchanged; port: pattern absent → `url.port === ''` (non-default port with no port in the pattern stays no-match), `*` → any, else exact against `DEFAULT_PORT`; path absent → match, else the same three verdicts with the same `%2a` marking. Differential proof: the pre- and post-refactor modules run side by side over 47 patterns × 68 arguments — 3196 pairs, 0 differences, all three verdicts exercised (275 match / 2038 no-match / 883 unevaluable). At head (isolated worktree): permissions suite 59/59; `tsc --noEmit` 0 errors; eslint on the file exit 0; full `pnpm lint` `2203 warnings`, exit 0 — back at the baseline. At `9c676927f` the same single-file eslint reports exactly the `matchUrl` complexity warning, confirming it was the one.

**`6ad461510` — memory entry `.agents/memory/pre-push-does-not-run-the-lint-ceiling.md` + index line.** Each claim checked against its owner: `scripts/harness/pre-push.mjs` contains no `pnpm lint` / `eslint` / `max-warnings` invocation; the root `lint` script carries `--max-warnings 2203` and `scripts/harness/lint-warning-baseline.json` is `{"warnings": 2203}`; `harness:verify:release` ends in `pnpm lint`; `.github/workflows/ci.yml:444` is the `Lint-warning ceiling` step; `scan-lint-warning-ratchet.mjs` (INFRA-039) states the ceiling lives in the `lint` script and is kept off pre-push because a full lint "is minutes". The "File ignored because of a matching ignore pattern" claim reproduced verbatim on a test file. The host-memory entry with the same slug exists; `scan-memory-mirror`, `scan-memory-neutrality` and `reference-kind-qualified` pass.

| Severity | Location | Problem |
| --- | --- | --- |
| NIT | `.agents/memory/pre-push-does-not-run-the-lint-ceiling.md:7` | "~3 min" is stated as a measurement; a cold run at head in an isolated worktree took 26 s. The ratchet scan's own wording is "minutes". Non-gating. |

ACTIONABLE FINDINGS: 0
REVIEWED BASE: d15c9ee8dc1836c38eb08922826b81cda1568da3
REVIEWED HEAD: 6ad4615106d4a1541c33130a41601bbe100d393f

woojubb:
## Review — merge head `164313138` (develop `b3e9cc6c8` merged into `6ad461510`)

**What the head is.** The reviewed branch tip `6ad461510` (verdict above: 0 actionable) with develop `b3e9cc6c8` (PR #2440, DOCS-031 batch 02, nine docs-only commits) merged in via update-branch. Both parents are ancestors of it.

**Checks on the merge itself.**
- The only files where develop's delta and the branch overlap are `.agents/loop-runs/backlog-execution-orchestrator.jsonl` and `.agents/loop-runs/user-execution-scenario.jsonl`. Against `origin/develop` each is `1 0` (one line added, none removed); both parse (18 and 16 lines), zero duplicate `runId`, the CORE-049 record exactly once in each (`r20260828160733`, `r20260828155714`).
- The non-ledger diff against `origin/develop` is identical, `index` lines aside, to the reviewed diff `d15c9ee8d..6ad461510` (exit 0).
- The tree is exactly the union: `git diff 164313138 6ad461510 -- . ':!.agents/loop-runs'` equals develop's DOCS-031 delta reversed (exit 0), and develop's delta touches 0 files under `packages/`. The verifications recorded against `6ad461510` — permissions suite 59/59, browser bundle free of `node:` imports, `tsc` 0 errors, `pnpm lint` at the 2203 baseline, the 3196-pair `matchUrl` differential with 0 verdict changes — apply to `164313138` unchanged.

| Severity | Location | Problem |
| --- | --- | --- |
| NIT (carried, non-gating) | `.agents/memory/pre-push-does-not-run-the-lint-ceiling.md:7` | "~3 min" stated as measured; a cold run took 26 s. Unchanged by the merge. |

ACTIONABLE FINDINGS: 0
REVIEWED BASE: b3e9cc6c892e41f29a7e7d43946083a660c2734e
REVIEWED HEAD: 164313138e90d45ff981d8816c28f8ac42d89318

PR #2496, opened 2026-08-29
docs: require English GitHub issues

### LINKED ISSUE #2490: [enhancement] Add gates for package boundaries and shared-file reduction
## Problem
In this monorepo, changing one package can trigger broad CI fan-out because the criteria for keeping code shared versus moving it into a package are not explicit or mechanically checked. This increases coupling, review scope, and verification time.

## Desired outcome
Define and enforce measurable package-boundary rules: retain a shared file only when at least two independent packages use a domain-neutral, owned API; move package-specific helpers, fixtures, constants, and test data into the owning package; and make CI report the affected scope and any full-suite promotion reason. Add a gate and apply the classification across the repository.

## Alternatives considered
Duplicating every shared file would reduce some CI fan-out but would create drift and duplicated maintenance. Keeping everything in root-level shared modules would increase coupling. A measured ownership and neutrality gate is the balanced approach.

## Context
This follows the observed broad verification fan-out and pre-push resource pressure tracked in #2489. Implementation requires triage, a backlog/spec, regression tests, and staged rollout.

Duplicate search: no existing issue covering this exact package-boundary gate was found.

woojubb:
## Background
A newly filed issue used Korean prose, which makes cross-session triage, search, and automation inconsistent.

## Summary
- state that GitHub issue titles and bodies default to English
- add the same requirement to the issue-triage skill
- translate issue #2490 into English while preserving its intent

Closes #2490

## Verification
The issue was updated and the repository language rule and triage skill now agree.

### DIFF
--- a/.agents/rules/naming-style.md
+++ b/.agents/rules/naming-style.md
@@ -13,6 +13,9 @@ Parent: [AGENTS.md](../../AGENTS.md) | Index: [rules/index.md](index.md)
 - **Everything else defaults to English**, unless the user explicitly requests otherwise: code and comments,
   ALL repository documents (including `.design/`), commit messages (conventional-commits format), and any other
   written artifact.
+- **GitHub issue titles and bodies default to English** for consistent triage, search, automation, and
+  cross-session handoff. Translate an issue written in another language when discovered; preserve quoted
+  user text only when needed as evidence and provide an English summary.
 
 ### Korean Writing Style (only when Korean output is explicitly requested)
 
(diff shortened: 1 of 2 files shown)


woojubb:
Remote scans identified a planning-order violation: the rule/skill implementation commit predates a valid Task/spec planning checkpoint. This PR is being closed rather than bypassing the repository gate. Reopen the work with a valid planning checkpoint commit before the implementation commit.

PR #2493, opened 2026-08-29
fix: close parent issues after decomposition

woojubb:
## Background

When a broad issue is decomposed into child issues, leaving the parent open causes repeated decomposition and keeps duplicate work in the queue.

## Summary

- require decomposed parent issues to link all child issues and close immediately
- preserve open issue behavior for unsplit conversion

Closes #2490

Lane: L2

## Verification

- pnpm harness:scan:consistency
- pnpm harness:scan:test-plans
- pre-push contract and affected scan passed

PR #2502, opened 2026-08-29
fix: preserve deterministic agreement finalization

woojubb:
Restore the deterministic AGREEMENT-parent finalization rule that was lost during conflict resolution of #2493. This follow-up keeps the conversion and issue-triage workflow deterministic. Local pre-push verification was attempted; the contract tier stalled alongside concurrent repository verification, so remote required CI is the authority.


woojubb:
Closing this follow-up because its standalone post-merge documentation edit is rejected by the repository planning-order gate. The required merge-completion and conflict-prevention rules are already merged in #2493; any further documentation change must start with a new valid planning checkpoint.

PR #2571, opened 2026-09-01
fix(harness): allow continuation retry after recorded failure

woojubb:
## Background

A repaired GATE-IMPLEMENT continuation could not be retried after any prior continuation FAIL because ordinary last-entry ordering was applied to the continuation route. That made a valid prior PASS permanently unusable without deleting audit evidence, blocking the approved AGREEMENT-006 B2 migration.

## Summary

- Selects the newest historical GATE-IMPLEMENT PASS only for explicit continuation routing.
- Preserves last-entry ordering for every ordinary gate and revalidates all current continuation criteria.
- Adds adversarial coverage for no-PASS history, ordinary later-FAIL ordering, and invalid current status, artifacts, ancestry, Task/PLAN, and worktree state.
- Records the completed PROC-030 Task/spec and append-only Work-Run/review evidence.

## Work evidence

- Independent exact-head review: ACTIONABLE FINDINGS: 0 at 9c1c32df50f13eca43ae21539d0d2cbee7b46492
- Focused suites: 7 files, 260/260 tests
- Pre-push contracts: 223 files, 4,638/4,638 tests; follow-up 222/222
- Hermetic tests: 71 files, 1,123/1,123 tests
- Affected harness scan: 55 passed, 2 declared skips
- Full build: passed
- Work-Run validators: included

Work-Run: f26123c5-11df-4527-9a87-b765fd06c2da


woojubb:
Superseded before review: the Work-Run opening-head attestation was not created before this PR's creation timestamp, so the immutable PR-evidence contract cannot validate this PR. No implementation finding is being bypassed; the byte-identical substantive commits will be replayed on a fresh develop-based branch under a new Work-Run, attested before replacement PR creation.

PR #2643, opened 2026-09-06
fix(infra-2635): discovery-loader tolerates a missing scripts/harness/ dir

### LINKED ISSUE #2635: [bug] discovery-loader.mjs crashes on ENOENT when scripts/harness/ is absent, breaking gate.test.mjs and blocking every push to develop
## Problem

`node scripts/harness/__tests__/gate.test.mjs` fails 76 of 93 tests on `develop`'s current tip. Every failure traces to the same crash:

```
Error: ENOENT: no such file or directory, scandir '/var/folders/.../robota-gate-XXXXX/scripts/harness'
    at readdirSync (node:fs:1584:26)
    at candidateFiles (scripts/harness/discovery-loader.mjs:20:10)
    at discoverAdditionalScans (scripts/harness/discovery-loader.mjs:66:22)
    at loadScanCommands (scripts/harness/discovery-loader.mjs:96:28)
    at scripts/harness/run-all-scans.mjs:1346:36
```

`run-all-scans.mjs` calls `loadScanCommands` at MODULE LOAD TIME (top-level `await`). `gate.test.mjs`'s
fixtures build a minimal isolated root (only the files one specific test needs) and never create a
`scripts/harness/` subdirectory — legitimate, since gate.mjs's own logic doesn't need one. But
`discovery-loader.mjs`'s `candidateFiles()` calls `readdirSync(harnessDir)` unconditionally, which
throws `ENOENT` instead of treating "no such directory" as "nothing to discover."

Since `pre-push-check.sh` / `.husky/pre-push` runs the full test suite (including `gate.test.mjs`)
before any push, **this currently blocks every push to `develop` from every branch**, regardless of
what that branch touches — not just branches that happen to import `run-all-scans.mjs` directly.

## Root cause

Introduced by `3381a7d4ab` ("fix(harness): discover self-declared scan entrypoints", issue #2399,
PR #2626), which added the `discovery-loader.mjs` module and wired it into `run-all-scans.mjs`'s
top-level `SCAN_COMMANDS` initialization without handling a root that has no `scripts/harness/`
directory at all.

## Fix

`candidateFiles()` now catches `ENOENT` specifically and returns an empty candidate list; any other
`readdirSync` failure (permissions, a path that names a file, …) still propagates. Added a regression
test in `scripts/harness/__tests__/scan-discovery.test.mjs` asserting `discoverAdditionalScans` resolves
to `[]` for a root with no `scripts/harness/` directory at all.

## Evidence

Before:
```
$ npx vitest run scripts/harness/__tests__/gate.test.mjs
 Test Files  1 failed (1)
      Tests  76 failed | 17 passed (93)
```

After:
```
$ npx vitest run scripts/harness/__tests__/scan-discovery.test.mjs scripts/harness/__tests__/gate.test.mjs
 Test Files  2 passed (2)
      Tests  96 passed (96)
```

woojubb:
## Background

`node scripts/harness/__tests__/gate.test.mjs` fails 76 of 93 tests on `develop`'s current tip, and since `.husky/pre-push` runs the full test suite before every push, this currently blocks every push to `develop` from every branch — not just branches that import `run-all-scans.mjs` directly. Filed as issue #2635.

## Purpose

`run-all-scans.mjs` loads scan commands at module-load time via `discovery-loader.mjs`. `gate.test.mjs`'s fixtures build a minimal isolated root (only the files one specific test needs) and never create a `scripts/harness/` subdirectory — legitimate, since `gate.mjs`'s own logic doesn't need one. `discovery-loader.mjs`'s `candidateFiles()` called `readdirSync(harnessDir)` unconditionally, which throws `ENOENT` instead of treating "no such directory" as "nothing to discover."

## What changes

- `scripts/harness/discovery-loader.mjs`: `candidateFiles()` now catches `ENOENT` specifically and returns `[]`; any other `readdirSync` failure still propagates.
- `scripts/harness/__tests__/scan-discovery.test.mjs`: new regression case asserting `discoverAdditionalScans` resolves to `[]` for a root with no `scripts/harness/` directory at all.
- Paired spec/Task: `.agents/spec-docs/done/INFRA-2635-*.md`, `.agents/tasks/completed/INFRA-2635-*.md` (Lane L1, approved under the standing `LANE-L0-L1` delegated class).

## Why this way

A defensive `try/catch` at the exact call site is the smallest change that removes the crash and matches this repository's general pattern of failing closed only on genuine errors, not on "nothing here." Widening this to a shared rule/helper is out of scope for this lane and would be its own item if a recurrence is observed elsewhere.

`REVIEW VERDICT`: local self-review, 0 findings (`pnpm harness:review:record -- --findings 0`).

## How it was verified

- `pnpm exec vitest run scripts/harness/__tests__/scan-discovery.test.mjs` — 3/3 pass, including the new no-directory regression case.
- `pnpm exec vitest run scripts/harness/__tests__/scan-discovery.test.mjs scripts/harness/__tests__/gate.test.mjs scripts/harness/__tests__/task-complete.test.mjs` — 115/115 pass (previously 76/93 and 1/19 failing on ENOENT respectively).
- `pnpm harness:test` (full suite) — 255/255 test files pass with the fix applied.
- `node scripts/harness/run-all-scans.mjs --affected --context pr` — 60/61 pass; the sole pre-existing red (`work-run-measurement`) is unrelated to and unaffected by this change.
- User Execution Test Scenario: not applicable — internal harness test-fixture fix with no runnable Robota CLI/TUI/browser/SDK surface.

## Not in this PR

- The pre-existing, unrelated `work-run-measurement` scan failure and other unrelated harness debt are out of scope.

Lane: L1

Closes #2635

### DIFF
--- a/.agents/evals/work-runs/8be3e594-26ce-43d3-913e-7717543ee864/g0-r0.json
+++ b/.agents/evals/work-runs/8be3e594-26ce-43d3-913e-7717543ee864/g0-r0.json
@@ -1,0 +1,86 @@
+{
+  "schemaVersion": 1,
+  "disposition": "included",
+  "runId": "8be3e594-26ce-43d3-913e-7717543ee864",
+  "generation": 0,
+  "revision": 0,
+  "identity": {
+    "repository": "woojubb/robota",
+    "branch": "fix/discovery-loader-enoent",
+    "baseCommit": "50c1f54547a219a248522727a0edaad0467dfb89",
+    "headCommit": "91c2cbd985947d7c4758ec2a08569916f9e6e90c",
+    "headTree": "8c25e6bb316c1337a671df3b8fcc0f383eeda8fe",
+    "commitOids": [
+      "b284508fafb1434d28cb71b4c3f9fbae4d71e163",
+      "91c2cbd985947d7c4758ec2a08569916f9e6e90c"
+    ],
+    "trailerDigest": "77b641e2e1bfda7069aa470f0ef707618daa4279e1d2750e527494bcd1773e2d",
+    "ownerFingerprint": "67c3371f71776dc0101e2b16c8dfc0d5b526ab031a2f98eaea8e47dc8eccfbd9"
+  },
+  "cohort": {
+    "key": "L1/infra",
+    "lane": "L1",
+    "workKind": "infra"
+  },
+  "events": [
+    {
+      "schemaVersion": 1,
+      "runId": "8be3e594-26ce-43d3-913e-7717543ee864",
+      "sequence": 1,
+      "previousHash": null,
+      "type": "work.claimed",
+      "at": "2026-09-06T06:59:04.109Z",
+      "data": {
+        "branch": "fix/discovery-loader-enoent"
+      },
+      "hash": "ad081672aa17c556152d1764818b2ce8e86b5722398bf466065a4724bae582f8"
+    },
+    {
+      "schemaVersion": 1,
+      "runId": "8be3e594-26ce-43d3-913e-7717543ee864",
+      "sequence": 2,
+      "previousHash": "ad081672aa17c556152d1764818b2ce8e86b5722398bf466065a4724bae582f8",
+      "type": "work.bound",
+      "at": "2026-09-06T07:25:30.308Z",
+      "data": {
+        "workId": "INFRA-2635",
+        "lane": "L1",
+        "workKind": "infra"
+      },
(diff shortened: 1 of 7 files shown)


woojubb:
Superseded — develop's tip already carries an equivalent fix for this exact issue (existsSync check in candidateFiles(), see origin/develop). Closing without merging to avoid a duplicate/conflicting change. Thanks!

PR #2645, opened 2026-09-06
ci(infra-2631): retire the non-functional Claude Code Review GitHub Action

### LINKED ISSUE #2631: Retire the non-functional Claude Code Review GitHub Action and its harness dependents
## What was observed

Retire the non-functional Claude Code Review GitHub Action and its harness dependents

## Expected outcome

Track this requested change as the canonical external work item before creating its Task/spec identifier.

## Location / context

Created by `allocate-work-item-id.mjs`; duplicate search was performed by exact title before creation.

## Duplicate search

Exact-title search found no existing Issue.

woojubb:
## Background

The Claude Code Review GitHub Action (`.github/workflows/claude-code-review.yml`) is treated as
permanently non-functional per owner direction. This change disables it via a deliberate,
git-tracked `if: false` (the same idiom `review-gate.yml` already uses for its own retired steps)
while preserving every other line — the action reference, `with:` block, full `prompt:`, and
incident-history comments — byte-for-byte, so the integration can be reinstated by reverting the
`if:` line alone.

The two harness scans that mechanically assert this workflow's static shape
(`scan-claude-review-coverage.mjs` and its test suite) are updated to recognize the exact
`CLAUDE-CODE-REVIEW: RETIRED (INFRA-2631)` marker together with `if: false` and skip their shape
checks for this job. Two skills that referenced the action's live review output
(`pr-finding-resolution-loop`, `automated-review-convergence`) are updated to state the retirement
explicitly so an empty review result is treated as the expected steady state, not a missing step.

`gh workflow disable` was considered and rejected — it leaves no git-tracked record and does not
change what the mechanical scans read.

Lane: L2
Closes #2631

## Test Plan

- `pnpm harness:test` — `scan-claude-review-coverage.test.mjs` (22 tests, including 3 new cases for
  the retirement marker) pass.
- `pnpm harness:scan` — full affected-scan run against `origin/develop`; findings unrelated to this
  diff (pre-existing on develop) documented in the spec's Evidence Log.
- Local review recorded: `pnpm harness:review:record -- --findings 0`.

### DIFF
--- a/.agents/evals/work-runs/bc3cc530-1351-46bb-b1ac-ae3af34379f1/g0-r0.json
+++ b/.agents/evals/work-runs/bc3cc530-1351-46bb-b1ac-ae3af34379f1/g0-r0.json
@@ -1,0 +1,87 @@
+{
+  "schemaVersion": 1,
+  "disposition": "included",
+  "runId": "bc3cc530-1351-46bb-b1ac-ae3af34379f1",
+  "generation": 0,
+  "revision": 0,
+  "identity": {
+    "repository": "woojubb/robota",
+    "branch": "ci/retire-claude-code-review-action",
+    "baseCommit": "35c4cbdfd35ce3b924e51812b912dd8ee5a1c639",
+    "headCommit": "5a33aa2f8956d2595d0576cb4c38a614dd07b399",
+    "headTree": "58699ca1df94a3207a34350d0d77a6fdd9fa5f46",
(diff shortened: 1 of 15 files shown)


woojubb:
Superseded by #2646 — this branch hit a false-positive `work-run-measurement: identity-mismatch` after `origin/develop` advanced during review. #2646 carries identical content rebuilt on the current develop tip.

Issue #2055, opened 2026-08-22
P1 contract: make memory retrieval provenance a live end-to-end contract


woojubb:
## Summary

Memory retrieval provenance is declared end to end but has no live producer. Recalled content can be injected into a turn while `memory_retrieved` is never emitted and `usedMemoryReferences` is never written, so `/memory used` remains empty.

There is also a duplicate recall-budget contract whose reader is on a dead controller path.

## Evidence

- [`memory_retrieved` is part of the transport event contract](https://github.com/woojubb/robota/blob/15800bf0f0023406cf06981a7afcc2cfdcd17f63/packages/agent-interface-transport/src/event-contracts.ts#L48-L58).
- [Framework formats it as a user-visible memory event](https://github.com/woojubb/robota/blob/15800bf0f0023406cf06981a7afcc2cfdcd17f63/packages/agent-framework/src/memory/memory-event-format.ts#L8-L22).
- [`InteractiveSessionHistoryTracker` stores, resets and reads `usedMemoryReferences`](https://github.com/woojubb/robota/blob/15800bf0f0023406cf06981a7afcc2cfdcd17f63/packages/agent-framework/src/interactive/interactive-session-history-tracker.ts#L54-L115), but production has no write site.
- [`/memory used` reads that permanently empty state](https://github.com/woojubb/robota/blob/15800bf0f0023406cf06981a7afcc2cfdcd17f63/packages/agent-command/src/memory/memory-command.ts#L177-L183).

## Proposed direction

Make the live recall path record references and emit/persist `memory_retrieved` through the existing memory-event/history path. Collapse the duplicate recall-budget contract onto the live retrieval owner.

## Acceptance

- A turn that recalls memory records exact references and a persisted `memory_retrieved` event.
- `/memory used` lists the references for that turn.
- Resume/replay preserves provenance without duplicating it.
- Recall budget has one owner and one live reader path.
- A framework-level scripted session test drives the real recall -> history -> command flow.

## Tracking

GitHub tracking for internal MEM-001. No open exact GitHub issue was present on 2026-08-22.

Not fixed inline because this audit was authorized for analysis/reporting and issue registration only; the fix crosses framework, transport and command contracts.

woojubb:
**Citation repair — this issue cites a source path that no longer exists on `develop`.** The work is unaffected; the file moved.

```
packages/agent-interface-transport/src/event-contracts.ts   ← gone
packages/agent-interface-session/src/event-contracts.ts     ← now here
```

[...]

Issue #2653, opened 2026-09-06
[bug] affected-scope build graph excludes devDependency-only workspace packages, breaking build/quality on partial PRs


woojubb:
## Summary

`scripts/harness/workspace-graph.mjs`'s **build**-operation dependency closure (`BUILD_DEPENDENCY_FIELDS = ['dependencies', 'optionalDependencies', 'peerDependencies']`, used by `pnpm build:affected`) deliberately excludes `devDependencies`. The **test**-operation closure (`testDependencies`, used by `pnpm test:affected`) correctly includes test-file-scanned imports, but `test:affected` never builds anything itself — it assumes the build step already produced a `dist/` for everything it needs.

When a PR's affected-scope diff doesn't happen to touch a `devDependencies`-only workspace package that some OTHER selected package's test files import at runtime, the build step never builds it, and `test:affected`/`typecheck:affected` then fail with `Cannot find package '<name>'` / `TS2307: Cannot find module`.

## Evidence

Reproduced on PR #2651/#2652 (`fix/scenario-git-dir-leak`, diff scoped only to `packages/agent-framework/examples/**` + one new test + a harness baseline line):

```
FAIL src/interactive/__tests__/interactive-session-background-tasks.test.ts
Error: Cannot find package '@robota-sdk/agent-session-analytics' imported from '...'
FAIL src/testing/__tests__/session-log-external-payload-replay-functional.test.ts
Error: Cannot find package '@robota-sdk/agent-provider-replay' imported from '...'
FAIL src/testing/__tests__/usage-assertion-functional.test.ts
Error: Cannot find package '@robota-sdk/agent-session-analytics' imported from '...'
```

CI's `build` job's "Build full or affected workspace" step built only 13 packages — neither `@robota-sdk/agent-session-analytics` nor `@robota-sdk/agent-provider-replay` was among them, even though `agent-framework/package.json` lists both as `workspace:*` devDependencies and three of its own test files import them directly.

Locally reproduced: temporarily removing `packages/agent-session-analytics/dist` and `packages/agent-provider-replay/dist`, then running `pnpm --filter @robota-sdk/agent-framework exec vitest run` on the three failing test files reproduces the identical error byte-for-byte.

Confirmed **not caused by any one PR's diff** — all three failing test files were last touched months ago on `origin/develop`, and this reproduces for any partial-affected PR touching `packages/agent-framework` (or any package with test-only devDependency workspace packages) whose own diff doesn't also touch the devDependency packages themselves.

[...]

Issue #2650, opened 2026-09-06
[bug] ARCH-042 workspace-authority scenario leaks ambient GIT_DIR, breaking identity isolation in a git-hook context


woojubb:
## Summary

`packages/agent-framework/examples/verify-workspace-project-authority.ts`'s `GitWorkspaceIdentityResolver.resolve()` shells out to `git -C <cwd> rev-parse --show-toplevel|--absolute-git-dir` without stripping ambient `GIT_*` environment variables. When this scenario runs inside a `git push` pre-push hook invoked from a linked worktree, Git exports `GIT_DIR` (pointing at the *hook's own* repository) into the hook's process tree. Node's `spawnSync`/`execFileSync` inherit `process.env` by default, so the scenario's nested `git -C <tempRepo> rev-parse --absolute-git-dir` calls silently ignore `-C` and resolve to the ambient `GIT_DIR` instead — collapsing the scenario's two independent temp repositories (`project`, `other-project`) onto the same identity key.

Effect: `pnpm --filter @robota-sdk/agent-framework scenario:verify:workspace-authority` (and therefore any `git push` from a linked worktree that runs `verify-change.mjs`'s scenario step for `packages/agent-framework`) fails with:

```
Error: a grant for one repository was accepted for a different root
    at assertCondition (.../verify-workspace-project-authority.ts:37:25)
```

even though the scenario passes cleanly when run standalone (no ambient `GIT_DIR`).

## Evidence

- Reproduced deterministically: `GIT_DIR="$(git rev-parse --absolute-git-dir)" pnpm exec tsx --conditions=source examples/verify-workspace-project-authority.ts` fails with the exact error above; the same command without `GIT_DIR` set passes every time.
- This exact hazard is already known and defended against elsewhere in this codebase: `packages/agent-cli/src/subagents/git-worktree-isolation-adapter.ts`'s `createGitEnvironment()` strips every `GIT_*`-prefixed env var before shelling out, with the comment "Git hooks export GIT_* variables that force child git commands back to the hook repository" — and its own test (`git-worktree-isolation-adapter.test.ts`, "ignores inherited Git hook environment variables") asserts exactly this. The scenario file never received the same treatment.

## Proposed direction

Give `GitWorkspaceIdentityResolver.resolve()` in `verify-workspace-project-authority.ts` the same `GIT_*`-stripping environment (mirroring `createGitEnvironment()`'s pattern) before its two `execFileSync('git', ...)` calls.

## Impact

[...]

PR #2654, opened 2026-09-06
fix(agent-framework): wire memory recall provenance and drop dead recall-budget field (MEM-2055)

woojubb:
## Summary

- Fixes issue #2055: `InteractiveSession.recallTurnMemory()` injected `<recalled-memory>` content into a turn but never recorded provenance — `usedMemoryReferences` stayed empty and no `memory_retrieved` event ever fired, so `/memory used` was constitutionally empty even when a recall happened.
- Records every returned reference and emits one `memory_retrieved` `memory_event` per reference, through the same `recordMemoryEvent` path capture already uses.
- Fixes a reset-ordering hazard: `resetUsedMemoryReferences()` now runs in the execution controller immediately before recall, not inside `executePromptTurn` (which runs after recall and was wiping a reference the same turn had just written).
- Separately collapses a duplicate recall-budget contract: `AutomaticMemoryController.retrieve()` was dead code (no production callers) and its `retrieval` config field duplicated the live `recallMemory.budget` seam; both are removed rather than wired live, to avoid making capture policy start gating recall — a coupling nothing in the issue asked for.
- Ports the ARCH-042 GIT_DIR/GIT_WORK_TREE leak fix from #2650 / PR #2652 (same fix, same regression test) since this branch pushes from a linked worktree touching `agent-framework` and hits the same leak otherwise. This duplication resolves to a no-op merge once that fix lands on `develop` and this branch rebases.
- Adds a genuinely automatable user-execution scenario (HARNESS-105): a new example script drives a real `InteractiveSession` with a scripted provider (no live credential needed) and asserts `getUsedMemoryReferences()` / the emitted `memory_retrieved` event — the same path `/memory used` reads.

## Expected CI note

This branch touches `packages/agent-framework`, so it is expected to hit the same **pre-existing, already-tracked** `build`/`quality` CI gap as PR #2652: issue #2653 (`workspace-graph.mjs`'s affected-build-graph excludes devDependency-only workspace packages, e.g. `@robota-sdk/agent-session-analytics` / `@robota-sdk/agent-provider-replay`, breaking `test:affected`/`typecheck:affected` on partial PRs). This is unrelated to this diff and left unfixed here as deliberately out of scope.

This PR also touches a `package.json` (an npm-script addition only — `pnpm-lock.yaml` has zero diff against `develop`), which flips the `dependency audit` job's diff-scoped gate to run the full lockfile scan. That surfaced 4 **pre-existing, untriaged** vulnerabilities already in `develop`'s own lockfile (`@xmldom/xmldom`, `browserslist`, `fast-uri`, `qs` — none reachable from this diff), never caught before because no recent PR happened to touch a manifest. Filed as issue #2655, left untriaged here as deliberately out of scope (dependency remediation for 4 unrelated transitive packages, and/or a CI-scoping redesign, is its own change).

## Test plan

- [x] `pnpm --filter @robota-sdk/agent-framework test` — 1650 tests green (includes new TC-07/08/09 regression coverage + the GIT_DIR-leak regression test)
- [x] `pnpm --filter @robota-sdk/agent-cli test` — 475 tests green
- [x] `pnpm exec tsx examples/verify-memory-recall-provenance.ts` (packages/agent-framework) — prints the expected automatable scenario result
- [x] `node scripts/harness/scan-spec-user-execution-section.mjs` — passes
- [x] Local pre-push gate (`git push`) — 118 scans passed, 2 skipped, 3 pre-existing advisory findings unrelated to this diff


woojubb:
Closing to apply a review-driven fix (a history-ordering bug an independent review caught: the recall notice rendered before the user's own message) plus a corrected PR body — the harness's post-PR reopen policy requires maintainer-authorized findings/red-check grounds to push into an existing PR, which isn't available in this autonomous session, so the fix is being rebuilt on a fresh branch/PR instead of pushed here. Superseding PR will reference this one.

Issue #2656, opened 2026-09-06
[bug] workspace-operation-selection.mjs's examples-typecheck branch always selects zero packages (checks owner's top-level directory, not its changed files)


woojubb:
## Problem

`scripts/harness/workspace-operation-selection.mjs`'s package-selection logic for the `examples-typecheck` operation is:

\`\`\`js
} else if (operation === 'examples-typecheck') {
  selectedNames = new Set(
    [...ownerNames].filter((name) => byName.get(name).directory.startsWith('examples/')),
  );
}
\`\`\`

This filters on whether the **owner package's own top-level directory** (e.g. \`packages/agent-framework\`) starts with \`"examples/"\`. No workspace package's directory is ever literally \`examples/*\` — packages live at \`packages/<name>\`; their example scripts live in an \`examples/\` *subdirectory* of that package. This condition can never be true, so \`selectedNames\` is always empty whenever any real package changed — which (via \`workspace-affected-plan.mjs\`) forces \`mode: 'none', packages: []\` for **any** diff, meaning the affected-mode \`examples-typecheck\` operation currently selects **zero packages, always**, regardless of what changed.

## Reproduction

Reproduced directly against \`origin/develop\` and against a **minimal synthetic single-file diff** (\`--changed-file packages/agent-framework/examples/some-file.ts\` alone):

\`\`\`
node scripts/harness/workspace-affected.mjs --operation examples-typecheck --base-ref origin/develop --format json
→ {"mode":"none","packages":[],"reason":"no changed workspace package is an example owner", ...}
\`\`\`

This is independent of any specific PR's diff shape — it reproduces for any change at all, because the top-level-directory check the filter runs against can structurally never be true.

## Why this hasn't been caught before

The CI `examples-typecheck` job (see `.github/workflows/ci.yml`) only runs its "guarantee affected example consumer dist" fallback step when `needs.build.result != 'success'` — i.e., only when the `build` job is *already* red for an unrelated reason (this surfaced while investigating PR #2654, which independently hits the pre-existing `build`/`quality` gap tracked as issue #2653). In the normal path, the same broken selection is exercised via `pnpm examples:typecheck:affected` → `workspace-affected-run.mjs`, but that path treats `mode: 'none'` as "0 tasks" and **silently passes** rather than failing — so under normal CI (`build` green), this job has been quietly typechecking nothing, for every PR, indefinitely.

## Suggested direction

[...]

Issue #2655, opened 2026-09-06
Umbrella: dependency, build-graph, and package-boundary integrity


woojubb:
## Problem

The \`dependency audit\` CI job (\`osv-scanner scan source --config osv-scanner.toml --lockfile pnpm-lock.yaml\`) only runs when a PR's diff touches \`pnpm-lock.yaml\` or any \`package.json\`. Any PR that does not touch those files skips the scan entirely — so a vulnerability that lands in the lockfile via an untouched transitive dependency is never surfaced until, by chance, some unrelated PR edits a \`package.json\` (e.g. adding an npm script) and triggers a full-lockfile scan for the first time in a while.

## Reproduction

Observed on PR #2654 (MEM-2055), which only added an npm script entry to \`packages/agent-framework/package.json\` (no dependency version changes — \`pnpm-lock.yaml\` has zero diff against \`origin/develop\`). This alone was enough to flip the job's \`changed\` gate to true and run the full scan, which failed with 4 pre-existing, unrelated findings already present in \`develop\`'s own lockfile:

- \`GHSA-6gmq-8vp8-gcm6\` — \`@xmldom/xmldom\` (dev), 0.8.13 / 0.9.10
- \`GHSA-73wf-gq98-2v4g\` / \`GHSA-c83g-rgw3-j3cx\` — \`browserslist\` 4.28.2
- \`GHSA-5jgf-p345-68v8\` / \`GHSA-f65p-4m7j-42xc\` / \`GHSA-fph4-wmhf-6fwf\` / \`GHSA-jqff-g426-hqxp\` — \`fast-uri\` 3.1.5
- \`GHSA-4mjr-xmp4-gh2g\` / \`GHSA-x5fp-wj9c-mxmx\` — \`qs\` 6.15.2

None of these are reachable from MEM-2055's diff (the lockfile is byte-identical to \`develop\`). They are pre-existing, untriaged findings that the job's own scoping rule has been silently hiding from every PR that doesn't happen to touch a manifest file.

## Why this matters

The job already has an established triage precedent — two prior findings are allowlisted inline with a reason and a tracking ID (\`GHSA-2p57-rm9w-gvfp\` → REMOTE-001, \`GHSA-f88m-g3jw-g9cj\` → INFRA-044). The 4 findings above have never gone through that triage because the job never ran against them. The scoping-by-diff design means "no findings" on develop's own CI history is not evidence of "no vulnerabilities" — it only means "no one touched a manifest recently."

## Suggested direction

[...]

Issue #2154, opened 2026-08-22
ARCH-050: root build does not model complete package artifact tasks and copied-asset edges


woojubb:
## Observed

The canonical root `pnpm build` executes workspace `build:js` stages, but the complete `agent-cli` product assembly lives in the package-level `build` script. As a result, `agent-cli-web` is not rebuilt or copied into `agent-cli/dist/web` by a clean root build.

The copied-asset dependency is also absent from the workspace manifest graph, so normal topological build ordering cannot discover it.

## Where

Root build orchestration, `packages/agent-cli/package.json`, `packages/agent-cli-web`, and the CLI web-asset copy step.

## Why not fixed inline

The architecture-refresh depth guardian classified this as FOUNDATIONAL: adding one command to the root script would repair today's CLI symptom while leaving package artifact capability discovery and copied-asset ordering undefined for the next custom artifact package.

## What it is not

This is not only a stale CLI web asset or a missing `devDependency`. The root cause is that the workspace build graph models the partial `build:js` convention rather than each package's complete publishable artifact contract.

## Evidence

- INFRA-063 previously measured the same root-build omission but intentionally left it outside that task's release-suite scope.
- HARNESS-053 observed that `agent-cli-web` had no comparable freshly generated dist after the root build.
- ACLI architecture-refresh run `r20260822142736`, finding `ACLI-R1-F005`.

## Scope extension from ACLI-R1-F003 (2026-08-23)

The complete artifact contract must be transactional and exact:

- assemble node, types, and copied web output in staging and replace publishable `dist` atomically;
- reject partial-failure mixtures of old and new generations;
- derive an expected emitted/packed file manifest and reject obsolete files retained by `clean:false` or removed source entries;
- exercise seeded stale files and build failure in the release-path corpus.

woojubb:
**처분: 보류 (DEFER) — 이 이슈는 열린 채로 둡니다** — 2026-09-04 일괄 이슈 정리

227건 일괄 분류에서 이 항목은 **바로 구현하지 않고 별도 스펙/이니셔티브로 진행**하기로 분류되었습니다. 독립 검증 결과 이 보류 판단은 타당한 것으로 확인되었습니다.

[...]

PR #2657, opened 2026-09-06
fix(agent-framework): wire memory recall provenance and drop dead recall-budget field (MEM-2055)

### LINKED ISSUE #2055: P1 contract: make memory retrieval provenance a live end-to-end contract

woojubb:
## Background

Supersedes #2654 (closed): rebuilt on a fresh branch to fold in an independent review's findings, since this harness's post-PR reopen policy needs maintainer-authorized grounds this autonomous session cannot self-grant.

With memory enabled, every recall injected `<recalled-memory>` content into the turn invisibly: the declared `memory_retrieved` event never fired, `usedMemoryReferences` provenance was never written, and `/memory used` was constitutionally empty even though a recall happened and its content reached the model. Separately, two recall-budget contracts claimed ownership of the same concern — `IAutomaticMemoryConfig.retrieval` (a required field) was read only by `AutomaticMemoryController.retrieve()`, which had zero production callers, so tuning it changed nothing; live per-turn recall budgeting came from the separate `recallMemory.budget` seam. Closes issue #2055.

## What changes

- `InteractiveSession.recallTurnMemory()` now returns the recall's `IMemoryEvent`s alongside the rendered context, instead of recording them itself. It also records `usedMemoryReferences` immediately (order-insensitive, needed for `/memory used`).
- The execution controller stashes those events and records them (via `SessionHistoryTracker.recordMemoryEvent`) in the turn's `finally`, mirroring exactly where the existing post-turn auto-capture events are recorded — **after** the turn's own user/assistant messages are already in history. Recall runs before `executePromptTurn` (its result feeds the turn as ephemeral context), so recording immediately, as the first cut of this fix did, rendered the "Memory recalled" notice ahead of the user's own message for that turn; an independent review caught this before merge.
- Fixed a related reset-ordering hazard: `resetUsedMemoryReferences()` now runs in the execution controller immediately before recall, not inside `executePromptTurn` (which runs after recall and was wiping a reference the same turn had just written).
- Deleted the dead `AutomaticMemoryController.retrieve()` method and the duplicate `retrieval` field from `IAutomaticMemoryConfig` — both removed rather than wired live, to avoid making capture policy start gating recall, a coupling nothing in the issue asked for. Updated `agent-cli`'s `buildMemorySessionOptions` and every affected test for the simplified shape.
- Adds a genuinely automatable user-execution scenario (HARNESS-105): `examples/verify-memory-recall-provenance.ts` drives a real `InteractiveSession` through its public constructor with a scripted provider (no live credential needed) and a fake `IMemoryStore`, asserting `getUsedMemoryReferences()` and the emitted `memory_retrieved` event — the same path `/memory used` reads. Wired into the package's `scenario:verify` chain.
- Ports the ARCH-042 GIT_DIR/GIT_WORK_TREE leak fix from issue #2650 / PR #2652 (byte-identical to that PR — same fix, same regression test) since this branch's own `git push` from a linked worktree touching `agent-framework` hits the same leak otherwise. Resolves to a no-op merge once #2652 lands on `develop` and this branch rebases.

## Why this way

The recall-vs-capture ordering mirrors an existing, already-reviewed precedent in this exact controller (`capturePostTurnMemory`'s post-turn recording) rather than inventing a new mechanism — the fix is "record recall events at the same point capture events are already recorded," not a new history-ordering concept. Removing the dead `retrieve()`/`retrieval` pair instead of wiring it live keeps this change a provenance fix, not a scope expansion into a second recall-budget mechanism nobody asked to activate.

## How it was verified

- `pnpm --filter @robota-sdk/agent-framework test` — 1650 tests green, including new TC-07/08/09 regression coverage (a recall records provenance and emits `memory_retrieved`; the used-references set resets each turn; an empty recall records/emits nothing) and the GIT_DIR-leak regression test.
- `pnpm --filter @robota-sdk/agent-cli test` — 475 tests green.
- Manually confirmed the history-ordering fix: `session.getFullHistory()` after a recall-triggering turn now reads `user → assistant → memory-event`, not `memory-event → user → assistant`.
- `pnpm exec tsx examples/verify-memory-recall-provenance.ts` (inside `packages/agent-framework`) — prints the expected automatable scenario result.
- `node scripts/harness/scan-spec-user-execution-section.mjs` — passes.
- Local pre-push gate (`git push`) — 118 scans passed, 2 skipped.
- Independent review pass (this repo's `pr-review-reviewer`) — caught the history-ordering MUST finding above (fixed), confirmed the GIT_DIR-fix port is byte-identical to PR #2652, confirmed the removed `retrieve()`/`retrieval` pair has no orphaned callers, confirmed the new example script is genuinely reachable (fails when the ordering fix is reverted).

## Not in this PR

- Issue #2653: `build`/`quality` fail on this branch — a pre-existing, already-tracked gap (`workspace-graph.mjs`'s affected-build-graph excludes devDependency-only workspace packages), reproduced identically on other PRs touching `agent-framework`.
- Issue #2655: `dependency audit` fails on this branch — this PR's `package.json` diff (an npm-script addition only, zero lockfile diff) flipped that job's diff-scoped gate to run a full lockfile scan for the first time in a while, surfacing 4 pre-existing, untriaged vulnerabilities in unrelated transitive packages (`@xmldom/xmldom`, `browserslist`, `fast-uri`, `qs`) never reachable from this diff.
- Issue #2656: `examples-typecheck`'s CI fallback step fails — a pre-existing, unrelated bug in `workspace-operation-selection.mjs`'s package-selection logic for that operation (it always selects zero packages, for any diff), reproduced against a synthetic single-file diff independent of this PR.
- `tui-e2e` fails on this branch — a repeat of the already-filed, already depth-classified **FOUNDATIONAL** issue #2154 (a copied-asset build dependency between `agent-cli` → `agent-cli-web` → `agent-transport-gui` is invisible to the workspace dependency graph by both of its detection mechanisms). Not a new defect; not fixed here.
Closes #2055

Lane: L1

Work-Run: 68122b14-a7fc-4b14-ba22-cb61bffe4cb1


woojubb:
Closing to apply findings from a second independent review pass (a missing regression assertion for the ordering fix, plus a disclosure gap for an already-red, unrelated CI check) — same reopen-policy constraint as the #2654→#2657 supersession. Superseding PR will reference this one.

Issue #2670, opened 2026-09-07
Umbrella: CLI/TUI usability, accessibility, and diagnostics


woojubb:
Spec `CLI-2004` § Decision verdict (i) adopted TWO pacing waits for screen-reader mode. Only one shipped.

## What is missing

The pre-write park — move the cursor to column 0, pause briefly, then write the changed line — so a
screen reader that tracks the caret restarts on the new text instead of resuming mid-line. It was to
be tunable as `ROBOTA_SCREEN_READER_PREPARK_MS` (default one Ink frame plus margin, bound 5000).

## Why it was not shipped with the rest

Every transcript line is written by Ink's own frame loop, synchronously. Wrapping that write with a
delay means either blocking the event loop or reordering frames, and neither is acceptable. Shipping
the environment variable anyway would have left a documented tunable that silently does nothing,
which is worse than its absence — so `screen-reader-pacing.ts` does not read it, and
`packages/agent-transport-tui/docs/SPEC.md` records the gap under Known limitations.

## What a fix needs

An owned write path: an Ink `stdout` wrapper (the `render()` option) that can defer a frame without
blocking, or a small queue in front of it. The behaviour to reach is the one verdict (i) states, and
`screen-reader-pacing.ts` is where the resolver belongs when it returns.

## Where to look

- `packages/agent-transport-tui/src/screen-reader-pacing.ts` — the module, and its header comment
  stating why the park is absent
- `packages/agent-transport-tui/docs/SPEC.md` — Pacing table and Known limitations
- `.agents/spec-docs/active/CLI-2004-tui-screen-reader-mode.md` § Solution 12 and § Decision verdict (i)

Found by the independent review of the CLI-2004 branch, 2026-09-08.

## Consolidated 90 → 10 work register

This Issue is one of ten umbrella work items. Every unfinished scope item from the frozen 90-Issue set is assigned to exactly one umbrella. Checked entries were already completed and are included only for historical traceability; unchecked entries remain part of this Issue's work until completed or explicitly split by a later owner decision.

- [ ] #1991 — P1: no /doctor — a broken provider, settings file, plugin or skill is diagnosed by reading logs — unfinished scope transferred here
- [ ] #1992 — P1: no session recap — a session that ran while the terminal was unfocused gives no summary of what happened — unfinished scope transferred here
- [ ] #1993 — P1: no transcript search — a long session can only be re-read by scrolling — unfinished scope transferred here
[...]

PR #2716, opened 2026-09-12
fix(harness): preserve bounded completion records

woojubb:
## Background

Required completion records for the delivered ARTIFACT-2655 work were rejected by planning validation. Approved documentation archival, historical failed merge attempts, and exact delivered Task/spec archival with parent projections needed bounded handling.

## Purpose

Permit those required records without granting executable planning authority or weakening exact merge evidence.

## What changes

- Extend shared staged/history record classifiers and add negative regression coverage.
- Preserve failed historical attempts and archive the verified ARTIFACT, MERGE and META records.
- Reconcile post-owner-merge verification when the required-check projection is confirmed empty; unknown discovery remains a failure.
- Preserve the parent initiative scope and its unfinished BOUNDARY-2655 work.

## Why this way

Reuse the existing metadata predicates rather than exempting all documentation or manufacturing retrospective plans. This is one coherent record-validation repair with its real delivery records; the larger diff includes immutable evidence and negative tests. Independent review found two MUSTs, both classified LOCAL and repaired: incomplete archive source comparison, and historical captures incorrectly consuming delivery allowance. Final local review: ACTIONABLE FINDINGS: 0. META GATE-DONE: PASS. This is not merge authorization or a remote CI result.

## How it was verified

- Focused memory-only regression suite: 66 tests passed, including observed RED-to-GREEN invalid scope substitution and history classification cases.
- Affected harness scan: 71 passed, one explicit skip, no failures in the final DONE invocation.
- Actual staged archival validation passed; committed-history validation accepted ten topic commits before the final review receipt.
- CI-only Git integration regressions were written and syntax-checked, not run locally. Remote CI must execute them before merge.
- No local worktrees, clones or Git fixtures were created. Product builds were not repeated for this harness-only change.
- User execution scenarios: not applicable. Internal repository record validation adds no CLI, TUI, browser or public SDK behavior.

## Not in this PR

BOUNDARY-2655 remains open: repository-wide ownership classification, package-specific migrations, drift enforcement and affected/full execution reporting. This PR does not complete the umbrella issue. No npm publication, main promotion, deployment or remote protection change.

Refs: issue #2655. Do not close the umbrella on this partial delivery.

PR #2752, opened 2026-09-19
feat(ui-terminal): one theme model drives every colour the terminal emits (SCREEN-2002 unit 1)

woojubb:
## Background

The terminal UI's colours were spread across two token modules and a scattering of inline literals:
a component that wanted a colour imported `PALETTE`, a status glyph carried its colour as part of its
own constant, a markdown renderer wrote raw ANSI escape bytes by hand, and two dependencies
(`marked-terminal`, `cli-highlight`) coloured whatever they liked from their own defaults. Nothing in
that arrangement can be re-pointed at a different theme, which is why the CLI ships exactly one
colour scheme today.

That is an accessibility problem, not a preference one. Roughly 1 in 12 men has a colour-vision
deficiency, and the palette leans on red/green for the distinction that matters most — a tool call
that succeeded versus one that failed, a diff line added versus removed. `cli-highlight`'s own
default theme is red/green too, and it falls back to it PER KEY, so a theme that misses a key gets
the red/green back silently. Users on light terminals get a palette tuned for dark ones. Users who
need still text get an animated waiting indicator, gated only on whether the terminal supports
colour at all.

## Purpose

Work unit 1 of 3 for SCREEN-2002. It builds the one theme model every colour in the terminal is
resolved through, and migrates every render site onto it — without changing a single byte of what a
user sees today. Choosing a theme (settings, the `/theme` command, the picker) is work unit 2; user
and plugin theme files are work unit 3.

## What changes

- **`packages/agent-ui-terminal/src/theme/`** — the new owner. `theme-contracts.ts` (the token
  model: `colors`, `markdown`, `syntax`, `motion`), `built-in-themes.ts` (`dark`, `light`,
  `dark-daltonized`, `light-daltonized`, plus `resolveTheme()`), `theme-styles.ts` (the one place
  Ink's colour grammar is decoded, and where both dependencies' option objects are built),
  `theme-context.tsx` (`ThemeProvider` / `usePalette` / `useMotionTokens` / `useMotion`), and
  `color-vision.ts` (a CVD projection used only by a test guard).
- **Every render site reads the live theme.** ~31 components moved from `PALETTE.x` to
  `usePalette().x`. `status-glyph.ts` keeps the glyph (a symbol is the same in every theme) and hands
  the colour to `statusGlyphColor(colors, kind)`, so the glyph and its colour cannot drift apart.
  Two view models now name a status KIND instead of a colour.
- **The two dependencies are told what to colour with.** `markdownRendererOptions()` and
  `syntaxHighlightTheme()` reproduce `marked-terminal`'s and `cli-highlight`'s own chain order, key
  for key, so the default `dark` theme renders byte-identically to what shipped before.
- **Reduced motion is a first-class input.** `useMotion()` is one rule over the colour gate, screen
  reader mode and the setting; `WaveText` arms no interval at all when it is false.
- **The anti-drift floor gained three ratchets** in `palette-consistency.test.ts`: no file outside
  `src/theme/` may name the built-in theme data or call a chalk colour function, and the motion gate
  has exactly one consumer.

## Why this way

The endorsed recommendation (`REVIEW VERDICT: ENDORSE`, `proposal-reviewer`, round 3, recorded in
`.agents/spec-docs/active/SCREEN-2002-…md` § Architecture Review) is a **single resolved theme
object published through React context**, with chalk as the name→SGR layer.

Alternatives weighed and rejected: (a) a second "accessible palette" module selected by a flag —
doubles the drift surface the SCREEN-006 floor exists to prevent and cannot reach the two
dependencies at all; (b) semantic token indirection over the existing `PALETTE` — leaves the
dependencies and the hand-written diff escapes untouched, which is where the red/green actually
lives; (c) passing the theme as a prop — one missed prop leaves one component on another theme,
and there is no mechanical way to catch it.

Deriving chalk instances from theme values is **not** the "name→SGR layer" SCREEN-006 declined to
invent: Ink's own `colorize` already performs exactly that mapping, and chalk is a direct dependency
of this package.

Depth verdict: `LOCAL` (`finding-depth-triager`) — the problem is the absence of a theme model, not
a symptom of something underneath it.

## How it was verified

- `pnpm --filter @robota-sdk/agent-ui-terminal exec vitest run` — 935 tests, 109 files, green.
- `pnpm lint --max-warnings 2356` — 2347 warnings, 0 errors.
- `pnpm harness:scan` — 161 passed, 1 skipped.
- **Byte-identity, on the real binary.** The SCREEN-006 PTY scenarios run the BUILT `robota` binary
  in a real PTY and assert the exact SGR bytes: `NO_COLOR=1` yields zero colour parameters in the
  whole transcript, and a normal run still emits `ESC[32m` for the success glyph, `ESC[36m` for the
  assistant label, and `38;5;120` on `48;5;22` for an added diff row.
- **The four recorded byte exceptions** (chalk's paired closers, the builders' chain order, `type`
  keeping `dim`, and `strong`/`em`/`listitem`/`hr` not being tokens) are asserted by name in
  `theme-styles.test.ts` rather than discovered later.
- **The CVD guard is a real floor**, not a decoration: it fails rather than skips on a value it
  cannot simulate, and carries a fixture proving the red/green pair falls below its distance floor.
- Local review: 2 rounds, 8 findings, all resolved; round 3 clean.
- User execution test scenario: **planned, gated in work unit 2** — the two scenarios recorded in the
  Task (a theme run and a reduced-motion run) drive the `/theme` command and the `--reduced-motion`
  flag, neither of which exists until unit 2. DONE-GATE-STAGE-1 PASS is recorded on the Task.

## Not in this PR

- **Choosing a theme** — settings keys, the `appearance-settings-patch` host action, the `/theme`
  command and the picker are work unit 2; **user and plugin theme files** and `parseThemeDocument`
  are work unit 3. Both are specified in the same spec document and land on this branch's successor.
- **The silent fallback in `foreground()`/`chalkNamed()`** for an unknown colour name is unreachable
  today (every built-in value is validated by TC-01) and is recorded against work unit 3's
  `parseThemeDocument` step, which is the boundary where it first becomes reachable.
- **A theme change does not repaint scrollback.** Committed transcript entries go through Ink's
  `<Static>` and are emitted once; a theme applies to the live region and to everything rendered
  after it. Recorded in the SPEC, not a defect.

Part of #2670


woojubb:
## Code review — 4 rounds, all findings resolved

**Refs:** base `origin/develop` = `9cced2d93c94e68b518e563eacdef111bc6c57bf` (= merge-base, unmoved across all four rounds) · head = `f16d5c7630b5db8feeac9b59781ee33afacacaf6`.

**Verification at head:** `pnpm --filter @robota-sdk/agent-ui-terminal exec vitest run` → 936 passed / 109 files · `tsc --noEmit` → exit 0 · `pnpm lint --max-warnings 2356` → 2347 warnings, 0 errors · `pnpm harness:scan` → 161 passed, 1 skipped, `scan receipt written` · `pnpm --filter @robota-sdk/agent-ui-terminal test:pty` → 33 passed against the freshly built binary.

### Findings raised and resolved

| Round | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | MUST | `diffMeta` gave a diff's `diff --git`/`index ` metadata rows a colour they never had — they were `ESC[2m` dim only. Theming a row that carried no colour is a rendering change dressed as a theme. | Key removed from the model entirely; `IDiffRowStyles.meta` is `base.dim(row)`. Both halves pinned: the hunk header takes the theme's colour, the metadata row does not. |
| 1 | SHOULD | The reduced-motion input had zero coverage. | `theme/__tests__/theme-context.test.tsx` added — no timer armed under `<ThemeProvider reducedMotion>`, timers armed when it is off, none when the colour gate is closed, and the static frame asserted as the muted token with every wave stop absent. |
| 1 | SHOULD | `isThemeColor` accepted `bg…` background names as token values. | Narrowed to foreground names; `bgRed`/`bgCyan` added to the rejected list, so the narrowing is actually guarded. |
| 2 | SHOULD | The anti-drift floor matched the theme data module's **path**, so the `theme/index.ts` barrel re-exported the same data straight past it — and `render-markdown.ts` was already taking `DARK_THEME` that way. | Floor matches the **symbol** now; the colour data is no longer re-exported at all. `resolveTheme()` answers "no theme was resolved" inside the module and `listBuiltInThemes()` serves enumeration, so unit 2's registry does not need the floor weakened. The fixture proves the barrel route fires. |
| 2 | SHOULD | The diff renderer forced chalk to level 3, overriding colour **depth**, not just colour: on a 256-colour terminal a hex-valued theme emitted truecolor for diff rows while the rest of the frame downsampled. | Only the OFF case is overridden. Pinned in both directions — 256-colour form at level 2, truecolor at level 0. |
| 2 | SHOULD | TC-03 — a declared unit-1 completion criterion — had no test anywhere. `cli-highlight` falls back **per key** to its own red/green default, so a theme that stopped reaching it would fail silently. | Both halves asserted and each proven to go red with the corresponding wiring removed. |
| 2 | SHOULD | `CHALK_STYLES` was built with `Object.fromEntries`, so it inherited `Object.prototype`: `constructor`, `toString`, `valueOf`, `__proto__` all answered truthy, the `undefined` guard in `foreground()` never fired, and the caller got `createChalk` back typed as a style function. `isThemeColor` was right about the same strings all along, because it uses a `Set`. | The table is a `Map`. Latent today; work unit 3's parsed user themes are the path that would reach it. |
| 2 | SHOULD | `CHALK_COLOR_CALL` exempted three chalk names while the test's own header and `docs/SPEC.md` both declared two. `chalk.reset` is not colour-neutral — it clears whatever style the caller established. | Exemption dropped; the fixture asserts positively that `chalk.reset(row)` is refused. |
| 2 | CONSIDER | The CVD guard treated a non-finite distance as passing: `NaN < MIN_DISTANCE` is false, so one bad coefficient would turn the guard green for every pair at once. | `!Number.isFinite(...)` covers `undefined` and `NaN` in one clause. |
| 3 | SHOULD | The TC-04 amendment's rationale asserted coverage that does not exist — it named "the two backgrounds", which the guard does not check and which measure 17.8 (light-daltonized, deuteranopia) against a floor of 50, so acting on the sentence would have turned the criterion red for colours that are correct; the third guarded pair went unmentioned. | Amendment names the three real pairs and states why the backgrounds are excluded, with the foreground figures (56.1–68.6) that make the exclusion principled. Verified line-by-line against the test. |
| 3 | CONSIDER | `foreground('bgRed')` still returned a background style, because the shared table holds the `bg…` entries `background()` needs. | Named branch gated on the foreground names; new test is red-proof (removing the gate fails it). |
| 3 | MUST | The committed head was not prettier-clean — the fix existed staged but uncommitted, and `pnpm harness:scan`'s `scan receipt NOT written: working tree is not clean` had been swallowed by a tail-grep for `scans passed`. | Committed. Verified against the **blob** (`git show <sha>:<path> \| prettier --check`), and confirmed a prettier fixed point, so the underlying non-idempotence — prettier re-indenting a multi-line HTML comment nested in a list item on every pass — is broken rather than quiet for one run. |
| 3 | NIT | Two residual imprecisions in the criterion sentence. | Category words now cover the syntax pair the guard checks; the timing claim now says "unsatisfiable against the palette the design itself calls for" rather than "the moment it was approved". |
| 1 | CONSIDER | `coefficients(published: string)` keeps published numeric tables as parsed strings, losing compile-time checking. | **Accepted as a reasoned trade, not folded.** Numeric literals produced 38 `no-magic-numbers` warnings against a ceiling this branch must hold; the concrete hazard (a typo becoming a silent `NaN`) is closed at the guard, and a typo producing a wrong-but-finite number would be equally invisible either way. |
| 1 | CONSIDER | `built-in-themes.ts` is 337 lines, over the 300-line ceiling. | Justification on the record in the file header: one concern (four colour tables of the same shape), and splitting per theme would separate the dark/light values that must correspond — the drift the token model exists to stop. |

### On the approved criterion that was amended

TC-04 originally read "pairwise Lab ΔE over **every** status and diff colour pair". Measured with the PR's own projection, `status.running` and `status.waiting` are the **same hex** in both daltonized themes (ΔE 0.0 under both simulations), so the criterion is unsatisfiable against the palette the design itself calls for — satisfying it would require a separately distinguishable colour per status kind, which is the opposite of SCREEN-005's rule that status is glyph + word + colour and never colour alone. The amendment is recorded in place with the prior wording quoted, the measurements, and the reasoning. The two must-FAIL fixtures are unchanged, so the criterion still cannot be met by choosing a permissive number.

### Checked and found clean

- **The load-bearing invariant.** The `dark` theme reproduces `marked-terminal@7.3.0`'s `defaultOptions` and the bundled `cli-highlight` `DEFAULT_THEME` key-for-key **including chain order** (`heading: chalk.green.bold`, `firstHeading: chalk.magenta.underline.bold`, `blockquote: chalk.gray.italic`, `del: chalk.dim.gray.strikethrough`, `href: chalk.blue.underline`, `type: chalk.cyan.dim`), `grey` ≡ `gray`, exactly 17 colour-carrying syntax keys, and `emphasis`/`strong`/`link` correctly omitted as modifier-only. `chalk.ansi256(n)` does not downsample at level 1 or 2, so the `dark` diff pairs are byte-stable at every non-zero level — there is no fifth exception.
- **React hooks discipline** across all 30 migrated components, including the two that needed restructuring.
- **`NO_COLOR` × the forced chalk instance** — the one place the new code could have leaked colour past the degradation gate. Correctly gated: the diff path returns before reaching the styles when colour is off.
- Non-React modules genuinely stopped holding colours (a tone key, a status kind), so the hole the previous floor could not see is closed rather than relocated.
- Nothing outside the package referenced the deleted `PALETTE` / `ANSI` / `STATUS_GLYPH`.

**ACTIONABLE FINDINGS: 0**

woojubb:
## Review verdict — SCREEN-2002 work unit 1

Four independent read-only review rounds against this PR, plus three local rounds before the first
push. The full findings table is in the preceding comment; this one carries the machine-readable
verdict at the exact refs it was formed on.

**What was verified at this head**

- `pnpm --filter @robota-sdk/agent-ui-terminal exec vitest run` → 936 passed / 109 files
- `pnpm exec tsc --noEmit` → exit 0
- `pnpm lint --max-warnings 2356` → 2347 warnings, 0 errors
- `pnpm harness:scan` → 161 passed, 1 skipped, `scan receipt written`
- `pnpm --filter @robota-sdk/agent-ui-terminal test:pty` → 33 passed against the freshly built
  `robota` binary, including the SCREEN-006 colour scenarios that assert the exact SGR bytes
- The load-bearing invariant checked at the DEPENDENCY level rather than against this PR's claim:
  the `dark` theme reproduces `marked-terminal@7.3.0`'s `defaultOptions` and the bundled
  `cli-highlight` `DEFAULT_THEME` key-for-key including chain order, with the four recorded byte
  exceptions and no fifth.

**Findings across all rounds:** 8 MUST/SHOULD and 5 CONSIDER/NIT raised, every one resolved or
recorded with its reasoning. Two were latent defects rather than style: a style table built with
`Object.fromEntries` answered truthy for `constructor`/`toString`/`__proto__`, so the builder
disagreed with the module's own validator; and the diff renderer forced chalk's colour DEPTH, not
just colour, so a hex-valued theme emitted truecolor on a 256-colour terminal while the rest of the
frame downsampled. Both are fixed with red-proved tests.

**Open review threads:** 0.

REVIEWED BASE: 9cced2d93c94e68b518e563eacdef111bc6c57bf
REVIEWED HEAD: f16d5c7630b5db8feeac9b59781ee33afacacaf6
ACTIONABLE FINDINGS: 0

woojubb:
PR_MERGE_DECISION
PR: 2752
HEAD: f16d5c7630b5db8feeac9b59781ee33afacacaf6
BASE: develop
BASE-OID: 9cced2d93c94e68b518e563eacdef111bc6c57bf
VERDICT: 0
CI-OBSERVER: ci-gate-watch
CI-RESULT: GREEN
REMOTE-FEEDBACK: empty
SCOPE: SCREEN-2002 work unit 1 — the ITuiTheme token model, four built-in themes with a colour-vision guard, the chalk style builder that drives Ink, marked-terminal and cli-highlight from one source, the theme provider and its reduced-motion gate, the ~31-component migration off the fixed palette, three anti-drift ratchets, and the package SPEC rewrite. No settings keys, no /theme command, no picker, no user or plugin theme files (work units 2 and 3).
AUTHORITY: owner-delegated
APPROVED: yes
APPROVED-BY: agent:claude (owner-delegated)

woojubb:
DELIVERY_COMPLETION_RECORD
OUTCOME: open-partial
ISSUE: 2670
PR: 2752
HEAD: f16d5c7630b5db8feeac9b59781ee33afacacaf6
MERGE: b1689aa8541ba03010053bd5452ba30bebfe4bcd
BASE: develop
LANDING: verified
BRANCH: deleted
BRANCH-DETAIL: feat/screen-2002-accessible-themes deleted locally and on origin after the merge was confirmed reachable from origin/develop; not deleted as part of the merge command
BASE-RESET: develop@b1689aa8541ba03010053bd5452ba30bebfe4bcd
CRITERIA: remaining: SCREEN-2002 work unit 2 (appearance settings keys, the appearance-settings-patch host action, /theme behind IThemeCataloguePort, --reduced-motion resolution, the theme picker with its keybinding context) and work unit 3 (user and plugin theme files with whole-file refusal); then FLOW-2006, BEHAVIOR-2437, SCREEN-2442 and SCREEN-2670
ACTION: leave-open

Issue #2664, opened 2026-09-07
Umbrella: gate correctness, approval ordering, and fail-closed enforcement


woojubb:
## Problem

`gate.mjs approve` (the `runApprove` handler, `scripts/harness/gate-operations.mjs:2152`) writes a `[GATE-APPROVAL] — ✅ PASS` entry after judging only GATE-APPROVAL's **own** listed criteria (via `judgeCriteria(...)` directly). It never calls the ordering check (`orderingResult`, `:1667`) that the Prior-gate map's row for GATE-APPROVAL requires ("GATE-APPROVAL | GATE-WRITE | `review-ready`"). That check is wired only into `runJudge` (`:1776-1777`) — the `judge` subcommand.

Consequence: `gate.mjs approve --route DIRECT ...` can write a standing GATE-APPROVAL PASS on a document that never received a GATE-WRITE PASS and never occupied `review-ready`, with nothing in the tool objecting. If the caller never separately runs `judge --gate GATE-APPROVAL` (which *would* catch it), the skip goes mechanically unnoticed — it did, in this repository, on `.agents/spec-docs/active/HARNESS-2660-a-filter-must-name-a-package-that-declares-the-script.md`, whose GATE-WRITE never ran: the document was first committed already at `status: approved`, having skipped `review-ready` entirely.

## Why it matters

`gate-catalogue.md` states the ordering check runs "before any gate's own criteria" — `approve` is a documented, sanctioned way to record an approval (route DIRECT/CLASS), not a bypass, but its implementation silently omits the one check that would have caught this specific misuse (running `approve` before GATE-WRITE's semantic criteria were guardian-judged). The two subcommands (`approve`, `judge`) should not disagree about whether the prior-gate map applies.

## Suggested fix

`runApprove` should call `orderingResult` (or equivalent) against the candidate document before writing the PASS entry, the same way `runJudge` does, and refuse (matching `judge`'s own refusal shape) if the prior gate's ordering condition is not met.

## How it was found

Landing HARNESS-2660 (#2660): a `backlog-gate-guard` dispatch, judging GATE-WRITE on that document after the fact, traced the skip to this exact code path and confirmed it against `gate-operations.mjs` directly — quoted in the spec document's Evidence Log (`[GATE-WRITE] — 🔴 NON-COMPLIANCE`, 2026-09-07).

## Consolidated 90 → 10 work register

[...]

PR #2753, opened 2026-09-19
feat(tui): choose the theme, the highlighting and the motion (SCREEN-2002 unit 2)

woojubb:
## Background

Work unit 1 of SCREEN-2002 built one theme model for the terminal — four built-in themes, a chalk
style layer driving Ink, `marked-terminal` and `cli-highlight` from one source, and a reduced-motion
gate. It shipped with no way to reach any of it. Every run still rendered the `dark` built-in,
because nothing could say otherwise.

That leaves the accessibility problem the item exists for exactly where it was. A user with a
colour-vision deficiency — roughly 1 in 12 men — still gets a palette that leans on red/green for
the distinction that matters most: a tool call that succeeded versus one that failed, a diff line
added versus removed. A user on a light terminal still gets a palette tuned for dark ones. A user
who needs still text still gets an animated waiting indicator.

## Purpose

Work unit 2 of 3. It adds the choosing: three settings keys, a `/theme` command, a picker that
previews, and a `--reduced-motion` flag. User and plugin theme files are work unit 3.

## What changes

- **Contracts** (`packages/agent-interface-command`): `IAppearanceSettings`,
  `TAppearanceSettingsPatch`, the `appearance-settings-patch` host action, the `show-theme-picker`
  UI intent, and the `IThemeCataloguePort` family.
- **`packages/agent-framework/src/command-api/appearance/`**: three FLAT settings keys — `theme`,
  `syntaxHighlighting`, `reducedMotion` — read with per-key fallback, a patch guard that refuses
  unknown keys, and an applier. Wired into the host's action path.
- **`/theme`** (`packages/agent-command/src/theme/`): `list`, `<id>`, `syntax on|off`,
  `motion on|off`, and bare `/theme` to open the picker. Behind an injected port.
- **`packages/agent-cli`**: `--reduced-motion` / `--no-reduced-motion`, the settings ← env ← flag
  resolver, and one registry instance reaching both the command port and `renderApp`.
- **`packages/agent-ui-terminal`**: the theme registry and catalogue port, refresh-on-result
  appearance reads, the live theme with preview, `ThemePicker.tsx`, and the `theme-picker`
  keybinding context across the catalogue, the published schema and the guide.

## Why this way

**Three flat keys, not one nested object.** `screenReader`, `promptHistory` and `outputStyle`
already sit at the settings document's root; this is the shape a user hand-editing the file expects,
and it keeps each key independently settable — a run can pin reduced motion without also pinning a
theme. The reader falls back PER KEY, so a mistyped `reducedMotion` still leaves a valid `theme`
applied. Discarding the whole document on one bad value would surface to the user as "the theme
stopped working", which is the failure hardest to attribute to its cause.

**The port lives in the contract package, not the command package.** `agent-ui-terminal` does not
depend on `agent-command`. The alternative — the surface re-declaring a structurally-compatible copy,
the way `IKeybindingsFilePort` is handled — type-checks at the composition root and drifts
everywhere else. Both sides already depend on `agent-interface-command`, which is types-only, so the
contract has one declaration.

**No port, no command.** Print mode and `--serve` render no themes, so `/theme` is not registered
there at all. A command that cannot do anything is worse present than absent.

**The picker does not write.** Selecting submits `/theme <id> syntax <on|off> motion <on|off>`
through the normal command path, so the HOST remains the only writer of the settings document and
the picker cannot persist something the command would refuse. One patch, not three: three
submissions are three chances to half-apply what the user saw as one decision.

**Preview on move.** The highlighted row IS the live region's theme while the picker is open,
because a colour scheme cannot be judged from a swatch beside the thing it does not apply to. The
transcript above keeps the colours it was written in — Ink's `<Static>` emits an entry once — so the
preview moves the input frame, the status bar and the overlay, which is the region a reader is
looking at while choosing.

**A flag counts as an override even when it agrees with the setting.** The user pinned it for this
run, so `/theme motion on` can report that it saved the value and that this run keeps what pinned
it, instead of appearing to do nothing.

Depth verdict: `LOCAL`. `REVIEW VERDICT: ENDORSE` for the design is recorded in
`.agents/spec-docs/active/SCREEN-2002-…md` § Architecture Review; this unit's GATE-IMPLEMENT
(continuation) PASS is in the same document's Evidence Log.

## How it was verified

- Package suites: 33 / 1760 / 361 / 964 / 524 passing across `agent-interface-session`,
  `agent-framework`, `agent-command`, `agent-ui-terminal` and `agent-cli`.
- `pnpm lint --max-warnings 2356` — 2350 warnings, 0 errors.
- `pnpm harness:scan` — 161 of 162 (see below).
- Local review: **5 rounds**, 11 actionable findings, all folded; round 5 clean.

**The findings worth reading, because three were defects and one was a pattern.**

`/theme syntax off` persisted its setting, reported success, and changed nothing —
`renderMarkdown`'s `syntaxHighlighting` option existed and no call site passed it. The picker had
joined one of four overlay-blocking lists, so while it was open Up/Down both previewed and walked
prompt history, Enter both selected and submitted, and Escape both cancelled and could abort an
in-flight turn. And the host's appearance writer spread the patch onto the settings document's ROOT,
where the provider profiles and their credentials live, while the guard checked types without
rejecting unknown keys — so a patch carrying `providers` would have reached them. The guard refuses
unknown keys now, and the applier writes the three keys by name so the guard is not the only thing
containing it.

**The pattern is the part worth carrying forward.** Four successive guards were written for the
first defect, and each measured a PROXY for the property that mattered rather than the property:
the setting was checked as persisted rather than rendered; then the prop appearing in a file rather
than reaching the provider; then the file matching a regex rather than being examined (a shared
`g`-flagged regex advanced `lastIndex`, so the floor silently skipped files); then an identifier
occurring N times rather than each call passing it — which three ordinary edits defeated with an
unguarded call still present.

The fix was to remove the obligation rather than police it a fifth time. `useRenderMarkdown()` binds
the theme, the syntax-highlighting setting and screen-reader mode, so no call site has an argument
to forget, and the floor collapses to a single-consumer ratchet: nothing outside that hook may
import the raw renderer. That is entailed by the property rather than standing beside it — in ESM
you cannot call a function you did not import — which is why it ends the sequence and the four
better regexes could not have.

User execution test scenario: **planned, gated in work unit 3** — Scenario 1 exercises custom and
plugin themes, which unit 3 delivers. DONE-GATE-STAGE-1 PASS is recorded on the Task.

## Not in this PR

- **User and plugin theme files** (`~/.robota/themes`, plugin scopes) and `parseThemeDocument`'s
  whole-file refusal are work unit 3, in the same spec document.
- **`task-merged-citation` is red** and is the one scan of 162 that fails. It is `advisory: true` in
  the `pr` context by registry declaration — "the history it grades is moved by OTHER pull requests,
  not by the change under review" — and it is red because unit 1's merged commits cite an item that
  stays `in-progress` by design across three PRs. It clears when SCREEN-2002 reaches a terminal
  status after unit 3.
- **A harness defect found while running this unit's gate** is recorded on issue #2664: the
  checkpoint evidence producer copies `git status --porcelain` unfiltered while its consumer admits
  no auto-generated-churn exemption, so the gate can pass while emitting a record that can never
  bind — and the invalid record then blocks every later run from opening. Not fixed here.
- **A silent per-key fallback** on a wrong-typed appearance value produces no diagnostic. Judged not
  a rule violation, recorded rather than fixed, because a new user-visible surface has no approved
  criterion in this unit.
- **A screen-reader seam in the picker** — `s` and `m` are unreachable in the mode while their state
  is still rendered. Not drift (TC-10's clauses are mode-separated and `/theme syntax off` stays
  reachable), but the fix is a design choice with no approved criterion, so it is carried into work
  unit 3 on the Task, which reopens that component anyway.

Part of #2670


woojubb:
## Review verdict — SCREEN-2002 work unit 2

Five independent read-only review rounds, 11 actionable findings, all folded; round 5 clean.

**What was verified at this head**

- Package suites: 33 / 1760 / 361 / 974 / 529 passing across `agent-interface-session`,
  `agent-framework`, `agent-command`, `agent-ui-terminal` and `agent-cli`.
- `pnpm --filter @robota-sdk/agent-ui-terminal test:pty` — 33 passing against the freshly built
  `robota` binary, 16 files.
- `pnpm lint --max-warnings 2356` → 2348 warnings, 0 errors.
- `pnpm harness:scan` → 161 of 162. The one failure, `task-merged-citation`, is declared
  `advisory: true` in the `pr` context — "the history it grades is moved by OTHER pull requests,
  not by the change under review" — and is red because work unit 1's merged commits cite an item
  that stays `in-progress` by design across three pull requests.

**The three defects worth naming.** `/theme syntax off` persisted its setting, reported success and
changed nothing, because `renderMarkdown`'s option existed and no call site passed it. The picker
had joined one of four overlay-blocking lists, so while it was open Up/Down both previewed and
walked prompt history, Enter both selected and submitted, and Escape both cancelled and could abort
an in-flight turn. And the host's appearance writer spread its patch onto the settings document's
root — where the provider profiles and their credentials live — while its guard checked types
without rejecting unknown keys.

**The pattern the review named, and the reason it ended.** Four successive guards were written for
the first defect, and each measured a proxy for the property that mattered rather than the property:
persisted rather than rendered; the prop appearing in a file rather than reaching the provider; the
file matching a regex rather than being examined; the identifier occurring N times rather than each
call passing it. Each held for exactly one round. The fix was to remove the obligation instead of
policing it a fifth time — `useRenderMarkdown()` binds all three appearance inputs, so no call site
has an argument to forget, and the floor collapses to a single-consumer import ratchet that is
entailed by the property rather than standing beside it.

**Two CI failures were pre-existing defects this unit exposed, not defects it introduced.** The
scrollback PTY case waited on a pattern the slash autocomplete also matches while the command is
being typed; the transcript was probed rather than guessed, and CI's quoted index was the
autocomplete's. The regression-red-proof gate's file mapping did not reach the barrel export this
unit added, so the export is now pinned through the barrel.

**Open review threads:** 0.

REVIEWED BASE: b1689aa8541ba03010053bd5452ba30bebfe4bcd
REVIEWED HEAD: 795b26bd1a5f381155b366fc5271ae4a37b5a15a
ACTIONABLE FINDINGS: 0

woojubb:
PR_MERGE_DECISION
PR: 2753
HEAD: 795b26bd1a5f381155b366fc5271ae4a37b5a15a
BASE: develop
BASE-OID: b1689aa8541ba03010053bd5452ba30bebfe4bcd
VERDICT: 0
CI-OBSERVER: ci-gate-watch
CI-RESULT: GREEN
REMOTE-FEEDBACK: empty
SCOPE: SCREEN-2002 work unit 2 — the three flat appearance settings keys with their host action and per-key fallback reader, the /theme command behind an injected IThemeCataloguePort, --reduced-motion resolution with its override tier, the theme picker with preview-on-move and the theme-picker keybinding context across catalogue, published schema and guide, and the useRenderMarkdown binding that removes the per-call-site appearance obligation. No user or plugin theme files and no parseThemeDocument (work unit 3).
AUTHORITY: owner-delegated
APPROVED: yes
APPROVED-BY: agent:claude (owner-delegated)

PR #2754, opened 2026-09-19
feat(tui): themes a user or a plugin writes (SCREEN-2002 unit3)

woojubb:
## Background

SCREEN-2002 ships accessible TUI themes in three work units. Unit 1 (PR #2752) put one token model
behind every colour the terminal emits; unit 2 (PR #2753) let a user choose one. This is unit 3, the
last: the themes a user or a plugin WRITES.

Until now the registry held four built-ins and nothing else. A user who wanted the daltonized
palette with a different accent, or a plugin that wanted to ship a palette at all, had no way in.

## Accepted recommendation

`proposal-reviewer` endorsed the SCREEN-2002 design (round 3, `REVIEW VERDICT: ENDORSE`, 2026-09-19);
`finding-depth-triager` returned `DEPTH VERDICT: LOCAL`. § Decision item 7 fixes the policy this unit
implements: one strict document shape, whole-file refusal, ids minted from where the file was found.

## What changed

**The file format** (`packages/agent-ui-terminal/src/theme/theme-document.ts`). `{ name?, base?,
overrides? }` — a built-in to start from and a sparse map over the same token paths the built-ins
use. The paths are not listed in the parser; they are WALKED from the base theme, so `ITuiTheme`
gaining a key makes it overridable and losing one makes it refused. There is no second declaration
of the token model to drift from the first. Values go through `isThemeColor`, which is also the
injection floor: a raw escape sequence is not in Ink's grammar, so it cannot enter through a theme.

**Whole or nothing.** The first unknown token path, non-colour value, wrong shape or parse failure
refuses the entire file with a `$.`-path naming what was wrong. A partially-applied theme is the
state where a reader cannot tell which colours are theirs and which the base's. One unreadable file
does not take its neighbours down with it.

**Where files come from** (`packages/agent-cli/src/startup/theme-sources.ts`). `~/.robota/themes`
through the same root-bounded host contribution source `~/.robota/output-styles` uses — home-only,
because a theme is a preference of the person at the terminal. Plugin themes from `<pluginDir>/themes`
for each installed plugin, via `pluginScopeDirs`, which includes the project scope; that asymmetry is
deliberate and stated. Ids are minted here, never by the document, so no file can claim a built-in's
id whatever it is called.

**A refused file is visible.** It is carried BESIDE the themes rather than among them, printed once
at startup as `Skipped theme "<file>": <diagnostic>`, and shown in the picker as a row that carries
the same reason and cannot be chosen.

**Two things carried here from unit 2's review.** The style builder answered an unknown colour with
two different silent fallbacks — the caller's base, and `chalk.reset`; it now throws, naming the
value, because a file-level refusal is what makes such a value a defect in this package rather than
in the user's file. And the picker's toggles row, whose `s`/`m` keys are not bound in screen-reader
mode, now names `/theme syntax` and `/theme motion` in that mode instead of implying a key that does
nothing.

## What running it found

TC-13 drives the built binary in a real PTY, and it found a line that contradicts itself. `/theme
list` on a pinned run printed the SAVED value beside the tier that pinned it — `reduced motion: off
(this run: reduced motion pinned by flag)` — and the picker's toggles row did the same. The tier
alone cannot say which way a run went, because `--no-reduced-motion` is an override too and pins the
opposite value. Both surfaces now report two facts: what this run does, and what is saved. The pin is
one value on the wire rather than two optional halves, so the half-supplied case is unconstructible.

TC-13's own Scenario 2 is amended with its reason recorded inline: the criterion it asked for was the
wrong question, and the coverage it named lives where it is deterministic (the NO_COLOR PTY scenario,
and TC-06's fake-timer tests) rather than in frame timing against a replayed turn.

## Verification

- **TC-13**: four PTY cases against the built binary — the skip line at startup, `/theme list` over
  built-in + user + plugin themes, `/theme custom:mine` changing the frame's bytes with no restart,
  the picker previewing on move and restoring on escape with the refused file as a row, `/theme
  syntax off` measured on the rendered line's own SGR, and the motion pin reported for the run.
- **TC-11**: the parser (19 cases), the loader over a temp HOME and a real plugin scope layout (15),
  the registry's skip carrier, and the picker's rows.
- **TC-12**: `pnpm --filter` build/test/typecheck green for the five affected packages
  (2 / 1760 / 361 / 1003 / 547); `pnpm lint --max-warnings 2356` → 2353, 0 errors;
  `pnpm harness:scan` → 161 of 162.

`task-merged-citation` is the one red scan. It is advisory in the `pr` context and already red on
`origin/develop`: it counts merged commits citing an `in-progress` Task. This PR improves it from 14
unreconciled commits to 6 by naming the Plan's units with the tokens the commits actually cite; the
remaining six are unit 1's, which cite `(SCREEN-2002)` with no token at all and cannot be reconciled
by any edit to the record. They stop counting when SCREEN-2002 reaches a terminal status, which is
the next thing to happen to it.

## Review

Six local rounds with `pr-review-reviewer`, converging 4 → 4 → 2 → 1 → 1 → 0 actionable findings.
The substantive ones, all fixed: a refused file could put an escape sequence on the terminal through
the message refusing it (both the 7-bit and the 8-bit spelling); a theme NAME was applied and
rendered for the session without being checked; a file name that could not be typed as an id was
listed and selectable and answered the usage line when chosen; a plugin manifest's name reached the
same places unchecked; and three tests measured proxies for the behaviour they named rather than the
behaviour.

Refs: issue #2670


woojubb:
Local review of this branch, six rounds with `pr-review-reviewer`, converging 4 → 4 → 2 → 1 → 1 → 0 actionable findings. Every round was verified at the source and each fix was red-proved against the code it replaced.

What the rounds found and this branch fixes:

- A refused theme file could put an escape sequence on the terminal **through the message refusing it** — the grammar is the injection floor for theme VALUES, and the diagnostic interpolated the rejected value raw. Both the 7-bit and the 8-bit spelling (U+009B CSI, U+009D OSC); `JSON.stringify` escapes only U+0000–U+001F, which is why the escaping is one exported policy rather than a call to it.
- A theme NAME is APPLIED and drawn on every list row and picker row for the session, and was unchecked — including the minted id used when a document supplies no name.
- A file or plugin name that could not be typed as an id produced a theme that was listed, was selectable, and answered the usage line when chosen: worse than one that does not load.
- `/theme list` and the picker reported the SAVED reduced-motion value beside the tier that pinned it, which contradicts itself — `--no-reduced-motion` is an override too and pins the opposite value. The pin is now one value on the wire, so the half-supplied case is unconstructible.
- Three tests measured a proxy for the behaviour they named rather than the behaviour, and one guard was removed rather than kept unfalsifiable, with the reason recorded where the code is.

CI on this head: every check green, `mergeable=MERGEABLE state=CLEAN`. `regression-red-proof` went red on an earlier head for a real gap — a defect-fix commit's only change to the style builder was one grammar string with no test reading it — and was fixed with a test rather than an opt-out. `task-merged-citation` is advisory in the `pr` context and already red on `origin/develop`; this branch improves it from 14 unreconciled commits to 6.

REVIEWED BASE: c6e3014e71b9fbd32933cdd61a06f686a44bd404
REVIEWED HEAD: a555afe803187453c50943634d0a17a320397d55
ACTIONABLE FINDINGS: 0

woojubb:
PR_MERGE_DECISION
PR: 2754
HEAD: a555afe803187453c50943634d0a17a320397d55
BASE: develop
BASE-OID: c6e3014e71b9fbd32933cdd61a06f686a44bd404
VERDICT: 0
CI-OBSERVER: ci-gate-watch
CI-RESULT: GREEN
REMOTE-FEEDBACK: empty
SCOPE: SCREEN-2002 work unit 3 — user and plugin theme files, their whole-file refusal, and the TC-13 PTY scenario
AUTHORITY: owner-delegated
APPROVED: yes
APPROVED-BY: agent:claude-opus-5 (owner-delegated)

